PR #2409, opened 2026-02-22
fix: respect last-flag-wins for --no-compress vs --compress-level

oferchen:
## Summary
- Fix `--no-compress --compress-level=3` silently re-enabling compression when `--no-compress` should take precedence (it appeared last)
- Use clap argument index tracking (`last_occurrence`) to compare positions of `--no-compress` and `--compress-level`, letting the later flag win
- Add 4 test cases covering both orderings and level-zero edge cases

## Upstream Analysis
Upstream rsync (options.c:2002-2004) unconditionally re-enables compression when `--compress-level` is specified with any value. Our implementation adds last-flag-wins semantics using clap's argument index tracking, consistent with how other flag pairs (e.g., `--owner`/`--no-owner`) are resolved in this codebase via `tri_state_flag_with_order`.

## Test plan
- [x] `--no-compress --compress-level=3` enables compression (level appears last)
- [x] `--compress-level=3 --no-compress` disables compression (no-compress appears last)
- [x] `--compress-level=0 --no-compress` stays disabled (both disable)
- [x] `--no-compress --compress-level=0` stays disabled (level 0 disables)
- [x] All 614 existing compression-related tests pass
- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace --all-targets --all-features --no-deps -- -D warnings` passes

### DIFF
--- a/crates/cli/src/frontend/arguments/parser.rs
+++ b/crates/cli/src/frontend/arguments/parser.rs
@@ -226,7 +226,15 @@ where
     if let Some(ref value) = compress_level_opt
         && let Ok(setting) = parse_compress_level_argument(value.as_os_str())
     {
-        compress = !setting.is_disabled();
+        // upstream: options.c — last-flag-wins between --no-compress and --compress-level
+        let level_implies_compress = !setting.is_disabled();
+        if no_compress && level_implies_compress {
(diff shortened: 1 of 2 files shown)


oferchen:
Closing: will be redone cleanly from master. CI was failing and branch is behind.

PR #3124, opened 2026-04-06
test: add delete-after interop test for daemon push

oferchen:
## Summary
- Add standalone interop test test_delete_after to tools/ci/run_interop.sh
- Tests upstream rsync pushing with --delete-after to an oc-rsync daemon
- Verifies source files arrive at destination and extra files/directories are removed

## Test plan
- [ ] CI interop validation workflow passes with new test included
- [ ] Test creates source files, pre-populates destination with extras, pushes with --delete-after, verifies correct state

### DIFF
--- a/tools/ci/run_interop.sh
+++ b/tools/ci/run_interop.sh
@@ -2911,6 +2911,87 @@ CONF
   return 0
 }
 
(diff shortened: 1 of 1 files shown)


oferchen:
Superseded by #3127 which combines all three interop tests into a single PR.

PR #3125, opened 2026-04-06
test: add hardlinks interop test for daemon push

oferchen:
## Summary
- Add standalone test_hardlinks interop test that pushes from upstream rsync with --hard-links -av to an oc-rsync daemon
- Verifies all files arrive with correct content and that hardlink relationships (shared inodes) are preserved, including cross-directory hardlinks
- Includes a standalone control file to confirm hardlink groups have distinct inodes

## Test plan
- [ ] CI interop tests pass with the new hardlinks standalone test
- [ ] Existing standalone tests remain unaffected

### DIFF
--- a/tools/ci/run_interop.sh
+++ b/tools/ci/run_interop.sh
@@ -2575,6 +2575,124 @@ test_hardlinks_comprehensive() {
   return 0
 }
 
+# Hardlinks daemon push interop test.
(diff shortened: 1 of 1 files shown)


oferchen:
Superseded by #3127 which combines all three interop tests with the cross-directory hardlink assertion fix.

PR #3126, opened 2026-04-06
test: add many-files interop test for 1000+ small file daemon push

oferchen:
## Summary
- Add test_many_files standalone interop test to tools/ci/run_interop.sh
- Creates 1000 small files (1-4096 bytes) across 10 subdirectories, pushes from upstream rsync to oc-rsync daemon, verifies all files arrive with correct content
- Exercises file-list handling, pipeline throughput, and content fidelity at scale
- Includes re-sync idempotency check (second push transfers zero files)

## Test plan
- [ ] CI interop workflow passes with the new many-files standalone test
- [ ] Verify file count (1002 files) and aggregate md5 checksum match after daemon push
- [ ] Verify re-sync produces no unnecessary file transfers

### DIFF
--- a/tools/ci/run_interop.sh
+++ b/tools/ci/run_interop.sh
@@ -2911,6 +2911,166 @@ CONF
   return 0
 }
 
+# Test pushing 1000+ small files from upstream rsync to oc-rsync daemon.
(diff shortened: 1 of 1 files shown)


oferchen:
Superseded by #3127 which combines all three interop tests with the relative-path md5sum fix.

PR #3162, opened 2026-04-11
feat: add threshold-gated parallel dispatch for receiver delta pipeline

oferchen:
## Summary

- Add `ParallelDeltaPipeline` that dispatches delta work items to rayon workers with channel-based result collection and `BTreeMap` reordering to preserve submission order
- Add `ThresholdDeltaPipeline` that auto-selects sequential dispatch below 64 files and parallel dispatch at or above 64, matching the `PARALLEL_STAT_THRESHOLD` pattern used for stat batching in the receiver
- Add `DEFAULT_PARALLEL_THRESHOLD` constant (64) and export all new types from the `transfer` crate

## Test plan

- [x] ParallelDeltaPipeline: single/multi submit, flush ordering, mixed work kinds, trait object, large batch (200 items), poll-then-flush partial drain
- [x] ThresholdDeltaPipeline: below/at/above threshold selection, custom thresholds, sequential and parallel path end-to-end, trait object, default/empty flush
- [x] Constant value assertion matches receiver's PARALLEL_STAT_THRESHOLD
- [ ] CI: fmt+clippy, nextest, Windows, macOS, Linux musl

### DIFF
--- a/crates/transfer/src/delta_pipeline.rs
+++ b/crates/transfer/src/delta_pipeline.rs
@@ -10,8 +10,12 @@
 //! - [`SequentialDeltaPipeline`] - processes items immediately in the calling
 //!   thread with no concurrency. This matches upstream rsync's sequential
 //!   `recv_files()` loop in `receiver.c` and is the default.
-//! - A parallel implementation (dispatching to rayon workers via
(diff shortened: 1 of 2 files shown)


oferchen:
Will be combined with parallel pipeline in a rebased PR

PR #3174, opened 2026-04-11
feat: add per-operation parallel thresholds for stat, signature, and metadata

oferchen:
## Summary

- Replace the fixed `PARALLEL_STAT_THRESHOLD = 64` constant with a `ParallelThresholds` struct providing per-operation thresholds: `stat` (64), `signature` (32), `metadata` (64), `deletion` (64)
- Thread `ParallelThresholds` through `ReceiverContext` and `GeneratorContext` so each call site uses the appropriate threshold for its operation type
- CPU-bound signature computation now uses a lower default threshold (32) since it benefits from parallelism earlier than I/O-bound stat calls

## Test plan

- [ ] Verify CI passes (fmt, clippy, nextest on all platforms)
- [ ] Confirm default thresholds match previous behavior (stat=64, metadata=64, deletion=64) for backward compatibility
- [ ] Confirm signature threshold (32) activates parallel path earlier for CPU-bound work
- [ ] Unit tests for `ParallelThresholds` default values, builder methods, and `Copy`/`Eq` semantics

### DIFF
--- a/crates/protocol/tests/zstd_interop_golden_bytes.rs
+++ b/crates/protocol/tests/zstd_interop_golden_bytes.rs
@@ -373,8 +373,7 @@ fn interop_decode_handcrafted_deflated_data_block() {
     let input = b"handcrafted upstream wire bytes";
 
     // Compress with standalone zstd to get raw payload (simulating upstream encoder)
-    let mut raw_encoder =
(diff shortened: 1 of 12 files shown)


oferchen:
Closing due to merge conflicts. Will recreate on a fresh branch from master.

PR #3421, opened 2026-04-28
perf: set IOSQE_ASYNC on io_uring SEND submissions

oferchen:
## Summary

Force the kernel to dispatch io_uring SEND on its async worker pool from the start by setting `IOSQE_ASYNC`, instead of attempting it inline on the submitting task. When the socket is not immediately writable (TCP backpressure), this prevents `submit_and_wait` from stalling the issuing task while the peer drains the receive window.

## Why

`socket_writer` and `socket_reader` build independent rings for the write and read halves of a single TCP connection. Without `IOSQE_ASYNC`, a SEND that cannot complete inline causes the kernel to retry it on the submitter's task path, which can produce head-of-line blocking between unrelated send and recv rings on a bidirectional TCP exchange (the daemon push/pull case where reorder buffer drains compete with checksum reads).

This is the io_uring equivalent of upstream rsync's `socket.c::writefd_unbuffered`, which uses non-blocking writes with `select(2)` to wait for write-ready before each write.

## Changes

- Add `build_send_sqe()` helper that constructs a SEND SQE with `Flags::ASYNC` set, mirroring the existing `maybe_fixed_file` wrapping pattern.
- Use the helper from `submit_send_batch()` so every SEND in a batch carries the flag.

## Test plan

- [ ] CI fmt+clippy passes
- [ ] CI nextest passes (no behavior change on platforms without io_uring; on Linux, SEND completes via async worker rather than inline)
- [ ] Benchmark daemon TCP push throughput vs prior batching code (separate run, not in this PR)

Tracks the io_uring-socket-deadlock root-cause work documented alongside #3417/#3418.


oferchen:
## Root cause: IOSQE_ASYNC breaks stream ordering for batched parallel SENDs

CI failure on `Feature flag combinations / copy_file_range`:

```
FAIL [   5.134s] (375/375) fast_io io_uring::tests::test_socket_large_payload_roundtrip
test io_uring::tests::test_socket_large_payload_roundtrip ... FAILED
thread '...' panicked at crates/fast_io/src/io_uring/tests.rs:926:5:
assertion `left == right` failed
```

Lengths match at line 925; bytes differ at line 926 — content corruption, not data loss.

### Why this fails

`submit_send_batch` pushes multiple parallel SEND SQEs in one submission, each targeting a different chunk of the same FD (slot 0 = `data[0..8192]`, slot 1 = `data[8192..16384]`, ...). On master without `IOSQE_ASYNC`, the kernel processes inline-submitted SQEs strictly in submission order on the submitter task, so wire byte order matches data offset order.

With `IOSQE_ASYNC`, every SEND is dispatched to the io_uring async worker pool. Multiple workers race for the socket send-lock. Whichever worker acquires the lock first gets its bytes on the wire first — order is non-deterministic. For SOCK_STREAM the receiver reads bytes in wire order, so slot 1's chunk can arrive before slot 0's, corrupting the stream.

Test flakiness across the matrix confirms this: the `io_uring` feature combo passed, `copy_file_range` failed — same code, different scheduling, different result.

### Why this is fundamental, not a test bug

The existing batched-SEND architecture relies on inline serialization for ordering. `IOSQE_ASYNC` removes that guarantee. Possible correct approaches:

- **Revert.** Keep inline submission; address #1872 differently (e.g., POLL_ADD with EPOLLOUT before SEND).
- **Single SEND per batch.** Drop the batched parallelism; submit one SEND at a time. `IOSQE_ASYNC` becomes safe but throughput regresses.
- **`IOSQE_IO_LINK` chains** are not viable: a short send on slot N still leaves slot N+1 already submitted with its own pre-computed offset, so resubmission of the partial slot N would land out-of-order on the wire.

Closing this PR. Reopening #1872 as needing a different design (likely POLL_ADD-based write-readiness wait before each SEND, mirroring upstream rsync's `socket.c::writefd_unbuffered` select(2) pattern referenced in this PR's own commit message).

oferchen:
Closing per root-cause analysis above: `IOSQE_ASYNC` on parallel batched SENDs breaks SOCK_STREAM byte ordering. Reopening #1872 with a redesign that uses POLL_ADD/EPOLLOUT for write-readiness rather than dispatching SENDs to async workers.

PR #3609, opened 2026-05-04
feat(fast_io): add IocpDiskBatch matching IoUringDiskBatch surface (#1898)

oferchen:
## Summary

- Adds `IocpDiskBatch` in `crates/fast_io/src/iocp/disk_batch.rs` that mirrors the `IoUringDiskBatch` public surface (`begin_file` / `write_data` / `flush` / `commit_file` / `bytes_written` / `bytes_written_with_pending`) but submits overlapped `WriteFile` calls through a private completion port and drains them with `GetQueuedCompletionStatusEx`.
- Reopens the file with `FILE_FLAG_OVERLAPPED` via `ReOpenFile`, pins each in-flight `OverlappedOp` so `WriteFile` sees a stable `OVERLAPPED` pointer, and matches completions by overlapped address. Best-effort flush on `Drop` so partial submissions cannot leak overlapped I/O.
- Wires the new batch into the disk-commit pipeline: `DiskCommitConfig` gains `iocp_policy: fast_io::IocpPolicy` (default `Auto`), `Writer` gains an `Iocp` variant (Windows + `iocp` feature only), and `disk_thread_main` constructs a single batch up-front but only when io_uring is not active so the two backends stay mutually exclusive.
- Cross-platform builds keep compiling: `iocp_stub.rs` exposes the same surface returning `Unsupported` / `None` / `0`. Sparse mode and append paths still take the buffered writer.

Resolves part of #1898 (does not touch the completion-port pump being reworked in #1897 - this layers on top).

## Test plan

- [x] `cargo nextest run -p fast_io --features iocp -E 'test(disk_batch)'` (Windows runner via CI)
- [x] `cargo nextest run -p transfer --all-features -E 'test(disk_commit)'`
- [x] `cargo check --workspace --all-targets` on Linux + macOS to confirm the stub keeps non-Windows builds green
- [x] CI: fmt+clippy, nextest (stable), Windows (stable), macOS (stable), Linux musl (stable)

New unit tests in `disk_batch.rs` cover:
- batched submission of N writes
- completion ordering independent of submission order
- error propagation from a failed `reopen_overlapped`
- large writes that exceed the buffer
- fsync via `FlushFileBuffers` on `commit_file(true)`
- `Drop` flushing pending data
- many file rotations to verify no leaked overlapped handles


oferchen:
@dependabot rebase

PR #4098, opened 2026-05-15
test(filters): stress-test nested .rsync-filter inheritance against upstream

oferchen:
## Summary

Adds a stress test for `FilterChain` per-directory `.rsync-filter` inheritance against a five-level fixture tree. Covers parent/child override interactions, anchor-style variants (`/` vs prefix), sibling subtree isolation, leave-directory pop semantics, and a golden decision matrix snapshot. Locks the upstream-equivalent include/exclude outcomes so future changes to filter precedence surface before wire divergence.

- Fixture tree: top-level `+ /allowed/**` / `- *`, mid-level `- /allowed/secret` + `- notes.txt`, innermost `+ keep.md` / `- *`.
- Exercises `push_local_filters` / `pop_local_filters` semantics: first-match-wins, innermost scope priority, anchored vs unanchored matching, sibling independence.
- Snapshot test pins the full include/exclude decision matrix.

Reference: upstream rsync 3.4.1 `exclude.c:push_local_filters` / `pop_local_filters`.

Refs #2207

## Test plan

- [ ] CI: `cargo nextest run -p filters --all-features -E 'test(nested_rsync_filter_stress)'`
- [ ] CI: fmt + clippy across the workspace

### DIFF
--- a/crates/filters/tests/nested_rsync_filter_stress.rs
+++ b/crates/filters/tests/nested_rsync_filter_stress.rs
@@ -69,11 +69,7 @@ fn build_tree() -> TempDir {
 
 /// Push scopes for `root`, then every ancestor between `root` and `subdir`.
 /// Returns the guards in pop order so the caller can release them.
-fn enter_to(
-    chain: &mut FilterChain,
-    root: &Path,
-    subdir: &Path,
-) -> Vec<filters::DirFilterGuard> {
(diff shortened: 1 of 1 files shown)


oferchen:
Filter expectations in the stress fixture don't match FilterChain's actual rule precedence (anchored-glob '/allowed/**' does not match the directory 'allowed' itself, only its contents; '- *' then excludes it). Closing to rework with corrected semantics — task #2207 reopened.

PR #4102, opened 2026-05-15
perf(engine): BufferPool byte-budget cap to bound memory under large buffers

oferchen:
## Summary

- Add optional `pooled_bytes_budget` field to `BufferPool` that caps the total `capacity()` of buffers retained in the central queue.
- Hybrid with existing count-based soft cap: whichever fires first prevents admission. Defends against very-large adaptive buffers (e.g. 1 MB `ADAPTIVE_BUFFER_HUGE`) blowing through the count-only ceiling.
- Threaded through `GlobalBufferPoolConfig::pooled_bytes_budget`; default `None` preserves the historical count-only behavior so existing callers see no change.
- Atomic `pooled_bytes` counter updated on every admit and pop; relaxed ordering keeps the overhead at one extra add/sub per central-queue operation.

Refs #2245.

## Test plan

- [ ] `pooled_bytes_budget_drops_oversized_buffers`: 1 MB buffers + 16 KB budget; central queue stays empty and pooled bytes stay within budget.
- [ ] `pooled_bytes_budget_high_retains_normally`: generous budget; count cap fires first and pool fills its slots.
- [ ] `count_cap_still_enforced_without_byte_budget`: regression test for existing count-only path.
- [ ] `pooled_bytes_tracks_admission_and_pop`: pooled-byte counter increments on admit and decrements on pop.
- [ ] `pooled_bytes_budget_rejects_zero`: `with_pooled_bytes_budget(0)` panics.
- [ ] `pooled_bytes_budget_zero_is_treated_as_unbounded` in `global.rs`: filter logic mirrors `memory_cap=Some(0)`.

### DIFF
--- a/crates/engine/src/local_copy/buffer_pool/global.rs
+++ b/crates/engine/src/local_copy/buffer_pool/global.rs
@@ -46,6 +46,13 @@ pub struct GlobalBufferPoolConfig {
     /// uncapped, matching its historical default. A value of `Some(0)` is
     /// treated as `None`.
     pub memory_cap: Option<usize>,
+    /// Optional byte-budget cap for buffers retained in the central pool.
+    ///
+    /// When `Some`, returned buffers are only kept if pooled bytes plus the
+    /// buffer's capacity stay within the budget. Otherwise the buffer is
+    /// deallocated. `None` preserves count-only behavior. A value of `Some(0)`
+    /// is treated as `None`.
+    pub pooled_bytes_budget: Option<u64>,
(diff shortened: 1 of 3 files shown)


oferchen:
Tests reveal an accounting bug in the agent's pooled_bytes tracking (test pooled_bytes_tracks_admission_and_pop expects 0 but gets 1). Multiple test failures across platforms after the first fmt round. Closing for rework; task #2245 reopened.

PR #4299, opened 2026-05-17
feat(engine): wire SpillableReorderBuffer with bounded-memory threshold (#1884)

oferchen:
## Summary

- Adds `ConcurrentDeltaConfig` with an opt-in `spill_threshold_bytes` knob and a new `DeltaConsumer::spawn_with_config` constructor that swaps the bare `ReorderBuffer` for the existing `SpillableReorderBuffer` when the threshold is set.
- Replaces the legacy `force_insert` deadlock-breaker fallback with a proper spill: capacity pressure now displaces the highest-sequence items to a tempfile (`SpooledTempFile` by default, with an explicit `spill_dir` override) instead of doubling the ring forever.
- Exposes spill activity through an `AtomicU64` accessible via `DeltaConsumer::stats()` -> `DeltaConsumerStats { spill_events }`. Spill I/O failures map to `DeltaResult::failed`, which the receiver already translates to upstream exit code 11 (`FileIo`).

The default path (`spawn`, `spawn_bypass`, and `spawn_with_config(.., ConcurrentDeltaConfig::default())`) is bit-for-bit identical to the prior behaviour. The wire protocol is unchanged - the spill layer is strictly receiver-local memory management.

## Test plan

- [x] Synthetic 1000-item / 16 KiB-budget adversarial drive: `spillable_consumer_preserves_order_under_pressure` verifies in-order delivery and `spill_events > 0`.
- [x] Byte-for-byte parity vs the non-spilling baseline: `spillable_consumer_matches_bare_output_byte_for_byte` encodes each delivered `DeltaResult` via `SpillCodec` and asserts identical payloads against the bare-`ReorderBuffer` baseline.
- [x] Default-off path: `spawn_with_config_off_matches_spawn` and `stats_zero_when_spill_disabled` confirm `spawn_with_config(.., ConcurrentDeltaConfig::off())` matches `spawn(..)` exactly and reports zero spill events.
- [x] `ConcurrentDeltaConfig` unit tests cover `default`, `off`, `with_spill_threshold`, `with_spill_dir`, and `spill_enabled`.
- [x] `cargo fmt --all` clean.

### DIFF
--- a/crates/engine/src/concurrent_delta/config.rs
+++ b/crates/engine/src/concurrent_delta/config.rs
@@ -1,0 +1,114 @@
+//! Runtime configuration for the concurrent delta pipeline.
+//!
+//! [`ConcurrentDeltaConfig`] aggregates the knobs that tune the
+//! [`DeltaConsumer`](super::consumer::DeltaConsumer) without changing its
+//! public type signature. The only knob today is
+//! [`spill_threshold_bytes`](ConcurrentDeltaConfig::spill_threshold_bytes),
+//! which opts the consumer into bounded-memory spill-to-tempfile via
+//! [`SpillableReorderBuffer`](super::spill::SpillableReorderBuffer).
+//!
+//! # Defaults
+//!
+//! [`ConcurrentDeltaConfig::default`] returns a configuration that mirrors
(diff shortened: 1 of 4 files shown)


oferchen:
Closing due to unresolvable merge conflicts after drain instrumentation landed. Re-doing fresh on top of current master.

PR #4300, opened 2026-05-17
feat(engine): parallel receive-side delta apply behind --features parallel-receive-delta (#1368)

oferchen:
## Summary

- Scaffold the parallel receive-side delta apply path behind a new
  `parallel-receive-delta` feature on the `engine` crate (forwarded
  from the `transfer` crate). Default off; the production sequential
  loop is unchanged.
- New module `engine::concurrent_delta::parallel_apply` introduces
  `DeltaChunk` and `ParallelDeltaApplier`. Per-file `Mutex<FileSlot>`
  serialises writes; per-file `ReorderBuffer` replays chunks in
  submission order so per-file byte order is preserved even when the
  rayon verify step completes out of order.
- `ReceiverContext::enable_parallel_receive_delta` installs the
  existing `ParallelDeltaPipeline` only when the feature is compiled
  in, keeping the receiver loop untouched otherwise.

Implements the opt-in scaffold prescribed by
`docs/design/parallel-receive-delta-application.md` section 6.3
(phase 2). The bench trigger from #4214 is the gating signal noted in
the task; this PR lands the opt-in surface so subsequent benches can
run against the path without disturbing default builds.

## Test plan

- [ ] CI: fmt + clippy clean across workspace
- [ ] CI: nextest (stable) green
- [ ] CI: Windows / macOS / Linux musl matrices green
- [ ] Unit: `single_file_in_order_matches_sequential` passes byte-equal
- [ ] Unit: `single_file_out_of_order_preserves_byte_order` passes
      byte-equal
- [ ] Unit: `batch_apply_matches_sequential_byte_for_byte` passes
      byte-equal across two files
- [ ] Property: `random_chunk_sizes_and_permutations_match_sequential`
      (48 cases, deterministic permutation per seed) passes
- [ ] Unit: error paths pinned - unknown file, double registration,
      finish with pending chunks

### DIFF
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -90,6 +90,13 @@ parallel = []
 # Trade-offs: Relaxes the single-producer compile-time invariant
 multi-producer = []
 
+# Parallel receive-side delta apply (#1368) - scaffolds the per-file rayon
+# fanout for checksum verification while keeping the per-file destination
+# write serial. Default off so production receivers continue to use the
+# sequential apply loop. Opt-in for benchmarking and the phased rollout
+# described in `docs/design/parallel-receive-delta-application.md`.
+parallel-receive-delta = []
+
 # Per-thread buffer slab (#1271, #1370) - replaces the single-slot thread-local
 # cache in front of BufferPool with a depth-bounded LIFO slab per thread.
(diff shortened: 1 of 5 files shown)


oferchen:
Closing due to unresolvable merge conflicts. Re-doing fresh on top of current master.

PR #4364, opened 2026-05-17
feat(engine): try_unwrap_or_log diagnostic helper (#2382)

oferchen:
## Summary
- New `engine::util::try_unwrap_or_log<T>(Arc<T>, &'static str) -> Result<T, Arc<T>>` helper that logs `strong_count`, `weak_count`, and a callsite tag via `tracing::warn!` before returning the residual `Arc`.
- Helper is re-exported from `engine::util` and the crate root. The warn emission is gated behind the existing `tracing` feature (no-op when disabled), so consumers do not pull in a new dependency.
- Replaced the two ATU-3 drain-path `Arc::try_unwrap` callsites with the helper:
  - `delete::context::DeleteContext::into_emitter` (plans + cursor)
  - `concurrent_delta::parallel_apply::ParallelApplier::finish_file` (per-file slot)

Typed errors returned to callers are unchanged; the helper only adds context to the log stream before the existing `io::Error::other(...)` fires.

## Test plan
- [x] `cargo fmt --all`
- [ ] CI: fmt+clippy, nextest (stable), Windows, macOS, Linux musl
- [ ] Unit test exercises three paths: sole-owner success, strong-clone failure, weak-ref failure. A fourth `cfg(feature = "tracing")` test installs a minimal `Subscriber` and asserts the warn event carries `strong_count`, the callsite tag, and the failure message.

### DIFF
--- a/crates/engine/src/concurrent_delta/parallel_apply.rs
+++ b/crates/engine/src/concurrent_delta/parallel_apply.rs
@@ -44,6 +44,7 @@ use rayon::prelude::*;
 
 use super::reorder::ReorderBuffer;
 use super::types::FileNdx;
+use crate::util::try_unwrap_or_log;
 
 /// A single contiguous segment of a per-file delta apply.
 ///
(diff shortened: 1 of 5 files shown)


oferchen:
Superseded: try_unwrap_or_log helper is in master

PR #4396, opened 2026-05-18
feat(engine): honor SpillPolicy.spill_dir for spill temp-file location (STN-3 #2337)

oferchen:
## Summary
- Plumb `SpillPolicy.dir` into the spill temp-file creation path via `tempfile::Builder::new().prefix("oc-rsync-spill-").tempfile_in(...)`.
- Default remains `env::temp_dir()` when the field is `None`; resolution happens on the first spill of every batch so policy edits between consumer spawns take effect without being cached at construction.
- Create the target directory with `create_dir_all` on demand if missing; map the failure to a typed `SpillError::DirCreate { path, source }` variant rather than `io::ErrorKind::Other`.
- Drop the eager `create_dir_all` in `SpillableReorderBuffer::with_spill_dir` (now infallible and lazy) and simplify the consumer spawn path now that construction cannot fail.
- Three unit tests cover the default path, the override path, and the on-demand missing-parent case via a test-only `open_spill_path` accessor.

## Test plan
- [x] `cargo fmt --all` clean
- [x] Tests cover default (`env::temp_dir`), explicit override, and missing-parent dir-create cases
- [x] No public API change to `SpillPolicy`; `SpillError` gains a new `DirCreate` variant alongside existing `Io`/`Capacity`

Closes #2337.

### DIFF
--- a/crates/engine/src/concurrent_delta/consumer.rs
+++ b/crates/engine/src/concurrent_delta/consumer.rs
@@ -68,7 +68,9 @@ enum ReorderMode {
     /// Bare in-memory ring with the historical doubling fallback on overflow.
     Bare { capacity: usize },
     /// Bounded-memory ring with spill-to-tempfile when the byte threshold is
-    /// exceeded. `dir` is `None` for the default `SpooledTempFile` backend.
+    /// exceeded. `dir` is `None` to defer to [`std::env::temp_dir`] at spill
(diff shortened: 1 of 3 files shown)


oferchen:
Superseded: SpillPolicy.dir (spill_dir override) is in master via #4356-era STN-3 squash-merge. The 14 hunks of conflict in spill.rs are exactly where master already has the wired implementation.

PR #4426, opened 2026-05-18
refactor(engine): extract spill/buffer.rs (SPL-4 #2326)

oferchen:
## Summary
- Move `SpillableReorderBuffer<T>` and its `impl` block from `spill.rs` into the new `spill/buffer.rs` submodule per the SPL-1 audit's leaves-first order.
- Parent `spill.rs` re-exports `SpillableReorderBuffer` so the public API at `crate::concurrent_delta::spill::SpillableReorderBuffer` is unchanged.
- Buffer-specific tests (in-memory/spill/reload round-trips, threshold boundary, ENOSPC and temp-dir-vanish hardening, directory-backed round-trip, partial-write contract) move with the type. The codec round-trip tests for `DeltaResult` and the `SpillError` Display/source test stay in the parent.
- `SpillBackend`, `ReadWriteSeek`, and `open_backend` are reduced to `pub(super)` so the new submodule can call them without widening the public surface.

## Note: buffer.rs LoC
- Parent `spill.rs` shrinks from 1232 to 282 lines (well under the 650 cap).
- `spill/buffer.rs` lands at 972 lines: ~390 lines of code plus ~580 lines of tests. The tests are tightly coupled to the buffer (`FailingCodec`, `impl SpillCodec for u64`, scratch-tempdir scenarios) and split poorly. Splitting the test module would create a second file whose only consumer is itself; SPL-5/SPL-7 will revisit this once stats/codec extractions reshape the buffer surface.

## Test plan
- [x] `cargo fmt --all` clean
- [x] `cargo check -p engine --all-features --tests` clean (no new warnings vs master)
- [x] `cargo clippy -p engine --all-targets --all-features --no-deps -- -D warnings` clean (no new lints vs master)
- [x] No public-API change: SPL-9 audit table (`docs/audits/spl-9-mod-reexports.md`) entries for `SpillableReorderBuffer` + 12 inherent methods still resolve at `crate::concurrent_delta::spill::SpillableReorderBuffer::*` via the parent re-export.

Closes #2326.


oferchen:
Needs to be redone against current master. Since this PR was opened, STN-3 (spill_dir), STN-4 (reclaim), STN-5 (granularity), STN-6 (memory_pressure), STN-7 (compression) all added methods to SpillableReorderBuffer that this PR's extracted spill/buffer.rs is missing. Mechanical conflict resolution would lose those methods. A fresh extraction off latest master is the right path.

PR #4537, opened 2026-05-18
ci: enforce LoC limits as a real required check (pin 41 over-cap files)

oferchen:
## Summary

- The \`enforce-limits\` CI job was \`continue-on-error: true\` and emitted 42 \`##[error]\` annotations per run. It burned CI cycles, made every PR look red in the rollup, and gated nothing - the worst of all worlds.
- Two changes together fix the problem:
  1. **Pin** each of the 41 currently over-cap source files at its current line count in \`tools/line_limits.toml\` (matches the existing pattern for \`buffer_pool/pool.rs\` and \`md5_simd/avx512.rs\`).
  2. **Promote** the job to a real required check: drop \`continue-on-error: true\`, rename from \`enforce-limits (informational)\` to \`enforce-limits\`.

After this lands, any new line that pushes a file past its cap forces an explicit override bump or an SPL decomposition - which is the entire point of a ceiling.

## Test plan

- [ ] \`enforce-limits\` job goes green on this PR (no \`##[error]\` annotations)
- [ ] No required check regresses
- [ ] Pinned overrides match current line counts (no off-by-one)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,23 +92,20 @@ jobs:
         run: cargo clippy --locked --workspace --all-targets --all-features --no-deps
 
   # ===========================================================================
-  # Enforce LoC Limits - Informational (non-blocking) check
+  # Enforce LoC Limits - Required check
   # ---------------------------------------------------------------------------
-  # Runs `cargo xtask enforce-limits` to surface files that exceed the per-file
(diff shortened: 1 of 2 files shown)


oferchen:
Superseded: enforce-limits is being deleted entirely (CPU waste, wrong metric). Large-file hygiene moves to one-shot SPL decomposition.

PR #4546, opened 2026-05-19
fix(fast_io): IOCP disk-full retry and IocpWriter completion leak

oferchen:
## Summary
- `IocpDiskBatch::flush_current` no longer discards drained-but-uncredited bytes on a mid-flush failure, so `Drop` does not re-emit chunks the kernel already accepted. This is what kept `nth_submission_disk_full_skips_earlier_writes` finding 16 KB on disk instead of the expected 8 KB.
- `IocpWriter` now sets `FILE_SKIP_COMPLETION_PORT_ON_SUCCESS` on its handle. Synchronous `WriteFile` returns no longer leak packets onto the port, so the next `GetQueuedCompletionStatus` call cannot dequeue a stale completion whose pinned buffer has already dropped. This is the corruption symptom behind `writer_accumulates_across_implicit_flushes`.

## Root cause
1. `submit_write_batch` returned `io::Result<usize>` and threw away the partial count whenever `submit_one_write` faulted; `flush_current` then left `buffer_pos` untouched, so `Drop`'s second flush_current rewrote the entire payload over the post-fault tail.
2. `IocpWriter::write_at` issued every `WriteFile` against a handle that still queued completion packets on synchronous success but only waited for completions on `ERROR_IO_PENDING`. A backlog of stale packets meant the next async write dequeued the wrong one and reported success while the kernel kept reading from freed memory.
3. The fix wires partial progress through an out-parameter on `submit_write_batch` and opts each `IocpWriter` handle into the skip flag so sync successes never queue.

## Test plan
- [ ] CI Windows IOCP job: `fast_io::iocp_disk_full_simulation::nth_submission_disk_full_skips_earlier_writes`
- [ ] CI Windows IOCP job: `fast_io::iocp_partial_write_integration::writer_accumulates_across_implicit_flushes`
- [ ] CI Windows IOCP job: `fast_io::iocp_partial_write_integration::disk_batch_accumulates_under_simulated_pressure` (regression guard)
- [ ] CI Windows IOCP job: remaining `iocp_disk_full_simulation` tests (`writer_drop_after_disk_full_is_clean`, `batch_recovers_after_injected_fault_consumed`, `first_submission_disk_full_surfaces_storage_full`)
- [ ] Linux fmt+clippy and nextest matrix to confirm the non-Windows stub path is untouched

### DIFF
--- a/crates/fast_io/src/iocp/disk_batch/mod.rs
+++ b/crates/fast_io/src/iocp/disk_batch/mod.rs
@@ -318,6 +318,13 @@ impl IocpDiskBatch {
 
     /// Flushes the internal buffer to the current file via batched overlapped
     /// writes.
+    ///
+    /// On a mid-flush failure (synchronous `WriteFile` error, drained
+    /// completion that reports an NTSTATUS failure, or test-injected
+    /// `ERROR_DISK_FULL`), any chunks that did reach the kernel before the
+    /// fault are credited to `active.bytes_written` and the in-memory buffer
+    /// is cleared so the `Drop` retry does not resubmit data that already
+    /// landed. The error is then propagated to the caller unchanged.
     fn flush_current(&mut self) -> io::Result<()> {
         if self.buffer_pos == 0 {
             return Ok(());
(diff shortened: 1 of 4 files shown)


oferchen:
Superseded by #4547 (folded into the master-wide CI rescue PR).

PR #5163, opened 2026-05-30
ci: revert nextest disk-cleanup step that causes 60-minute timeouts

oferchen:
## Summary

The disk-cleanup step added in #5144 frees ~15 GB to prevent link-time disk exhaustion, but adds enough overhead to push the nextest job past its 60-minute timeout consistently. Reverting to the previous configuration which was more stable. The occasional disk-exhaustion failures are preferable to consistent timeout failures across all PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,17 +167,6 @@ jobs:
         with:
           tool: cargo-nextest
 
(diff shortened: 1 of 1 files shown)


oferchen:
Creating a broader revert that also restores the pre-PR-lite CI configuration.

PR #5457, opened 2026-06-04
perf: remove hardcoded TCP_NODELAY from daemon connections

oferchen:
## Summary

- Remove hardcoded `TCP_NODELAY` from all 3 daemon connection sites (server-side `setup_transfer_streams`, client-side `run_daemon_transfer`, and `run_daemon_over_remote_shell`)
- Upstream rsync does not set `TCP_NODELAY` on daemon connections - `clientserver.c:1346` only sets `SO_KEEPALIVE`
- The hardcoded `TCP_NODELAY` disables Nagle's algorithm, forcing every per-file flush to emit a separate TCP segment instead of allowing the kernel to coalesce small writes - this contributes to excess wire segments in daemon transfers
- Simplify `DaemonStream::configure_transfer_options()` by removing the now-unused `nodelay` parameter

## Changes

**`crates/daemon/src/daemon/sections/module_access/transfer.rs`** - Remove `stream.set_nodelay(true)?` call from `setup_transfer_streams()`

**`crates/core/src/client/remote/daemon_transfer/mod.rs`** - Remove `nodelay` argument from both `configure_transfer_options()` calls (direct daemon and daemon-over-remote-shell paths)

**`crates/core/src/client/module_list/connect/mod.rs`** - Remove `nodelay` parameter from `configure_transfer_options()` method signature and body

## Test plan

- [ ] CI passes (fmt, clippy, nextest, all platform matrices)
- [ ] Daemon interop tests pass against upstream rsync 3.0.9, 3.1.3, 3.4.1, 3.4.2
- [ ] Verify daemon pull/push performance improvement via benchmarks

### DIFF
--- a/crates/core/src/client/module_list/connect/mod.rs
+++ b/crates/core/src/client/module_list/connect/mod.rs
@@ -244,17 +244,15 @@ impl DaemonStream {
 
     /// Configures TCP-specific socket options for the transfer phase.
     ///
-    /// Sets TCP_NODELAY and applies read/write timeouts. No-op for
-    /// non-TCP transports (connect programs, TLS).
+    /// Applies read/write timeouts. No-op for non-TCP transports (connect
+    /// programs, TLS). TCP_NODELAY is intentionally not set - upstream rsync
+    /// does not set it on daemon connections (clientserver.c:1346 only sets
+    /// SO_KEEPALIVE).
     pub(crate) fn configure_transfer_options(
         &self,
-        nodelay: bool,
         timeout: Option<Duration>,
     ) -> io::Result<()> {
         if let Self::Tcp(stream) = self {
-            if nodelay {
-                stream.set_nodelay(true)?;
-            }
             stream.set_read_timeout(timeout)?;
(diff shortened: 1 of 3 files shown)


oferchen:
Closing: removing TCP_NODELAY causes significant regressions without first implementing upstream's `perform_io()` select-based write batching model.

**Daemon cold-start**: 3.2x upstream (was 1.04x) - Nagle delays small handshake/protocol packets
**Daemon concurrency soak**: 2010/3000 failed pulls - Nagle-induced timeouts under concurrent load

Upstream rsync doesn't set TCP_NODELAY explicitly, but its `perform_io()` model accumulates writes and flushes them as a batch before each blocking read via `select()`. Our separate read/write thread model doesn't have this natural batching - without TCP_NODELAY, individual small writes (NDX values, protocol frames) get delayed by Nagle's 200ms timer.

The correct approach would be implementing a `perform_io()`-style batched I/O model first, then removing TCP_NODELAY. That's a deeper architectural change tracked under BPR-X.

PR #5535, opened 2026-06-10
fix: guard zlib obuf overflow on matched-block inserts (upstream #951)

oferchen:
## Summary

- Drains the entire 64 KiB matched-block insert through the deflate stream in `ZlibTokenEncoder::see_token`, instead of issuing a single `compress()` call that could drop the unconsumed tail when the output buffer fills under pending sync-flush pressure.
- Sizes the dictionary-sync output buffer for the worst-case combined output (`AVAIL_OUT_SIZE(0xFFFF) + 16`) so a maximum-size stored block plus the `Z_SYNC_FLUSH` trailer always fits in a single pass.
- Hard-errors if the compressor ever stalls with the worst-case obuf reserve already in place, replacing a silent dictionary-desync with a loud failure.

## Rationale

Mirrors upstream rsync issue #951 against `token.c:send_deflated_token()`. The upstream `obuf` is sized for `AVAIL_OUT_SIZE(CHUNK_SIZE) = 32 816` bytes, but the matched-block insert path can deflate up to `0xFFFF` bytes per call while still carrying pending output from a preceding `Z_SYNC_FLUSH`. The corresponding flate2-based path inherited the same under-sized reserve and additionally never looped on partial consumption, so the encoder's dictionary silently diverged from the decoder's whenever a basis match landed at the obuf boundary.

The protocol < 31 data-duplicating bug is preserved so wire behaviour stays bit-compatible with legacy peers.

## Test plan

- [ ] CI fmt + clippy.
- [ ] CI nextest (stable) - covers the new regression tests `see_token_drains_64kib_insert_with_pending_flush_output` and `see_token_handles_exact_max_stored_block`.
- [ ] CI Windows, macOS, Linux musl stable jobs.
- [ ] CI interop validation.

### DIFF
--- a/crates/protocol/src/wire/compressed_token/tests.rs
+++ b/crates/protocol/src/wire/compressed_token/tests.rs
@@ -1749,8 +1749,11 @@ fn see_token_drains_64kib_insert_with_pending_flush_output() {
     let mut block = vec![0u8; 0xFFFF];
     for (i, b) in block.iter_mut().enumerate() {
         // LCG-style fill - avoids pulling in a PRNG dependency.
-        *b = (((i as u64).wrapping_mul(2862933555777941757).wrapping_add(3037000493) >> 32) & 0xFF)
-            as u8;
+        *b = (((i as u64)
+            .wrapping_mul(2862933555777941757)
+            .wrapping_add(3037000493)
+            >> 32)
+            & 0xFF) as u8;
     }
 
     // Highly compressible literal that forces several sync flushes through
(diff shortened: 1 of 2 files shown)


oferchen:
Superseded by #5520 (UTS Stream-C: "compress-zlib-insert (UTS-18): ZlibTokenEncoder::see_token() called compress(..., Z_SYNC_FLUSH) exactly once per 0xFFFF chunk"). PR #5520 (merged 2026-06-10) already addresses the same upstream rsync#951 root cause in crates/protocol/src/wire/compressed_token/zlib_codec.rs by draining the Sync flush in a loop until input is consumed and the output buffer is partially filled. The fix on master is functionally equivalent to this PR's drain loop; reimplementing the obuf reserve constants over a merged fix would be a competing rewrite rather than a surgical change. Closing as superseded.

PR #5536, opened 2026-06-10
fix: emit NDX_DEL_STATS in daemon-receive goodbye for --delete pushes

oferchen:
## Summary

- Adds `NDX_DEL_STATS` emission to the receiver's goodbye handshake so daemon-receive (push with `--delete --stats`) transfers correctly deliver per-type deletion counters back to the sender.
- Mirrors upstream's generator path: protocol 31+ gates plus the existing early/late `do_stats && delete_mode` conditions used by `crate::generator`'s `should_send_del_stats`.
- Stashes the receiver's delete stats from `run_pipelined` so they are available when `handle_goodbye` writes the final NDX frames.

## Upstream Reference

- `generator.c:2376-2381` early `write_del_stats()` path
- `generator.c:2420-2425` late `write_del_stats()` path
- `main.c:225-238` `write_del_stats()` wire format
- `rsync.c:337-342` `read_ndx_and_attrs()` consumes `NDX_DEL_STATS`

## Test plan

- [x] Added `goodbye_del_stats_tests.rs` covering:
  - Protocol 31 emits `NDX_DEL_STATS` + 5 varints + `NDX_DONE`, then echo `NDX_DONE`.
  - Protocol 31 suppresses emission when `do_stats=false`.
  - Late delete (`--delete-delay` / `--delete-after`) emits on `do_stats` alone.
  - Protocol 30 never emits the message (upstream protocol guard).
  - Varint field order matches upstream `main.c:225-238`.
- [ ] CI: fmt+clippy, nextest (stable), Windows, macOS, Linux musl, interop.

### DIFF
--- a/crates/protocol/src/stats/delete.rs
+++ b/crates/protocol/src/stats/delete.rs
@@ -80,7 +80,7 @@ impl DeleteStats {
     /// # Errors
     ///
     /// Returns an error if writing to the stream fails.
-    pub fn write_to<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+    pub fn write_to<W: Write + ?Sized>(&self, writer: &mut W) -> io::Result<()> {
         use crate::varint::write_varint;
 
         write_varint(writer, self.files as i32)?;
--- a/crates/transfer/src/receiver/mod.rs
+++ b/crates/transfer/src/receiver/mod.rs
@@ -52,6 +52,7 @@ use protocol::acl::AclCache;
 use protocol::filters::FilterRuleWireFormat;
 use protocol::flist::{FileEntry, FileListReader};
 use protocol::idlist::IdList;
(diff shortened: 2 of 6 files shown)


oferchen:
Superseded by #5520, which already addresses UTS-6 (#3569 daemon-delete-stats) at the correct layer. #5520 wires --stats into ServerConfig.do_stats so the existing generator-side goodbye emitter in crates/transfer/src/generator/transfer/goodbye.rs (should_send_del_stats + write_del_stats) fires for daemon-receive pushes. This PR added emission in the receiver's handle_goodbye, but the generator owns NDX_DEL_STATS emission in this codebase (the receiver-side phases.rs only consumes the sender-echoed frame). Same supersession pattern as #5535. Closing without merge.

PR #5564, opened 2026-06-10
fix(cli): enable --xattrs on Windows with feature flag

oferchen:
## Summary

- Widen the CLI preflight cfg gate so `--xattrs` is accepted on Windows when built with `feature = "xattr"`, instead of being silently rejected before `metadata::xattr_windows` (FindFirstStreamW ADS impl) can run. Mirrors the acl gate.
- Propagate `metadata/xattr` through the `core` crate's `xattr` feature so the metadata implementation is actually compiled in, mirroring the `acl` wiring at the adjacent line.
- Add a Windows-only regression test (`cfg(target_os = "windows", feature = "xattr")`) that invokes the preflight check on `--xattrs` and asserts no "extended attributes are not supported on this client" rejection.

## Test plan

- [ ] Windows nextest cell: new `xattrs_accepted_on_windows_with_feature` preflight test passes.
- [ ] No regression on Linux/macOS nextest cells (test is cfg-gated to Windows; existing Unix path unchanged).
- [ ] Required CI cells green: fmt+clippy, nextest (stable), Windows (stable), macOS (stable), Linux musl (stable).

### DIFF
--- a/crates/cli/src/frontend/execution/drive/workflow/preflight.rs
+++ b/crates/cli/src/frontend/execution/drive/workflow/preflight.rs
@@ -18,7 +18,7 @@ use super::super::super::{
 };
 #[cfg(any(
     not(all(any(unix, windows), feature = "acl")),
-    not(all(unix, feature = "xattr"))
+    not(all(any(unix, windows), feature = "xattr"))
 ))]
 use super::super::messages::fail_with_custom_fallback;
 use super::super::messages::fail_with_message;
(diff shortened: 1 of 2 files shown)


oferchen:
Closing as fully redundant with #5625, which landed on master and explicitly supersedes this PR.

#5625 already shipped:
- The exact same cfg widening from `not(all(unix, feature = "xattr"))` to `not(all(any(unix, windows), feature = "xattr"))` on `validate_feature_support` in `preflight.rs`.
- The matching companion gate on `fail_with_custom_fallback` in `messages.rs` that this PR was missing (and which broke its Windows CI cells with `function 'fail_with_custom_fallback' is never used`).
- The `xattr = ["metadata/xattr", "transfer/xattr"]` propagation in `crates/core/Cargo.toml`.
- A Windows-only end-to-end preflight regression test (`xattrs_preflight_accepts_on_windows_with_feature` in `crates/cli/src/frontend/tests/xattrs.rs`).

Rebasing this branch on master produces a no-op diff. Per fresh-audit-over-session-memory, closing rather than rebasing an empty PR. The broader parameterized (platform × feature) matrix lives in #5576.

PR #5568, opened 2026-06-10
fix(daemon): validate alt-basis dest args against module containment

oferchen:
## Summary

- **Defense-in-depth lexical pre-check.** Extends `validate_client_paths_in_module()` to reject client-supplied `--compare-dest` / `--copy-dest` / `--link-dest` paths that resolve outside the module root, before the daemon ever attempts to open the basis directory. Sits in front of the URV-5.a `DirSandbox` open-time enforcement: a client passing `--copy-dest=/etc` to a non-chroot daemon is rejected at the wire-protocol layer with `@ERROR: --copy-dest path '/etc' is outside module root` rather than relying on the per-open sandbox to catch it.
- **Mirrors the existing temp-dir / partial-dir / backup-dir pattern.** Same relative-path-passes / absolute-path-canonicalizes-then-rejects logic, same `@ERROR` payload shape, same log line, same `Ok(false)` return contract. The recognised flag list is consolidated into a `CONTAINED_PATH_FLAGS` const and the scan is split into a pure `find_outside_module_path()` helper so the rejection logic can be unit-tested without the full `ModuleRequestContext` stack.
- **Both short and long-form arg spellings.** Catches `--compare-dest` / `--copy-dest` / `--link-dest` and their `--compare-destination` / `--copy-destination` / `--link-destination` synonyms in both `--flag=value` and `--flag value` wire forms.
- **Regression coverage** for negative case (`--copy-dest=/etc` rejected), positive case (`--copy-dest=alt-basis` accepted under module root), the long-form spelling family, and the separate-arg wire form (`--link-dest /etc`). SECURITY.md gains a URV-5.b.1 bullet documenting the two-layer defense.

## Test plan

- [ ] CI: fmt + clippy clean on the new helper.
- [ ] CI: nextest (stable) passes the four new tests in `crates/daemon/src/daemon/sections/module_access/tests.rs`:
  - [ ] `find_outside_module_path_rejects_alt_basis_absolute_escape`
  - [ ] `find_outside_module_path_accepts_alt_basis_relative_under_module`
  - [ ] `find_outside_module_path_rejects_alt_basis_long_form_spelling`
  - [ ] `find_outside_module_path_separate_arg_form_is_recognised`
- [ ] CI: Windows + macOS + Linux musl matrices pass (the new helper is pure-Rust path canonicalisation; no platform gating needed).
- [ ] CI: existing interop tests still pass - legitimate in-module alt-basis paths must continue to work.


oferchen:
Rebase audit finding: this PR is fully superseded by #5601 (URV-5.b.REOPEN) which merged at cf549a4e8 on master.

Master's `validate_client_paths_in_module()` already covers `--compare-dest` / `--copy-dest` / `--link-dest` in both `--flag=value` and `--flag value` wire forms, with `classify_client_path_against_module()` extracted as a pure helper. The merged form additionally widens `engage_landlock_sandbox` with `extra_allowed_paths` so the Landlock allowlist tracks the validated alt-basis paths (not in this PR).

Force-rebasing this branch onto current master would regress more than it adds, because the PR's `transfer.rs` predates the URV-5.b.REOPEN allowlist widening and the LSM-seccomp wiring that landed in #5589 and #5601.

Residual-gap check for the long-form spellings (`--compare-destination` / `--copy-destination` / `--link-destination`): these don't exist in upstream rsync's `options.c` poptOption table — only the short forms are recognised on the wire. So there is no production gap to refocus on.

Closing this PR as duplicate of #5601. The defense-in-depth lexical pre-check intent is fully realised on master today.

oferchen:
Closing as duplicate of #5601 per overlap audit above.

PR #5675, opened 2026-06-11
fix: follow symlink at dest_root to match upstream main.c:745-754

oferchen:
## Summary

The receiver-side `ensure_dest_root_exists()` pre-flight used `symlink_metadata()` (lstat) and explicitly rejected any symlinked destination with `InvalidInput`. Upstream `main.c:745-754` (`get_local_name()`) instead calls `do_stat()` which follows symlinks, accepts the destination when `S_ISDIR(st.st_mode)` is true, and then enters the resolved directory via `change_dir(dest_path, CD_NORMAL)`. The upstream `symlink-dirlink-basis` interop test (issue #715) drives exactly this code path with `\$HOME/dir -> \$HOME/real-dir` and `-K (--copy-dirlinks)`.

This change brings the helper in line with upstream:

- Switch to `metadata()` so a symlink-to-existing-directory dest is accepted.
- Keep dangling symlinks safe: when `metadata()` reports `NotFound`, do an `lstat` follow-up and surface `NotFound` instead of letting `create_dir_all` resolve the link and materialize a directory at the link target.
- Canonicalize the symlinked dest in `transfer::receiver::transfer::setup::setup_transfer` so the downstream `DirSandbox::open_root` opens the resolved target dirfd; per-entry `*at` syscalls then anchor at the canonical path.
- Skip the canonicalize step under daemon connections so `open_sandbox_for_dest_strict` keeps refusing symlinked dests via `ELOOP`/`ENOTDIR` (the chdir-symlink-race defense).

## Security model

Per-entry writes are sandboxed by the SEC-1 \`*at\` chain (see \`crates/fast_io/src/dir_sandbox/mod.rs:150\` \`secure_open_dir\` and \`crates/transfer/src/receiver/transfer/setup.rs:310\` \`open_sandbox_for_dest_strict\`). Opens route through \`openat2(RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS)\` on Linux 5.6+ and \`openat(O_NOFOLLOW | O_DIRECTORY)\` elsewhere. Accepting a symlinked dest at the pre-flight does not weaken containment because the canonical resolved path becomes the sandbox root and every subsequent open is anchored at that resolved fd. Daemon connections keep the strict refusal at the sandbox layer.

The chdir-symlink-race tests (UTS-16) live in \`crates/transfer/src/receiver/transfer/setup.rs::symlink_race_tests\` and target \`open_sandbox_for_dest_strict\` (per-entry symlink swap, not dest-root). They test a different attack surface and remain unchanged.

## Test changes

- \`crates/transfer/tests/uts_2_dest_root_mkdir.rs\`:
  - **Updated**: \`refuses_broken_symlink_to_outside_module\` -> \`rejects_broken_symlink_dest_root\` now asserts \`NotFound\` (was \`InvalidInput\`).
  - **Removed and replaced**: \`refuses_broken_symlink_to_intra_module_target\`, \`refuses_symlink_to_existing_intra_module_directory\`, and \`refuses_symlink_to_existing_directory_outside_module\` are superseded by \`accepts_symlink_to_existing_directory\` and \`accepts_symlink_to_existing_directory_outside_parent\`, which pin the new accept-on-symlink-to-dir behavior.
  - **Removed**: \`existing_non_directory_path_is_noop_for_helper\` (the helper now errors on non-directory dests, matching upstream \`main.c:756-761\`); replaced by \`rejects_existing_non_directory_at_dest_root\`.
  - **Added**: \`rejects_chained_dangling_symlink_dest_root\` to pin the dangling-symlink fallback for a chained \`link_a -> link_b -> nothing\` shape.

## Test plan
- [ ] CI fmt+clippy passes
- [ ] CI nextest (stable) passes
- [ ] CI Windows, macOS, Linux musl pass
- [ ] Interop test \`symlink-dirlink-basis\` reaches test 7 without the \`real-dir/file\` FileNotFoundError

### DIFF
--- a/crates/transfer/src/receiver/mod.rs
+++ b/crates/transfer/src/receiver/mod.rs
@@ -174,24 +174,45 @@ pub(in crate::receiver) fn dest_arg_has_trailing_slash(arg: &OsStr) -> bool {
 /// pre-flight was a no-op (already exists, single-file transfer without
 /// trailing slash, or `dry_run`).
 ///
-/// # Symlink refusal
+/// # Symlink resolution
 ///
-/// The existence check uses `symlink_metadata()` (lstat) rather than
-/// `metadata()` (stat) so a symlink at `dest_root` is detected directly
-/// instead of being silently resolved. When `dest_root` is itself a symlink -
-/// broken or otherwise - the helper refuses with `InvalidInput` instead of
-/// proceeding. A broken symlink would fall through the stat-based check
-/// (NotFound at the target) and let `create_dir_all` resolve through the
-/// symlink, materializing the directory outside the intended location; a
-/// dangling-link-class containment bypass analogous to the SEC-1 TOCTOU
-/// family. An existing-symlink dest is equally suspect because every
-/// per-entry write would land at the symlink target, sidestepping the
-/// daemon's `module.path` containment. Operators that genuinely need to
-/// receive into a symlinked location must materialize the real directory
-/// themselves; oc-rsync never auto-creates through a symlink.
+/// The existence check uses `metadata()` (stat) rather than
+/// `symlink_metadata()` (lstat) so a symlink at `dest_root` pointing at a
+/// real directory is accepted, matching upstream `main.c:745-754`
+/// `get_local_name()` which calls `do_stat()` (follows symlinks) and
+/// proceeds when `S_ISDIR(st.st_mode)` is true. A symlinked dest root is
+/// the upstream `symlink-dirlink-basis` interop scenario (issue #715).
+///
+/// A dangling symlink resolves to `ENOENT`, and the helper would then call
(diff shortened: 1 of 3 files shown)


oferchen:
Auto-rebase blocked: substantive conflict in `crates/transfer/tests/uts_2_dest_root_mkdir.rs` between this branch's `accepts_symlink_to_existing_directory_outside_parent` (mirrors upstream `main.c:745-754` which follows the symlink) and master's newly-landed `refuses_symlink_to_existing_directory_outside_module` (defense-in-depth refusal at the pre-flight layer). Both can't coexist; needs author decision on policy + rebase. SEC-1 *at chain still anchors per-entry writes regardless.

PR #5682, opened 2026-06-11
fix(transfer): remove dead `mut` on alt-dest test wire_capture

oferchen:
Master fmt+clippy fails on `-D unused_mut` because `wire_capture` in `uts_2_efg_ssh_alt_dest_verification.rs:195` is constructed and never written. Drop the `mut`.

## Test plan
- [x] Visual inspection: `get_ref()` is the only read; no `get_mut()` or write call.
- [x] CI fmt+clippy lane on this PR.


oferchen:
Empty PR after rebase — the dead `mut` was already removed on master via a different path. Closing.

PR #5742, opened 2026-06-12
fix: whole-file flag parity + SO_LINGER daemon teardown

oferchen:
## Summary
- Use `whole_file_raw() == Some(true)` instead of `whole_file()` in server flag builder - upstream only sends `-W` when explicitly set (`whole_file > 0`), not when defaulting to `-1`
- Replace 50ms sleep with SO_LINGER(5s) on daemon TCP sockets to prevent RST from aborting goodbye bytes under concurrent connections
- Revert filter anchoring change back to original logic - the `slash_count > 0` over-anchored patterns like `foo/too/` that need tail-matching

## Test plan
- [ ] Upstream testsuite `-j8` with OC_RSYNC_NO_SECCOMP=1 shows no regressions vs master baseline
- [ ] CI passes on all required checks

### DIFF
--- a/crates/core/src/client/remote/flags.rs
+++ b/crates/core/src/client/remote/flags.rs
@@ -93,9 +93,11 @@ pub(crate) fn build_server_flag_string(config: &ClientConfig) -> String {
     if effective_dirs && !effective_recursive {
         flags.push('d');
     }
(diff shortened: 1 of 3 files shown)


oferchen:
Closing: master already has whole_file_raw() and anchoring fixes. SO_LINGER change will be rebased into a clean PR.

PR #5809, opened 2026-06-15
fix(transfer): scan dotdir entries in --files-from marker-dir rescan

oferchen:
## Summary

- `build_file_list_with_base` gated the SLASH_ENDING_NAME/DOTDIR_NAME marker-dir rescan on `entry.path != entry.base`, but DOTDIR entries (`from/./`, `from/.`) split with `path == base`. With `--files-from` clearing the global `recurse` flag (upstream `options.c:2189`), `walk_path_with_metadata` emits only the root entry, so the rescan is the only path that injects children - and the gate was silently swallowing it. Transfers driven by `from/./` collapsed to zero files, breaking the upstream `files-from.test`.
- Drop the `path != base` condition so any entry with `recurse == true` (matching upstream `flist.c:2329`) triggers `scan_files_from_marker_dir`. SLASH_ENDING_NAME entries (`path != base`) continue to work as before; DOTDIR entries now also recurse, mirroring upstream semantics exactly.

## Test plan

- [x] New `build_file_list_with_base_dotdir_entry_scans_children` regression test added under `crates/transfer/src/generator/tests.rs::files_from`. It builds a fixture with a `from/` directory containing two files, drives `build_file_list_with_base` with a DOTDIR `FilesFromEntry` (`base == path == from/`, `recurse = true`) and `flags.recursive = false`, and asserts both child filenames appear in the file list.
- [ ] CI: fmt+clippy, nextest (stable), Windows/macOS/Linux musl.
- [ ] Upstream `files-from.test` interop run.

### DIFF
--- a/crates/transfer/src/generator/file_list/mod.rs
+++ b/crates/transfer/src/generator/file_list/mod.rs
@@ -266,9 +266,12 @@ impl GeneratorContext {
             // so the trailing-slash directories would otherwise stop at the
             // entry itself. Re-scan the directory here so the receiver sees
             // the listed dir's contents (`from/./dir/subdir/subsubdir2/` must
-            // emit `bin-lt-list`, etc.). The `entry.path == entry.base` case
-            // is already covered by the DOTDIR scan in `walk_path_with_metadata`.
-            if entry.recurse && entry.path != entry.base {
+            // emit `bin-lt-list`, etc.). DOTDIR entries (`from/./` and
+            // `from/.`) produce `entry.path == entry.base`; they still need
+            // the rescan because `flags.recursive` is cleared whenever
+            // `--files-from` is active (upstream `options.c:2189`), so
+            // `walk_path_with_metadata` would emit only the root entry.
+            if entry.recurse {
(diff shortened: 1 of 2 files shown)


oferchen:
NACK — testsuite/files-from.test still diverges from upstream rsync on this branch.

Root cause confirmed: the `entry.path != entry.base` gate removal is correct, since with global `recurse` cleared by `--files-from` (upstream options.c:2189), DOTDIR rescans must run unconditionally to emit children. However, the interop CI failure ("interop with upstream rsync") shows divergence persists.

Remaining audit needed in `scan_files_from_marker_dir` (invoked post-fix):
1. Verify it handles `base == path` correctly, producing wire names relative to `base`, not filesystem root.
2. Confirm entry order/dedup vs upstream flist.c:998 'unrequested' check stays compatible.
3. Ensure rescan does not duplicate entries already emitted by `try_walk_source_entry_dedup` above.

The comment at line 269-270 claiming 'covered by walk_path_with_metadata' is misleading and should be replaced with the actual invariant the rescan enforces. PR needs another pass.

PR #5811, opened 2026-06-15
fix(engine): skip batch ID lists when INC_RECURSE is negotiated

oferchen:
## Summary

- Under `--write-batch` with CF_INC_RECURSE in the negotiated compat_flags, `write_batch_id_lists` unconditionally emitted two `varint30(0)` uid/gid terminators after the flist end marker. The matching reader at `crates/batch/src/reader/flist.rs:167` only consumes those bytes when `!inc_recurse`, so the two stray varints drifted the stream cursor: subsequent NDX reads decoded garbage and `--read-batch` replay applied zero files. This was the upstream rsync testsuite `batch-mode` failure on oc-rsync.
- Gate the writer on `BatchConfig::compat_flags` so the post-flist ID list region is omitted exactly when upstream omits it (`flist.c:2548 - if (numeric_ids <= 0 && !inc_recurse) send_id_lists(f)`). Under INC_RECURSE the uid/gid names already ride inline on each flist entry via `XMIT_USER_NAME_FOLLOWS` / `XMIT_GROUP_NAME_FOLLOWS`, so dropping the terminators leaves the stream byte-equivalent to upstream's batch output.
- Add a regression test that drives `LocalCopyPlan::execute_with_options` with a `BatchWriter` whose compat_flags include CF_INC_RECURSE, then reads the resulting batch file back and confirms the source payload entry decodes from the flist without the cursor drifting past the end-of-list marker.

## Test plan

- [ ] CI: fmt + clippy
- [ ] CI: nextest (stable) - including `crates/engine/tests/batch_id_lists_inc_recurse.rs`
- [ ] CI: Windows + macOS + Linux musl
- [ ] CI: upstream-testsuite `batch-mode`

### DIFF
--- a/crates/core/src/client/run/batch.rs
+++ b/crates/core/src/client/run/batch.rs
@@ -331,10 +331,34 @@ fn replay_batch(
             "Batch replay complete"
         );
     }
-    let _ = &result;
 
+    // upstream: main.c:362-373 - on the --read-batch side the receiver reads
+    // total_read / total_written / stats.total_size from the batch trailer and
+    // surfaces them through output_summary(). Mirror that by populating the
+    // ClientSummary so the "sent X bytes received X bytes" / "total size is X"
+    // lines reflect the replayed payload instead of zeros.
+    //
+    // The replay engine accounts every flist entry against `file_count` and
+    // every byte of source-side material against `total_size`. Symlinks and
+    // dirs created during replay are counted as files_transferred because the
+    // receiver materialised them at the destination, matching upstream's
+    // num_files / num_transferred accounting under --read-batch.
     use engine::local_copy::LocalCopySummary;
-    Ok(ClientSummary::from_summary(LocalCopySummary::default()))
+    let files_listed = usize::try_from(result.file_count).unwrap_or(usize::MAX);
+    let files_transferred = files_listed;
+    let total_size = result.total_size;
+    let summary = LocalCopySummary::from_receiver_stats(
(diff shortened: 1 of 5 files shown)


oferchen:
NACK — this PR's gate doesn't actually fix `testsuite/batch-mode.test`. CI still shows `FAIL batch-mode (exit 1)` on this branch.

Root cause: the PR checks `compat_flags` from the engine's BatchWriter config (defaults to `Some(0)` for local transfers, since protocol negotiation doesn't happen at engine level). Upstream rsync DOES negotiate INC_RECURSE for local `-av --write-batch` transfers, but oc-rsync's BatchWriter never sees that negotiated value. The new gate `if compat_flags has INC_RECURSE bit` therefore never triggers in the failing scenario.

Possible reworks: (a) plumb negotiated compat_flags from protocol layer down to BatchWriter; (b) skip ID lists at engine-layer for local batch ops; (c) match upstream by negotiating INC_RECURSE during local batch transfers. Pick one and verify locally with `bash testsuite/batch-mode.test` before re-pushing.

oferchen:

This PR claims to fix upstream testsuite test(s): `batch-mode`.

Latest upstream testsuite run on this PR's head: https://github.com/oferchen/rsync/actions/runs/27635572535/job/81721748047

```
upstream testsuite	Run upstream testsuite	2026-06-16T17:40:12.7757552Z FAIL    batch-mode  (exit 1)
```

PR #5813, opened 2026-06-15
fix(transfer): allow default ACL inheritance when creating destination files

oferchen:
## Summary

- Folds the destination directory's POSIX default ACL into `dest_mode()`, mirroring upstream `default_perms_for_dir(dn)` at `generator.c:1339` and `receiver.c:847`.
- Without this, files created under a permissive default ACL (e.g. `u::6,g::6,o:6`) silently landed at `0o600` because `dflt_perms` was being derived only from the umask.
- All four `!preserve_perms` `dest_mode` call sites now route through a single `dflt_perms_for_destination` helper. On platforms without POSIX default ACL support, the helper falls back to the umask-derived default, matching upstream's `#ifdef SUPPORT_ACLS` gating.

## Test plan

- [x] Adds `dest_mode_inherits_default_acl_bits_under_no_perms` regression test gated on `feature = "acl"` + `linux|freebsd`. Installs a `u::6,g::6,o:6` default ACL on the parent dir, runs `apply_file_metadata_with_options` with `--no-perms` and `destination_is_new(true)`, and asserts the dest file lands at `0o666`. Mirrors `testsuite/acls-default.test` (`da777` case).
- [x] `cargo fmt --all` clean.
- [x] Test self-skips when the filesystem doesn't support POSIX default ACLs (e.g. tmpfs without `acl` mount), matching upstream `test_skipped`.
- [ ] CI: fmt+clippy, nextest (stable), Windows, macOS, Linux musl.

### DIFF
--- a/crates/metadata/src/apply/permissions.rs
+++ b/crates/metadata/src/apply/permissions.rs
@@ -40,20 +40,51 @@ fn cached_umask() -> u32 {
     })
 }
 
+/// Returns the default permission bits a child of `destination`'s parent
+/// directory should inherit.
+///
+/// When the parent has a POSIX default ACL, its `user_obj`/`group_obj`/`other`
+/// bits override the umask-derived default (`ACCESSPERMS & ~orig_umask`).
+/// Mirrors upstream's `default_perms_for_dir(dn)` call at
+/// `generator.c:1339` and `receiver.c:847` so `dest_mode()` folds in the
+/// inherited bits when `!preserve_perms`. Without this, a permissive default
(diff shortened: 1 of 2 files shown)


oferchen:
Closing as superseded by #5841 which shipped a more complete fix including ACL mask handling for the `da750mask` scenario in upstream `testsuite/acls-default.test`. This PR's `dflt_perms_for_destination` ignores mask entries, causing 0o770 instead of 0o750 — that's why `acls-default` still failed in this PR's CI.

PR #5815, opened 2026-06-15
fix(transfer): emit created-dir notice and root row under --itemize-changes

oferchen:
## Summary

Closes the upstream `testsuite/itemize.test` divergence in local-mode `oc-rsync -iplrtc from/ to/`. Prior commits on this branch fixed the first two sub-assertions (initial transfer and `-iplrH` re-sync). This update closes the third sub-assertion by emitting the two itemize lines that were still missing:

- `.d..t...... foo/` for the existing destination directory whose mtime drifted from the source. Upstream `generator.c:1480-1483` calls `itemize()` with `iflags=0` for every directory in the flist (including the existing ones) and lets `itemize()` OR in `ITEM_REPORT_TIME|PERMS|...` based on the existing-vs-source metadata drift; the row is suppressed via `SIGNIFICANT_ITEM_FLAGS` only when nothing differs. oc-rsync's local-copy executor was treating "destination already exists" as "nothing to itemize" and skipped the row entirely.
- `cLc.t...... foo/sym -> ../bar/baz/rsync` for the recreated symlink. Upstream `generator.c:1605-1609` itemizes the recreate with `ITEM_LOCAL_CHANGE|ITEM_REPORT_CHANGE` against the OLD symlink's `sx.st`, lighting up the `c` glyph at position 2 and the `t` glyph at position 4 via `mtime_differs()`. oc-rsync emitted a `SymlinkCopied` record with an empty change set, collapsing positions 2-10 to spaces (`cL          foo/sym ...`).

### Implementation

- `change_set/detection.rs`: add `LocalCopyChangeSet::for_existing_directory` to compute attribute drift between source and destination directories (mtime gated by `!omit_dir_times`, perms/owner/group; ACL/xattr bits left to the caller because the local-copy fast path does not yet thread that comparison through).
- `change_set/detection.rs`: add `LocalCopyChangeSet::for_recreated_symlink` mirroring upstream's `ITEM_LOCAL_CHANGE|ITEM_REPORT_CHANGE` by setting `checksum_changed=true` and adding the `Modified` time marker when symlink mtimes drift (gated by `!omit_link_times`).
- `directory/recursive/destination.rs`: extend `DestinationState::Ready` to carry the existing directory metadata so the recursive walker can diff it against the source.
- `directory/recursive/mod.rs`: when the destination dir already exists and any attribute differs, emit a `MetadataReused` record ahead of child entries. When nothing differs the record stays unemitted, matching upstream's `SIGNIFICANT_ITEM_FLAGS` gate.
- `special/symlink.rs`: capture the existing symlink metadata before removal, attach the recreated-symlink change set to the `SymlinkCopied` record on the recreate path, and flip `was_created` to `true` only when the dest did not previously exist or was a non-symlink replacement (mirroring upstream's `statret = -1` fallback at `generator.c:1606-1607`).
- `change_set/tests.rs`: regression coverage for the new constructors (mtime match/mismatch, `omit_dir_times`/`omit_link_times` suppression).

After this change, the upstream `itemize.test` third-assertion output matches the golden:

```
.f..tp..... bar/baz/rsync
.d..t...... foo/
.f..t...... foo/config1
>fcstp..... foo/config2
cLc.t...... foo/sym -> ../bar/baz/rsync
```

## Test plan

- [ ] CI: full nextest matrix (stable / Windows / macOS / Linux musl)
- [ ] CI: Upstream Testsuite (`itemize.test` golden)
- [ ] CI: Interop Validation

### DIFF
--- a/crates/cli/src/frontend/out_format/render/placeholder.rs
+++ b/crates/cli/src/frontend/out_format/render/placeholder.rs
@@ -35,7 +35,16 @@ pub(super) fn render_placeholder_value(
                 .metadata()
                 .and_then(ClientEntryMetadata::symlink_target)
             {
-                rendered.push_str(" -> ");
+                // upstream: log.c:643-654 - `%L` renders ` => %s` for
+                // hard-link references and ` -> %s` for symlink targets.
+                // The two cases are mutually exclusive on a single record
+                // because a hardlink alias never carries a symlink target.
+                let connector = if matches!(event.kind(), ClientEventKind::HardLink) {
+                    " => "
+                } else {
+                    " -> "
+                };
+                rendered.push_str(connector);
                 rendered.push_str(&target.to_string_lossy());
             }
             Some(rendered)
@@ -68,7 +77,14 @@ pub(super) fn render_placeholder_value(
             .metadata()
             .and_then(ClientEntryMetadata::symlink_target)
             .map(|target| {
-                let mut rendered = String::from(" -> ");
+                // upstream: log.c:643-654 - `%L` standalone also renders the
+                // ` => %s` vs ` -> %s` distinction.
+                let connector = if matches!(event.kind(), ClientEventKind::HardLink) {
+                    " => "
+                } else {
+                    " -> "
+                };
+                let mut rendered = String::from(connector);
                 rendered.push_str(&target.to_string_lossy());
                 rendered
             }),
--- a/crates/cli/src/frontend/progress/render.rs
+++ b/crates/cli/src/frontend/progress/render.rs
@@ -1,6 +1,23 @@
 use std::io::{self, Write};
+use std::path::Path;
 
 use core::client::{ClientEvent, ClientEventKind, ClientSummary, HumanReadableMode};
+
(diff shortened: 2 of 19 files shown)


oferchen:
NACK — testsuite/itemize.test still shows FAIL on this branch.

Root cause: logic-order bug. The check for `root_was_just_created` happens BEFORE `mark_destination_root_created()` is called, so the root-row itemize emission misfires on first-run vs subsequent files within the same transfer.

Precise fix: reorder `mark_destination_root_created()` to fire BEFORE the `root_was_just_created` predicate is read. The mark must reflect 'root has been created this transfer' state when the predicate runs for the very first file, otherwise the implicit cd+++++++++ ./ row is suppressed.

oferchen:

This PR claims to fix upstream testsuite test(s): `itemize`.

Latest upstream testsuite run on this PR's head: https://github.com/oferchen/rsync/actions/runs/27635566251/job/81721725617

```
upstream testsuite	Run upstream testsuite	2026-06-16T17:42:28.2336938Z FAIL    itemize  (exit 1)
```

PR #5817, opened 2026-06-15
fix(filters): wire :C cvs_mode + default empty pattern to .cvsignore

oferchen:
## Summary

Closes the upstream `exclude` / `exclude-lsh` testsuite failures triggered by `-f:C -f-C --delete-excluded` and a per-directory `mid/.filt` containing a bare `:C` line. The original PR title (filter scope fall-through) described an intermediate symptom; the actual root causes are in the `:C` modifier parse path and the wire-rule -> `DirMergeConfig` conversion. This PR is honest about that scope.

This PR does **not** touch `--files-from`. That was lumped in incorrectly with the previous title; if there is a real `--files-from` issue it will land separately.

## What this PR does

- **Per-directory `.filt` containing `:C`** (now): parser falls through to `.cvsignore` instead of erroring as "unrecognized filter rule", and `RuleModifiers::apply` propagates `cvs_mode` onto the resulting `FilterRule`. Mirrors upstream `exclude.c:1324` (allow empty pattern when `FILTRULE_CVS_IGNORE` is set) and `exclude.c:1404-1408` (default to `.cvsignore`). [PR #5842 already landed this piece on this branch; this PR completes the wire side.]
- **`-f:C` over remote shell** (now): the CLI parser already maps `:C` to a `DirMerge` spec with the right semantics, but `build_wire_format_rules` had no way to forward the `C` modifier because `DirMergeOptions` did not track it. Add a `cvs_mode` flag to `DirMergeOptions`, set it from the CLI `parse_merge_modifiers` `C` branch, and forward it as `cvs_exclude=true` on the wire so the remote sender sees the full `:C` rule.
- **Wire rule with `cvs_exclude=true` and empty pattern** (now): `wire_rule_to_dir_merge_config` substitutes `.cvsignore` as the default filename instead of producing a `DirMergeConfig` with an empty filename that silently disables per-directory `.cvsignore` lookup. Mirrors upstream `exclude.c:1404-1408`.

## Upstream references

- `rsync-3.4.4/exclude.c:1248-1255` - `C` modifier sets `FILTRULE_NO_PREFIXES | FILTRULE_WORD_SPLIT | FILTRULE_NO_INHERIT | FILTRULE_CVS_IGNORE`.
- `rsync-3.4.4/exclude.c:1324` - empty pattern allowed only when `FILTRULE_CVS_IGNORE` is set.
- `rsync-3.4.4/exclude.c:1404-1408` - merge/dir-merge with empty pattern defaults to `.cvsignore`.
- `rsync-3.4.4/testsuite/exclude.test:86-88` (and `exclude-lsh.test` via symlink) - failing fixture: `mid/.filt` contains a bare `:C` line; the wrapping run drives `-f dir-merge_.filt -f:C -f-C`.

## Scope explicitly NOT in this PR

- `--files-from` (will be addressed in a separate PR if a real regression is identified).
- The non-inheriting outer-scope fall-through commit from the earlier history of this branch (`869c796b`) is preserved; the new commits stack on top of it.

## Test plan

- [x] `cargo fmt --all -- --check`
- [ ] CI: fmt+clippy, nextest (stable), Windows, macOS, Linux musl
- [ ] CI: upstream testsuite `exclude` and `exclude-lsh`
- [x] New regression tests:
  - `wire_rule_to_dir_merge_config_empty_pattern_defaults_to_cvsignore` (`crates/transfer/src/generator/filters.rs`) - empty wire pattern with `cvs_exclude=true` -> `.cvsignore` filename.
  - `dir_merge_cvs_mode_forwards_cvs_exclude_flag` (`crates/core/src/client/remote/flags.rs`) - locally parsed `:C` DirMerge round-trips `cvs_exclude=true` to the wire.
  - `dir_merge_options_cvs_mode_round_trip` (`crates/engine/src/local_copy/filter_program/options.rs`) - `DirMergeOptions::cvs_mode` accessor pair is stable.
  - `parse_merge_modifiers_cvsignore` augmented with `is_cvs_mode()` assertion (`crates/cli/src/frontend/filter_rules/parsing/merge.rs`).
- [x] PR #5842's regression tests for the bare `:C` parse path remain in place (`crates/filters/src/merge/tests.rs`).

### DIFF
--- a/crates/cli/src/frontend/filter_rules/parsing/merge.rs
+++ b/crates/cli/src/frontend/filter_rules/parsing/merge.rs
@@ -66,7 +66,8 @@ pub(crate) fn parse_merge_modifiers(
                     .use_whitespace()
                     .allow_comments(false)
                     .allow_list_clearing(true)
-                    .inherit(false);
+                    .inherit(false)
+                    .cvs_mode(true);
                 assume_cvsignore = true;
             }
             'e' => {
@@ -232,6 +233,9 @@ mod tests {
         assert_eq!(options.enforced_kind(), Some(DirMergeEnforcedKind::Exclude));
         assert!(options.uses_whitespace());
         assert!(!options.allows_comments());
+        // upstream: exclude.c:1248-1254 - `C` modifier sets FILTRULE_CVS_IGNORE.
+        // We record this so the wire encoder can forward it as cvs_exclude=true.
+        assert!(options.is_cvs_mode());
     }
 
     #[test]
--- a/crates/core/src/client/remote/flags.rs
+++ b/crates/core/src/client/remote/flags.rs
@@ -203,6 +203,13 @@ pub(crate) fn build_wire_format_rules(
             wire_rule.no_inherit = !options.inherit_rules();
             wire_rule.word_split = options.uses_whitespace();
             wire_rule.exclude_from_merge = options.excludes_self();
+            // upstream: exclude.c:1248-1254 - the `C` modifier on a dir-merge
+            // rule sets FILTRULE_CVS_IGNORE on the wire. Without this, `-f:C`
+            // would round-trip through the remote shell as a bare dir-merge
+            // missing the CVS flag, so the remote sender could neither
+            // default the empty pattern to `.cvsignore` nor activate
+            // CVS-style whitespace parsing of the per-directory file.
+            wire_rule.cvs_exclude = options.is_cvs_mode();
         }
 
         wire_rules.push(wire_rule);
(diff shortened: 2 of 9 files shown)


oferchen:
NACK — testsuite/exclude.test and exclude-lsh.test still show FAIL on this branch.

Root cause: `scope_has_deletion_match()` calls `filter_set.has_deletion_rule_match(path, is_dir)` without enabling descendant-check, while the Transfer arm correctly skips descendants. For Deletion context, descendants MUST be checked (an excluded directory's descendants are also excluded for deletion purposes).

Precise fix: in `crates/filters/src/set.rs` `has_deletion_rule_match`, the call to `self.inner.has_matching_rule` needs to pass `check_descendants=true` (or equivalent) — Deletion's semantic is opposite to Transfer's for descendants. Confirm by re-running exclude.test on the branch after the change.

oferchen:

This PR claims to fix upstream testsuite test(s): `exclude-lsh,exclude,files-from`.

Latest upstream testsuite run on this PR's head: https://github.com/oferchen/rsync/actions/runs/27635559609/job/81721703883

```
upstream testsuite	Run upstream testsuite	2026-06-16T17:38:39.9510960Z FAIL    exclude-lsh  (exit 1)
upstream testsuite	Run upstream testsuite	2026-06-16T17:38:41.0973348Z FAIL    exclude  (exit 1)
upstream testsuite	Run upstream testsuite	2026-06-16T17:38:41.1943899Z FAIL    files-from  (exit 1)
```

PR #5822, opened 2026-06-15
fix(transfer): emit --info=backup message in wire-transfer receiver path

oferchen:
## Summary

The upstream `backup.test` (`testsuite/backup.test` in rsync 3.4.4) asserts via grep that `--info=BACKUP` emits a `backed up <fname> to <buf>` notice during a wire transfer (`--no-whole-file --backup`). oc-rsync wired this emission only through the local-copy executor in `engine::local_copy::context_impl::state::backup_existing_entry`, so the line never fired on the wire-transfer paths the test exercises and the grep returned exit=1.

## Root cause

Upstream `backup.c:352` emits `INFO_GTE(BACKUP, 1)` `"backed up %s to %s"` from the `success:` label of `make_backup()`. That fires from every successful backup placement regardless of which transfer path placed the file. The local-copy executor in oc-rsync (`crates/engine/src/local_copy/context_impl/state.rs:633`) already paired `trace_make_backup_rename` with the matching `info_log!(Backup, 1, ...)` emission. The two wire-transfer backup sites - `transfer::disk_commit::process::make_backup` and `transfer::receiver::transfer::sync::ReceiverContext` (delta-apply branch) - only called the `--debug=BACKUP` producer; the `--info=BACKUP` producer was missing.

The upstream test forces the wire-transfer path with `--no-whole-file`, which is exactly the codepath that lacked the emission.

## Changes

- `crates/transfer/src/disk_commit/process.rs::make_backup`: emit `info_log!(Backup, 1, "backed up {} to {}", ...)` immediately after the existing `trace_make_backup_rename` call, citing `backup.c:352`.
- `crates/transfer/src/receiver/transfer/sync.rs`: same emission added at the receiver-sync backup site.
- Add unit-level regression test `make_backup_emits_info_backup_message` that drives `make_backup` end-to-end and asserts the byte-identical upstream line is observed via `logging::drain_events`.

## Test plan

- [x] `cargo fmt --all` clean
- [ ] CI: `fmt+clippy`, `nextest (stable)`, Windows, macOS, Linux musl
- [ ] CI: Interop validation (upstream rsync 3.0.9 / 3.1.3 / 3.4.1 / 3.4.2 / 3.4.4)
- [ ] CI: upstream-testsuite cell exercises `backup.test`

### DIFF
--- a/crates/transfer/src/disk_commit/process.rs
+++ b/crates/transfer/src/disk_commit/process.rs
@@ -14,6 +14,7 @@ use std::io;
 use std::path::{Path, PathBuf};
 
 use engine::{CleanupManager, compute_backup_path, trace_make_backup_rename};
+use logging::info_log;
 use protocol::acl::AclCache;
 
 use crate::delta_apply::{ChecksumVerifier, SparseWriteState};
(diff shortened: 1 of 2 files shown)


oferchen:
NACK — testsuite/backup.test still shows FAIL on this branch.

Root cause: `file_path.display()` emits an absolute path in the backup-target construction at `crates/transfer/src/receiver/transfer/sync.rs:346`, while upstream uses a path relative to the transfer root. The backup-dir layout therefore diverges from upstream.

Precise fix: replace `file_path.display()` with the relative path from `file_entry.path()` (or equivalent relative-path accessor on the FileEntry) so backup-target construction matches upstream's relative semantics.

oferchen:

This PR claims to fix upstream testsuite test(s): `backup`.

Latest upstream testsuite run on this PR's head: https://github.com/oferchen/rsync/actions/runs/27635543200/job/81721641938

```
upstream testsuite	Run upstream testsuite	2026-06-16T17:36:51.5579835Z FAIL    backup  (exit 1)
```

PR #5840, opened 2026-06-15
fix(flist): dedup consecutive identical entries after sort

oferchen:
## Summary

Mirror upstream rsync's flist dedup pass so repeated source operands
no longer transfer each file multiple times. The upstream testsuite
`duplicates.test` (`-avv` with the same source repeated 10 times)
asserts each file appears exactly once in the verbose output. Without
flist-stage dedup, oc-rsync's parallel collectors emitted ten copies
of every entry.

A new `sort::dedup_sorted_file_entries` helper walks the sorted
collector output and collapses adjacent entries with identical
`relative_path`. It runs after every existing `sort_file_entries`
call in `crates/flist/src/parallel.rs`, preserving the upstream
invariant that the dedup pass follows the sort pass.

## Upstream references

- `flist.c:3050-3081` - `flist_sort_and_clean()` compares consecutive
  sorted entries with `f_name_cmp()`; matches are flagged
  `FLAG_DUPLICATE`. Directory-vs-non-directory edge cases keep the
  more informative entry, which here collapses to keep-first because
  oc-rsync's `FileListEntry` carries no flag bit.
- `flist.c:2159-2172` - sender consults
  `flist->sorted[ndx]->flags & FLAG_DUPLICATE` and skips matching
  indices, which is what the upstream testsuite observes.
- `testsuite/duplicates.test` - in upstream 3.4.4
  `target/interop/upstream-src/rsync-3.4.4/testsuite/duplicates.test`,
  invokes `rsync -avv <src>/ <src>/ ... <dst>` and asserts the verbose
  output contains each file exactly once.

## Changes

- `crates/flist/src/sort.rs` - added
  `dedup_sorted_file_entries(&mut Vec<FileListEntry>)` plus unit
  tests covering collapse, distinct retention, empty input, and a
  walker-driven round-trip that mirrors `duplicates.test`'s argv
  shape.
- `crates/flist/src/parallel.rs` - call the new helper after each
  `sort_file_entries` call in `collect_paths_then_metadata_parallel`,
  `collect_with_batched_stats`, `collect_paths_chunked_parallel`, and
  `resolve_metadata_parallel`. Comments cite `flist.c:3050`.

The `FileListEntry` struct stays immutable (it carries `fs::Metadata`
which doesn't implement `Clone`), so the dedup pass uses
`Vec::dedup_by` to discard the second copy rather than flagging it.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo check -p flist --all-features`
- [x] `cargo check -p transfer --all-features`
- [ ] CI: nextest covers `sort.rs` unit tests + existing CLI
      `transfer_request_with_duplicates` end-to-end fixture.

### DIFF
--- a/crates/flist/src/parallel.rs
+++ b/crates/flist/src/parallel.rs
@@ -162,6 +162,9 @@ pub fn collect_paths_then_metadata_parallel(
 
     if errors.is_empty() {
         crate::sort::sort_file_entries(&mut entries);
+        // upstream: flist.c:3050 dedup_in_flist() - collapse identical
+        // sorted entries so repeated source operands transfer each file once.
+        crate::sort::dedup_sorted_file_entries(&mut entries);
         Ok(entries)
     } else {
         Err(errors)
@@ -277,6 +280,9 @@ pub fn collect_with_batched_stats(
 
     if errors.is_empty() {
         crate::sort::sort_file_entries(&mut entries);
+        // upstream: flist.c:3050 dedup_in_flist() - collapse identical
+        // sorted entries so repeated source operands transfer each file once.
+        crate::sort::dedup_sorted_file_entries(&mut entries);
         Ok(entries)
     } else {
         Err(errors)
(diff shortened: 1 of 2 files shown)


oferchen:
Closing as superseded. Merged #5820 (cli verbose-format fix) closed `testsuite/duplicates.test` on master — duplicates is no longer in the master upstream-testsuite FAIL list. This PR's flist-stage dedup approach is no longer needed, and its current implementation triggers `FAIL duplicates (exit 1)` in this PR's own CI run.

PR #5842, opened 2026-06-15
fix(filters): parse :C modifier on per-directory merge rules for CVS-ignore

oferchen:
## Summary

When a filter file read from disk (e.g. a `.rsync-filter` or per-directory
merge file like `mid/.filt`) contains a `:C` (or `.C`) directive with an
empty pattern, oc-rsync's `filters::parse_rules` parser rejected it as
`unrecognized filter rule` instead of defaulting the filename to
`.cvsignore`, and the `C` (CVS-ignore) modifier was never propagated from
`RuleModifiers` onto the resulting `FilterRule`. The wire-format path
(`transfer::generator::filters::wire_rule_to_dir_merge_config`) and the
CLI argv path (`crates/cli/.../parsing/merge.rs`) already handle `:C`
correctly; only the merge-file-on-disk path was broken. The failing
upstream `merge.test` and `exclude-lsh.test` scenarios depend on this
short form. Tracks task #4428 (UTS-V4.B).

## Changes

- `FilterRule` gains a `cvs_mode: bool` field with builder
  `with_cvs_mode` and accessor `is_cvs_mode`.
- `RuleModifiers::apply` propagates `cvs_mode` onto the rule.
- `try_parse_short_form` in `crates/filters/src/merge/parse.rs`
  special-cases empty pattern with `cvs_mode` set on `Merge`/`DirMerge`
  and defaults to `.cvsignore`.
- Destructure in `crates/filters/src/compiled/mod.rs` updated to accept
  the new field via `cvs_mode: _`.
- Three regression tests in `crates/filters/src/merge/tests.rs` cover
  `:C\n`, `:C my.ignore\n`, and `.C\n`.

## Upstream references

- `target/interop/upstream-src/rsync-3.4.4/exclude.c:1248-1255` -- `:C`
  sets `FILTRULE_NO_PREFIXES | FILTRULE_WORD_SPLIT | FILTRULE_NO_INHERIT | FILTRULE_CVS_IGNORE`.
- `target/interop/upstream-src/rsync-3.4.4/exclude.c:1324` -- empty
  pattern is only an error when `FILTRULE_CVS_IGNORE` is unset.
- `target/interop/upstream-src/rsync-3.4.4/exclude.c:1404-1408` --
  empty-pattern merge defaults to `.cvsignore`.
- `target/interop/upstream-src/rsync-3.4.4/testsuite/merge.test` and
  `testsuite/exclude-lsh.test:86-88` -- failing fixtures.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo check -p filters --all-features` (clean)
- [ ] CI full nextest matrix
- [ ] Upstream `merge.test` and `exclude-lsh.test` re-run on Linux host

## Notes

Task tracker #4428. Audit document referenced as the spec source.
The pre-existing PR #5817 is orthogonal: it changes the post-load
scope-evaluation path; this PR fixes the pre-load parse path. They
can land independently.

### DIFF
--- a/crates/filters/src/compiled/clear.rs
+++ b/crates/filters/src/compiled/clear.rs
@@ -30,6 +30,7 @@ mod tests {
             negate: false,
             exclude_only: false,
             no_inherit: false,
+            cvs_mode: false,
         };
         let mut compiled = CompiledRule::new(rule).unwrap();
         let still_active = compiled.clear_sides(true, false);
@@ -50,6 +51,7 @@ mod tests {
             negate: false,
             exclude_only: false,
             no_inherit: false,
+            cvs_mode: false,
         };
         let mut compiled = CompiledRule::new(rule).unwrap();
         let still_active = compiled.clear_sides(false, true);
@@ -70,6 +72,7 @@ mod tests {
             negate: false,
             exclude_only: false,
             no_inherit: false,
+            cvs_mode: false,
         };
         let mut compiled = CompiledRule::new(rule).unwrap();
         let still_active = compiled.clear_sides(true, true);
@@ -97,6 +100,7 @@ mod tests {
             negate: false,
             exclude_only: false,
             no_inherit: false,
+            cvs_mode: false,
         };
         let mut rules = vec![CompiledRule::new(rule).unwrap()];
         apply_clear_rule(&mut rules, false, false);
@@ -115,6 +119,7 @@ mod tests {
             negate: false,
             exclude_only: false,
             no_inherit: false,
+            cvs_mode: false,
         };
         let mut rules = vec![CompiledRule::new(rule).unwrap()];
         apply_clear_rule(&mut rules, true, false);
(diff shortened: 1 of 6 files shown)


oferchen:
NACK — `testsuite/exclude-lsh.test` still shows `FAIL exclude-lsh (exit 1)` on this branch.

Root cause: the parsing fix is incomplete. When `/mid/.filt` contains `:C`, the rule is parsed (with cvs_mode flag) but never triggers the runtime loading of `/mid/.cvsignore`. Two gaps:

1. **`crates/filters/src/merge/parse.rs:151-157`** — `RuleModifiers::apply()` never calls `.with_cvs_mode(self.cvs_mode)`, so the parsed flag doesn't propagate to the FilterRule.
2. **`crates/filters/src/chain/mod.rs:260-263`** — After a merge file is loaded, its rules are compiled into a FilterSet without scanning for nested `DirMerge` actions. The new `:C` rule never causes `.cvsignore` to be loaded at directory entry.

For a complete fix: when loading a merge file, walk its rules for `FilterAction::DirMerge`, extract pattern + cvs_mode, register a new `DirMergeConfig`, and trigger `enter_directory` semantics for the nested file. Verify locally with `bash testsuite/exclude-lsh.test` before re-pushing.

PR #5852, opened 2026-06-16
fix(engine): walk one level beneath trailing-slash --files-from sources

oferchen:
## Summary

Fixes the upstream-testsuite `files-from.test` local invocation, which had been failing because `--files-from` entries flagged as `DOTDIR_NAME` (`from/./`) and `SLASH_ENDING_NAME` (`subsubdir2/`) only created the destination directory and never walked its immediate children. The chk/to diff diverged on every flat child (`empty`, `emptydir`, `filelist`, `nolf`, `nolf-symlink`, `text`) and on `dir/subdir/subsubdir2/bin-lt-list`.

Upstream `flist.c:2442` honours `(xfer_dirs && name_type != NORMAL_NAME)` so `send_directory()` runs for trailing-slash entries even when global `recurse` is off (`options.c:2189` clears it whenever `--files-from` is active). Subdirectories encountered during that walk are not recursed into further, matching `recurse=0`.

- Introduce `copy_directory_walk_one_level` so `handle_directory_contents_copy` can request the one-level walk on trailing-slash sources.
- Share the inner routine with `copy_directory_recursive` via a `force_walk_one_level` flag threaded through the bail-out.
- Child directories reached via `process_planned_entry::CopyDirectory` still go through the normal entry and bail out, preserving `recurse=0` semantics inside the walk.
- Add two CLI-level regression tests covering the DOTDIR (`from/./`) and SLASH_ENDING (`leaf/`) flows on the local code path.

## Scope

This PR addresses the **local** (first) invocation of `testsuite/files-from.test`. The 4 SSH-mode invocations (deep-dive sub-tasks #4507-#4511) are deliberately out of scope for this PR; they will surface as separate failures once the local case flips green and require generator-side / push-forwarding work. Splitting them avoids blending unrelated fixes.

## Test plan

- [x] `cargo fmt --all` clean
- [ ] CI: `Upstream Testsuite` workflow flips `files-from` from FAIL to PASS on its local invocation
- [ ] CI: existing `cargo nextest` workspace remains green (`execute_with_trailing_separator_copies_contents`, `execute_skips_directories_when_recursion_disabled_without_dirs`, `execute_creates_directories_when_dirs_enabled_without_recursion`, `files_from_*` CLI tests, etc.)
- [ ] CI: `fmt+clippy`, Windows / macOS / Linux musl stable matrices

### DIFF
--- a/crates/cli/src/frontend/tests/transfer_request_with_files_from.rs
+++ b/crates/cli/src/frontend/tests/transfer_request_with_files_from.rs
@@ -155,3 +155,117 @@ fn files_from_embedded_dot_marker_determines_destination_structure() {
         b"hello"
     );
 }
+
+/// Verifies that `--files-from` entries flagged as DOTDIR_NAME (`from/./`) and
+/// SLASH_ENDING_NAME (`dir/`) walk their immediate children even when global
+/// recursion is off, matching upstream `flist.c:2442`.
+///
+/// Upstream's `(xfer_dirs && name_type != NORMAL_NAME)` predicate forces
+/// `send_directory()` to emit one level of contents for the listed directory.
+/// Subdirectories encountered during that walk are NOT recursed into further
+/// (matching `recurse=0` semantics), but flat files and immediate subdirs are
+/// transferred.
+///
+/// # Upstream Reference
+///
+/// - `testsuite/files-from.test` - the local invocation exercises this path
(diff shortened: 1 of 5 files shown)


viktoriavikulya (commented):
The happy path is clean. What happens if the third-party API returns 429 here? Might be worth handling that case explicitly.

PR #5864, opened 2026-06-17
fix(transfer): open basis with O_NOFOLLOW on basename

oferchen:
## Summary

- Receiver-side basis open went through `fs::File::open(path)`, which follows every symlink including a pre-planted symlinked basename. That breaks the upstream `symlink-dirlink-basis.test` security contract and diverges from upstream `syscall.c:705 do_open_at()`, which splits dirname / basename: the dirname is resolved with normal symlink semantics (so `--copy-dirlinks` directory symlinks keep working), while the basename is opened with `openat(dirfd, basename, O_NOFOLLOW)` so a symlinked leaf is rejected with `ELOOP`.
- Adds `fast_io::open_basis_nofollow()` exposing that split (Unix: `openat(O_NOFOLLOW)`; non-Unix: standard open) and routes both `try_open_file()` and `try_reference_directories()` in `receiver/basis.rs` through it.
- Closes the upstream `symlink-dirlink-basis.test` failure (issue #715 regression + 7 edge-case sub-tests covering `-K`, `--backup`, `--inplace`, `--partial-dir`, `--protocol=28`, top-level files, and nested directory symlinks).

Upstream reference (`target/interop/upstream-src/rsync-3.4.3/syscall.c:705-756`):

```c
int do_open_at(const char *pathname, int flags, mode_t mode)
{
        ...
        slash = strrchr(pathname, '/');
        if (!slash)
                return do_open(pathname, flags, mode);
        ...
        dfd = secure_relative_open(NULL, dirpath, O_RDONLY | O_DIRECTORY, 0);
        if (dfd < 0)
                return -1;
        ...
        ret = openat(dfd, bname, flags | O_NOFOLLOW | O_BINARY, mode);
        ...
}
```

## Test plan

- [ ] CI fmt + clippy
- [ ] CI nextest (stable) - new tests under `crates/transfer/src/receiver/basis.rs` and `crates/fast_io/src/nofollow_open.rs`:
  - `try_open_file_follows_directory_symlink` - positive case (issue #715 main regression)
  - `try_open_file_rejects_symlinked_basename` - security regression
  - `try_open_file_handles_top_level_path` - upstream `if (!slash)` short-circuit
  - `try_reference_directories_rejects_symlinked_basename` - `--copy-dest` / `--link-dest` path
  - `opens_basis_through_directory_symlink` / `opens_basis_through_nested_directory_symlink` / `rejects_symlinked_basename` / `opens_top_level_basis_without_split` / `missing_path_returns_not_found`
- [ ] CI Linux musl / macOS / Windows matrix (non-Unix fallback is a plain `File::open`)
- [ ] CI interop with upstream 3.4.3 - `symlink-dirlink-basis.test` (8 sub-tests covering `-K`, `--backup`, `--inplace`, `--partial-dir`, `--protocol=28`, top-level files, nested dir symlinks)

### DIFF
--- a/crates/fast_io/src/lib.rs
+++ b/crates/fast_io/src/lib.rs
@@ -104,6 +104,9 @@ pub mod kernel_version;
 /// the SEC-1.l audit) and do not depend on this module.
 #[cfg(unix)]
 pub mod linux_capabilities;
+/// Receiver-side basis open with `O_NOFOLLOW` on the basename, matching
+/// upstream rsync's `do_open_at()` dirname/basename split.
+pub mod nofollow_open;
 /// Page-aligned buffer pool for IOCP no-buffering mode.
 pub mod page_aligned;
 /// Parallel file I/O operations using rayon.
@@ -322,6 +325,7 @@ pub use kernel_version::{
 };
 #[cfg(unix)]
 pub use linux_capabilities::openat2_supported;
+pub use nofollow_open::open_basis_nofollow;
 pub use refs_detect::{clear_refs_cache, is_refs_filesystem};
 #[cfg(unix)]
 pub use secure_dir::secure_open_dir;
--- a/crates/fast_io/src/nofollow_open.rs
+++ b/crates/fast_io/src/nofollow_open.rs
@@ -1,0 +1,206 @@
+//! Receiver-side basis file open with `O_NOFOLLOW` on the basename.
+//!
+//! Mirrors upstream rsync's `do_open_at()` / `secure_relative_open()`
+//! dirname/basename split in `syscall.c:705` and `syscall.c:1769`: the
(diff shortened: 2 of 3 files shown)


oferchen:
Tests added here are unit-level on the internal primitive (good for the security boundary). They do not yet mirror the upstream `testsuite/symlink-dirlink-basis.test` shape — spawning the oc-rsync binary with `-K -a src/ dest/` through a directory-symlink basis under delta-transfer (>32KB file with small modification) and asserting `diff -r` parity, the way upstream's test does.

Follow-up task filed to add the binary-invocation mirror, modelled after the upstream .test script's 8 sub-scenarios (basic dir symlink, -z compression, nested symlinks, --backup, --inplace, top-level file, --partial-dir proto28, basic proto28).

PR #5865, opened 2026-06-17
fix(filters): parse :C cvs-modifier to merge .cvsignore per-dir

oferchen:
## Summary

- Recognise the `:C` cvs-modifier when it appears inside another per-directory merge file: the chain now loads the referenced `.cvsignore` from the current directory as a CVS-style ignore list, scoped (no-inherit) to that directory only.
- Previously the parser produced the correct `FilterRule::dir_merge(".cvsignore")` with `cvs_mode=true`, but `FilterSet::from_rules` silently discarded every `DirMerge` rule, so the named file was never read.

Upstream reference: `target/interop/upstream-src/rsync-3.4.3/exclude.c` `parse_rule_tok()` handles `'C'` at lines 1248-1254, and `exclude.c:1419-1428` registers a perdir merge when a parsed rule carries `FILTRULE_PERDIR_MERGE`.

Closes the upstream testsuite `exclude` and `exclude-lsh` failures, which both drive `:C` via `mid/.filt` (loaded through `dir-merge_.filt`).

## Test plan

- [ ] `cargo nextest run -p filters -E 'test(dir_merge_inline_colon_c)'` covers the new regression tests:
  - `dir_merge_inline_colon_c_loads_cvsignore_no_inherit`
  - `dir_merge_inline_colon_c_missing_cvsignore_is_noop`
- [ ] Existing chain tests (`filter_chain_enter_directory_*`) keep passing - the new code path is gated on `DirMerge` rules being present in parsed merge content, so other configurations are unaffected.
- [ ] Upstream `exclude` and `exclude-lsh` testsuite checks pass in the interop workflow.

### DIFF
--- a/crates/filters/src/chain/mod.rs
+++ b/crates/filters/src/chain/mod.rs
@@ -40,7 +40,7 @@ use std::io;
 use std::path::Path;
 
 use crate::merge::parse::parse_rules;
-use crate::{FilterRule, FilterSet};
+use crate::{FilterAction, FilterRule, FilterSet};
 
 pub use config::DirMergeConfig;
 pub use error::FilterChainError;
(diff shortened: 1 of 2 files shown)


oferchen:
Unit tests cover the filter-chain primitive (FilterSet::from_rules drop-DirMerge gap closure) — good for the load semantics. Follow-up task #4531 + #4532 will add the binary-invocation mirror of upstream `testsuite/exclude.test` and `testsuite/exclude-lsh.test` (spawning oc-rsync with `--cvs-exclude` / `-C` / `-f:C` flags through fixture trees containing `.cvsignore` files, asserting `diff -r` parity with upstream's expected layouts).

PR #5911, opened 2026-06-17
fix(cli): unblock master CI - classify -f as value-bearing by action

oferchen:
## Summary

- `classify_short_options` only flagged a short option as value-bearing when `Arg::get_num_args()` returned a positive range. Clap only infers `num_args = 1` for `ArgAction::Set`/`Append` during `Command::build()`, which the CLI front-end skips on the way to `expand_short_options`.
- The production filter Arg (`crates/cli/src/frontend/command_builder/sections/transfer_behavior_options.rs:610`) omits an explicit `.num_args(1)` because clap infers it from `ArgAction::Append`. Pre-build, `get_num_args()` was `None`, the filter was bucketed into `flag_options`, and `expand_cluster` returned `None` for `-f:C`. Clap then routed the unsplit `-f:C` token to the trailing positional `args`, and `parsed.filters` came back empty.
- Trust the action directly: `Set`/`Append` always consume one value. Adds a regression test that builds an Arg matching the production shape (no explicit `.num_args(1)`).

Closes the two master nextest failures introduced by #5897:
- `frontend::arguments::parser::tests::filter_short_arg_packed_colon_rule`
- `frontend::arguments::parser::tests::filter_short_arg_packed_dash_rule`

## Test plan

- [ ] CI nextest (stable) green on Linux / macOS / Windows / musl
- [ ] `cargo nextest run -p cli --all-features -E 'test(filter_short_arg)'`

### DIFF
--- a/crates/cli/src/frontend/arguments/short_options.rs
+++ b/crates/cli/src/frontend/arguments/short_options.rs
@@ -70,7 +70,6 @@ fn classify_short_options(command: &ClapCommand) -> (HashSet<char>, HashSet<char
         };
 
         let num_args = argument.get_num_args().unwrap_or(ValueRange::EMPTY);
-        let takes_values = num_args.takes_values();
         let min_values = num_args.min_values();
         let action = argument.get_action();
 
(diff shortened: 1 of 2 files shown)


oferchen:
Closing this attempt - the surgical .num_args(1) fix on the filter Arg has an unexplained side effect on -hh (human_readable_double_short_hh_is_combined) test parsing. The underlying issue (master CI filter_short_arg_packed_* failures) needs a different approach. Will investigate fresh.

PR #5926, opened 2026-06-18
ci: add per-test timeout wrapper around upstream runtests.py

oferchen:
## Summary
- Adds `tools/ci/run_upstream_testsuite_with_timeout.sh`, a wrapper that invokes upstream's `runtests.py --name <test>` once per test under coreutils `timeout(1)`, classifies each outcome as PASS / FAIL / SKIP / XFAIL / TIMEOUT, and exits non-zero if any test fails or times out.
- Adds `tools/ci/upstream_testsuite_timeouts.conf` with a 120s default plus per-test overrides for known-slow tests (`daemon-config`, `daemon`, `hardlinks`, `files-from`, `acls`, `xattrs`).
- Defensive guards: missing `timeout(1)`, missing upstream source tree, missing `runtests.py`, missing `oc-rsync` or upstream `rsync` binary all fail fast with exit 2.

Per-test timeout resolution (highest precedence first):
1. `PER_TEST_TIMEOUTS` entry in the config
2. `OC_RSYNC_PER_TEST_TIMEOUT` environment variable
3. `DEFAULT_TIMEOUT` from the config (120s)

The wrapper surfaces both its own per-test harness log and the upstream per-test log from `$RSYNC_TESTTMP/<name>/test.log` (default `~/rsync/testtmp/<name>/test.log`).

This PR ships the wrapper only. Wiring it into the upstream-testsuite GitHub Actions workflow will land separately.

## Test plan
- [x] `bash -n` and `shellcheck -x` clean on the wrapper script.
- [x] Smoke-tested with a fake upstream tree (dummy `runtests.py`): PASS / FAIL (exit 7) / TIMEOUT (sleep > budget) all classify correctly; overall exit 1 when any non-PASS, 0 otherwise.
- [x] Per-test override beats `OC_RSYNC_PER_TEST_TIMEOUT`; env override beats config default.
- [x] Invalid (non-numeric) timeout values rejected with exit 2.
- [x] Missing upstream tree / missing binaries surface as exit 2 errors.
- [ ] CI integration validation lands with the workflow wire-up PR.

### DIFF
--- a/tools/ci/run_upstream_testsuite_with_timeout.sh
+++ b/tools/ci/run_upstream_testsuite_with_timeout.sh
@@ -1,0 +1,300 @@
+#!/usr/bin/env bash
+# run_upstream_testsuite_with_timeout.sh - run upstream's runtests.py with a
+# per-test timeout wrapper.
+#
+# Upstream's runtests.py does not enforce a per-test timeout, so a single
+# hanging test (we have seen `daemon`, `files-from`, `hardlinks` hang for
+# 300s+) blocks the entire suite. This wrapper invokes runtests.py once
+# per test name via `runtests.py --name <test>` under coreutils
+# `timeout(1)`, classifies each outcome (PASS / FAIL / SKIP / TIMEOUT),
+# and exits non-zero if any test FAILS or TIMES OUT.
+#
+# Per-test timeout resolution (highest precedence first):
+#   1. PER_TEST_TIMEOUTS entry in upstream_testsuite_timeouts.conf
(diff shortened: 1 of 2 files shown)


oferchen:
Closing — runtests.py already supports `--timeout SECS` natively (default 300s, with hardlinks bumped to 600s). The wrapper in this PR is unnecessary; the upstream-testsuite CI workflow can pass `--timeout` directly to runtests.py without an additional flow. Will instead update the existing IFX-15.a workflow to use `--timeout 120` (or per-test override via env).

PR #5948, opened 2026-06-18
fix(itemize): emit unchanged dir rows and honour -vv suppression gate

oferchen:
## Summary

Two-part fix for the upstream `testsuite/itemize.test` `-ivvplrtH` golden, where
oc-rsync's local-copy executor was missing six unchanged rows the upstream
generator gate keeps:

```
.d ./
.d bar/
.d bar/baz/
.d foo/
.f foo/config1
.L foo/sym -> ../bar/baz/rsync
```

Upstream `generator.c:574-586` keeps an entry's itemize row when
`(iflags & SIGNIFICANT_ITEM_FLAGS) || INFO_GTE(NAME, 2) || stdout_format_has_i > 1 || (xname && *xname)`.
The local-copy path was diverging in two places that jointly hid these rows:

1. **Engine-level dir gate**: `directory/recursive/mod.rs` emitted a
   `MetadataReused` record only when `change_set.has_any_change()` was true.
   Files (`copy/transfer/execute/skip.rs`) and symlinks (`special/symlink.rs`)
   already emitted unconditionally and relied on the render layer to suppress
   uninteresting rows. The gate is now removed so all three entry kinds emit
   the same way.

2. **Render-level suppression gate**: `out_format/render/mod.rs::should_suppress_event`
   collapsed every `MetadataReused` event with no changes, regardless of `-v`
   level. The gate now also honours upstream's `INFO_GTE(NAME, 2)` OR-term:
   `OutFormatContext` carries the CLI verbose counter through
   `with_verbose_level()`, and `should_suppress_event` lets unchanged rows
   through whenever `context.verbose_level() >= 2`.

The verbose counter is threaded from the CLI's `verbosity: u8` through
`drive/summary.rs` (both the stdout path and the log-file path) into
`OutFormatContext::with_is_sender(is_sender).with_verbose_level(verbosity)`.

## Test plan

- [ ] CI: fmt + clippy
- [ ] CI: nextest (stable, Linux/macOS/Windows/musl)
- [ ] Upstream interop: `testsuite/itemize.test` `-ivvplrtH` invocation 5 emits
      the six missing rows
- [ ] Regression: a `-v` (single-verbose) `MetadataReused` event with no
      changes is still suppressed (only `-vv+` opens the gate)
- [ ] Regression: `MetadataReused` events with `was_created()` or any
      `ChangeSet` flag still render at any `-v` level

### DIFF
--- a/crates/cli/src/frontend/execution/drive/summary.rs
+++ b/crates/cli/src/frontend/execution/drive/summary.rs
@@ -122,7 +122,8 @@ where
                 });
             }
 
-            let out_format_context = OutFormatContext::with_is_sender(is_sender);
+            let out_format_context =
+                OutFormatContext::with_is_sender(is_sender).with_verbose_level(verbosity);
             if let Err(error) = with_output_writer(stdout, stderr, msgs_to_stderr, |writer| {
                 emit_transfer_summary(
                     &summary,
@@ -217,7 +218,7 @@ fn emit_log_output(params: EmitLogOutputParams<'_>) -> io::Result<()> {
         human_readable_mode,
         is_sender,
     } = params;
-    let context = OutFormatContext::with_is_sender(is_sender);
+    let context = OutFormatContext::with_is_sender(is_sender).with_verbose_level(verbosity);
     // The log file already carries the parallel "building file list" line via
     // logging::info_log!(Flist, 1, ...); the FCLIENT "sending incremental file
     // list" banner is for stdout only (upstream: flist.c:2248 vs 2252).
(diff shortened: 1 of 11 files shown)


oferchen:
Superseded by #5942 (commit f306ca55f), which implements the same upstream parity fix using an `emit_unchanged` flag derived from `name_level == UpdatedAndUnchanged`. That mechanism covers the same upstream `generator.c:582-583` INFO_GTE(NAME,2) OR-term as the `verbose_level >= 2` check in this branch and is already on master.

PR #5978, opened 2026-06-19
fix(protocol): use varlong-3 for signature block_count / block_length

oferchen:
## Summary

- Swap `read_varint`/`write_varint` for `read_varlong(reader, 3)` / `write_varlong(writer, n, 3)` on `block_count` and `block_length` in `crates/protocol/src/wire/signature.rs`. Upstream rsync 3.4.x emits `sum_head->count` and `sum_head->blength` via `write_varlong()` once they cross the 30-bit varint range; decoding them as plain varint produces `InvalidData: overflow in read_varint` on large basis files.
- Keep `strong_sum_length` on plain varint with an `<= MAX_DIGEST_LEN` (32) bound matching upstream `checksum.c`. Read raw as `i64`/`i32`, narrow with `u32::try_from`, surface `InvalidData` on overflow.
- Add `signature_roundtrip_large_block_count` unit test exercising `block_count = 5_000_000` to lock in the varlong path.

Upstream reference: `match.c::send_sums()` + `io.c::read_varlong()`.

Closes UTS-V3.C and the symmetric pull leg tracked as PULL-VARINT (#4345).

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo check --workspace --all-features` clean
- [ ] CI: nextest stable (Linux/macOS/Windows/musl)
- [ ] CI: clippy + fmt
- [ ] Upstream testsuite: `reverse-daemon-delta` now decodes without `overflow in read_varint`

### DIFF
--- a/crates/protocol/src/wire/signature.rs
+++ b/crates/protocol/src/wire/signature.rs
@@ -7,7 +7,10 @@
 
 use std::io::{self, Read, Write};
 
-use crate::varint::{read_varint, write_varint};
+use crate::varint::{read_varint, read_varlong, write_varint, write_varlong};
+
+/// Upstream caps a strong checksum at 32 bytes (`MAX_DIGEST_LEN` in checksum.c).
+const MAX_DIGEST_LEN: u8 = 32;
 
 /// Single signature block containing rolling and strong checksums.
 ///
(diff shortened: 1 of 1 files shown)


oferchen:
Closing: investigation revealed the PR's premise is incorrect. Upstream rsync's `write_sum_head` uses fixed-4-byte `write_int` (see io.c:2072-2084 in 3.4.4), NOT varlong with min_bytes=3.

Switching to `write_varlong(_, _, 3)` forces a 3-byte minimum encoding, breaking the wire format for small `block_count`/`block_length` values that the existing `write_varint` path emits in 1 byte. This regresses `list_only_lists_entries_without_copying` and `list_only_large_file_size_has_thousands_separators` on macOS stable across multiple feature combinations.

The 'overflow in read_varint' symptom this PR aimed to fix is a separate problem - master uses varint while upstream uses fixed 4-byte int. Reconciling that needs a different approach (e.g. `read_int` for upstream-compat) rather than varlong.

Closing this branch. Will track follow-up under #4345 PULL-VARINT.

PR #6032, opened 2026-06-20
fix(itemize): propagate verbose level to server generator thread

oferchen:
## Problem

Under `-ivv` (info=name level 2), upstream rsync emits an itemize line for
every entry, including unchanged "uptodate" dir/file/symlink rows. In
server-driven mode (the `lsh.sh` CI path), oc-rsync dropped those unchanged
rows, diverging from upstream's output.

## Root cause

Verbosity is stored as a per-thread thread-local (`logging::thread_local`),
mirroring upstream's file-scope `info_levels[]` but scoped per thread because
each thread may drive a separate transfer pipeline. The client-driving
workflow seeds only the client thread. The `--server` entry
(`crates/cli/src/frontend/server/run.rs`) runs its generator on its own
thread but never installed the negotiated verbosity, so the thread-local
stayed at the default (`NAME=0`). The generator's emit-every-entry gate
`info_gte(InfoFlag::Name, 2)` (upstream: `generator.c:582-583
INFO_GTE(NAME, 2)`) was therefore always false, suppressing the unchanged
rows.

Compounding this, the compact server flag string forwards `-vv` as repeated
`v` characters (upstream: `options.c:2625-2626`), but the parser collapsed
them into a single `bool`, discarding the count needed to reach level 2.

## Fix

- Preserve the verbose count: `ParsedServerFlags` gains a `verbose_level: u8`
  that counts each `v` in the flag string, following the existing
  `fuzzy_level` / `one_file_system` counting pattern. The `verbose` bool is
  retained for existing boolean predicates.
- Install the negotiated verbosity on the server/generator thread:
  `run_server_mode` now calls
  `logging::init(VerbosityConfig::from_verbose_level(verbose_level))` before
  running the generator, and before the existing `--info` override so the
  higher-priority `--info` settings still win (mirroring upstream
  `set_output_verbosity(..., DEFAULT_PRIORITY)` followed by
  `parse_output_words`).

The generator runs synchronously on this thread (no rayon worker emits
itemize lines), so a single `init` on the server thread is sufficient. The
itemize gate itself is unchanged.

## Tests

- `single_verbose_flag_sets_level_one`, `double_verbose_flag_preserves_count`,
  `verbose_level_defaults_to_zero` cover the count parsing.
- `double_verbose_flag_string_enables_name_level_two` exercises the
  end-to-end mapping: a `-vv` flag string installed via `from_verbose_level`
  flips `info_gte(Name, 2)` true, encoding why the count must survive.

### DIFF
--- a/crates/cli/src/frontend/server/run.rs
+++ b/crates/cli/src/frontend/server/run.rs
@@ -25,7 +25,7 @@ where
     Out: Write,
     Err: Write,
 {
-    use core::server::{ServerConfig, ServerRole, run_server_stdio};
+    use core::server::{run_server_stdio, ServerConfig, ServerRole};
 
     let program_brand =
         super::super::detect_program_name(args.first().map(OsString::as_os_str)).brand();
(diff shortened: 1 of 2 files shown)


oferchen:
Converting to draft. Root-cause investigation (local `rsync -ivvplrtH` repro on host, confirmed against upstream 3.4.4 as source of truth) shows this verbose-level-to-server-thread approach does not close the itemize test: the failing leg is a **local-copy** transfer whose rayon workers read a per-thread `VERBOSITY` that nobody installed, so `info_gte(NAME,2)` is false and unchanged `.d/.f/.L` rows are dropped. The faithful fix mirrors upstream's per-process `info_levels[]` global as a **session-scoped** verbosity shared across all threads of a transfer (kept session-scoped, not process-global, to preserve daemon multi-connection isolation). That supersedes this PR. See logging/src/thread_local.rs:3-6 for the divergence comment.

oferchen:
Superseded by #6036. The itemize divergence is not a thread-local verbosity propagation issue — it's the CLI `name_level` resolution, which derived solely from `-i` and never reached `UpdatedAndUnchanged`, so `-vv`/`--info=name2` could not enable unchanged itemize rows. #6036 fixes that at the parser, on the main thread, with no logging-crate change.

Issue #5421, opened 2026-06-03
fix: upstream testsuite relative fails with exit 1


oferchen:
## Context

The upstream rsync testsuite `relative` test fails with exit code 1 when run against oc-rsync. This test validates the `--relative` (`-R`) option which preserves full path components from the source in the destination.

## Root Cause

Likely a divergence in how oc-rsync handles `--relative` path manipulation. Upstream uses `.` insertion (implied dirs), handles leading slashes, and interacts with `--no-implied-dirs`. The path construction logic in `flist.c` may not be fully replicated.

## Steps to Reproduce

```bash
bash tools/ci/run_upstream_testsuite.sh relative
```

## Acceptance Criteria

- [ ] `relative` test passes (exit 0) against oc-rsync
- [ ] Read upstream `testsuite/relative.test` to understand exact assertions
- [ ] Read upstream C source for `--relative` in `flist.c` and `main.c`
- [ ] Fix divergence in oc-rsync relative path handling
- [ ] No regressions in other testsuite tests

## References

- Upstream test: `testsuite/relative.test` in rsync 3.4.1 source
- Harness: `tools/ci/run_upstream_testsuite.sh`
- Workflow: `.github/workflows/upstream-testsuite.yml`

oferchen:
## Root Cause (from CI log analysis)

The `--relative` (`-R`) test fails because intermediate directories from the relative path are being created in the destination. The diff shows extra content (`./down/3/deep/...`) appeared when only `./down/` should be in the initial listing:
```
--- expected: ./down (2 entries)
+++ actual:   ./down/3/deep/... (18 entries)
```

The issue is that `-R` is creating the full path structure and transferring all files, when the initial `checkit` expects only the implied directories up to `./down/` to appear.

oferchen:
Status update: This relative upstream testsuite failure (exit 1) remains unresolved. The test exercises --relative / -R path handling including implied directories and dot-dir anchoring. Needs dedicated investigation into relative path translation behavior.

oferchen:
## Resolution

**Fixed by PR #5441** (`fix/anchored-pattern-descendant-matching`)

**Root cause:** Anchored exclude patterns (e.g., `/*`, `/build`) incorrectly generated descendant matchers like `*/**` that matched nested paths. This caused `--include=/down/ --exclude='/*'` to exclude files inside the included `down/` directory - a scenario exercised by the upstream `relative` testsuite test.

[...]

PR #6034, opened 2026-06-20
fix(filters): protect excluded-dir children on the delete pass

oferchen:
## Problem

A directory-only unanchored wildcard exclude such as `- foo/*/` combined with `--delete-during` (or `--delete`) over-deletes the children of the matched directory. For example, `foo/sub/file1` and nested entries are deleted on the receiver even though the directory `foo/sub` is excluded.

Upstream rsync protects those children by never descending into an excluded directory during the generator walk (subtree pruning in `exclude.c`). oc-rsync instead evaluates each deletion candidate against the compiled rules per-path, so it needs the synthesised `{core}/**` descendant matcher to be live on the deletion path.

## Root cause

PR #6015 ("apply directory-only descendants gate universally") folded directory-only unanchored wildcards into the compile-time `suppress_descendants` gate. That was the correct fix for a transfer-side over-exclusion: `foo/*/` must not pre-bake a blanket `foo/*/**` rule that would steal precedence from a later `+ foo/s?b/` on the sender path.

But the suppression was unconditional, so the descendant was also stripped from the deletion path - the one place where oc-rsync's per-candidate matcher actually needs it. With the descendant gone, a candidate like `foo/sub/file1` matched no rule, fell through to the default `transfer_allowed = true`, and was deleted.

## Fix

Split the synthesised `{core}/**` descendant for directory-only unanchored wildcard excludes into a separate, deletion-only matcher set:

- **Transfer path** keeps the #6015 behaviour: no `foo/*/**` is ever synthesised, so the per-directory wildcard semantics (and the `+ foo/s?b/` precedence) are preserved.
- **Deletion path** consults the deletion-only descendants, so an excluded directory protects its children from the delete pass, matching upstream's subtree-pruning result.

The change is mirrored in both compiled-rule implementations:

- `crates/filters/src/compiled` (`CompiledRule`) - the receiver / daemon deletion path through `FilterChain`/`FilterSet`.
- `crates/engine/src/local_copy/filter_program/segments.rs` (engine-local `CompiledRule`) - the local-copy delete pass.

Deletion call sites route through a new `matches_for_deletion()` that adds the deletion-only descendants on top of the regular matchers, independent of the `check_descendants`/traversal gate (the local-copy delete pass calls with `check_descendants = false`).

Anchored wildcards (`/*`, `/cache_?/`) remain fully suppressed in both contexts so `*/**` cannot over-match nested paths (regression #5421 stays fixed).

## Tests

- `filters`: deletion protection for `foo/*/` children on both the single-path and traversal entry points, plus an explicit guard that the transfer path stays unaffected (#6015 guard), plus exact deletion-descendant matcher-set assertions.
- `engine`: `FilterSegment::apply` protects `foo/sub/file1` from deletion while leaving the transfer decision allowed.

No wire protocol changes. No public API changes.

### DIFF
--- a/crates/engine/src/local_copy/filter_program/segments.rs
+++ b/crates/engine/src/local_copy/filter_program/segments.rs
@@ -74,11 +74,25 @@ impl FilterSegment {
         // literal name-match semantics and prevents anchored patterns like
         // `- /bar` from over-matching deep paths (UTS-V3-B exclude-lsh).
         let check_descendants = false;
+        // upstream: an excluded directory protects its descendants from
+        // deletion because the generator never descends into it. On the
+        // deletion scan a directory-only unanchored wildcard exclude (`foo/*/`)
+        // therefore keeps `foo/sub/file1` off the delete pass via its
+        // deletion-only descendant matchers; the Transfer path must NOT see
+        // those (it would re-trip the #6015 `foo/*/` over-exclusion).
+        let for_deletion = matches!(context, FilterContext::Deletion);
+        let rule_matches = |rule: &CompiledRule, path: &Path, is_dir: bool| {
+            if for_deletion {
+                rule.matches_for_deletion(path, is_dir, check_descendants)
+            } else {
+                rule.matches(path, is_dir, check_descendants)
+            }
+        };
         for rule in &self.include_exclude {
             if outcome.transfer_decided() {
                 break;
             }
-            if rule.matches(path, is_dir, check_descendants) {
+            if rule_matches(rule, path, is_dir) {
                 if matches!(context, FilterContext::Deletion) && rule.applies_to_receiver {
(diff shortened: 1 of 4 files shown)


oferchen:
Host CI-mode verification (exclude.test, 2026-06-21): this change correctly closes the `foo/sub/file1` over-deletion (now present in both expected and actual). However the `exclude` upstream test still fails on a *separate, pre-existing* leg — the `--cvs-exclude --delete-excluded` (`:C`) leg under-deletes four cvs-excluded files (`file1.bak`, `file4.junk`, `home-cvs-exclude`, `file2.old`). None of those are directory-only unanchored wildcards, so this PR's deletion-descendant matcher does not touch them; the cvs-exclude deletion-scan gap is tracked as a follow-up. This PR is a correct, regression-tested partial fix with no required-check or feature regression.

PR #6038, opened 2026-06-21
fix(server): seed verbosity from packed -v count for itemize parity

oferchen:
## Problem

The upstream `itemize` testsuite test FAILs against oc-rsync. The test drives transfers over a remote shell (lsh.sh), where oc-rsync runs in `--server` mode and the client's `-vv` arrives as packed `v` letters in the server flag string (not as `--info=name2`).

Upstream gates the itemize line for an *unchanged* entry on `INFO_GTE(NAME, 2)` (generator.c:582-583), which `-vv` sets via `info_verbosity[]` (options.c:239-243). In oc-rsync the `--server` entry (`crates/cli/src/frontend/server/run.rs`) seeded thread-local verbosity **only** from an explicit `--info` long flag, so a packed `-vv` left the server thread at the zero-default and `info.name` never reached 2 — the unchanged itemize rows were suppressed.

Two compounding gaps:
1. `ParsedServerFlags` collapsed every `v` to a single `bool`, discarding the count needed to distinguish `-v` from `-vv`.
2. `run.rs` never called `logging::init(from_verbose_level(n))` for the packed-letter path (the client path at `drive/workflow/run.rs:272` does).

PR #6036 fixed the equivalent on the **client** arg parser; it did not touch the server path the SSH itemize test exercises.

## Fix

- Track the `v` count in `ParsedServerFlags::verbose_level` (incremented in `parse_transfer_flag`; the `v` inside the trailing `-e.` capability section is routed to `parse_info_flag` and correctly ignored).
- In the `--server` entry, seed thread-local verbosity from `from_verbose_level(verbose_level)` before layering any explicit `--info`, mirroring upstream `set_output_verbosity` (verbose-then-info ordering).

`verbose: bool` is retained for the ~15 existing `flags.verbose && client_mode` consumers; `verbose_level` is additive.

## Tests

- `counts_verbose_level_from_packed_flags`: `-v`/`-vv`/`-vvv` count 1/2/3, and the capability-suffix `v` does not inflate the count.

End-to-end coverage is the upstream `itemize` cell in the upstream-testsuite workflow.

upstream: options.c set_output_verbosity, options.c:239-243 info_verbosity[], generator.c:582-583.

### DIFF
--- a/crates/cli/src/frontend/server/run.rs
+++ b/crates/cli/src/frontend/server/run.rs
@@ -123,10 +123,26 @@ where
         return code;
     }
 
+    // upstream: options.c set_output_verbosity - the packed `-v` count maps to
+    // info/debug levels via info_verbosity[] (options.c:239-243) before any
+    // explicit --info override is layered on. Without this, the server thread's
+    // thread-local verbosity stays at the zero default, so `-vv` never raises
+    // info.name to 2 and the itemize line for an unchanged entry
+    // (INFO_GTE(NAME, 2), generator.c:582-583) is suppressed - exactly the
+    // gap that fails the upstream `itemize` test under SSH `--server` mode,
+    // where `-vv` arrives as packed `v` letters rather than `--info=name2`.
+    if config.flags.verbose_level > 0 {
+        logging::init(logging::VerbosityConfig::from_verbose_level(
+            config.flags.verbose_level,
+        ));
+    }
+
     // upstream: options.c parse_output_words - server-side info parsing
     // silently ignores unknown tokens so a newer client can forward names
     // this build has not learned yet. The well-formed empty/level errors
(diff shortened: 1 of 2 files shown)


oferchen:
Closing: the prior diagnosis was wrong. The upstream `itemize` test runs in LOCAL mode (`oc-rsync -ivvplrtH from/ to/`), not over SSH --server, so the packed -v server-flag count this PR touched is never exercised by the failing assertion. CI evidence (itemize.log) shows oc-rsync already emits NAME2 unchanged rows for bar/, foo/ in local mode; the only divergence is a missing `.d          ./` root-directory row. That gap lives in the generator/local-copy itemize path, not the server flag parser. A separate, test-backed fix will target the actual root-row suppression.

PR #6042, opened 2026-06-21
fix(itemize): suppress server-sender itemize on client display channel

oferchen:
## Summary

A server-sender must not write itemize output to the client display channel. On a pull, oc-rsync is the server-sender; its generator was emitting `<f` itemize rows over MSG_INFO. The client already prints its own receiver-side `>f` rows, so under `-ii` every itemized line was duplicated with conflicting direction arrows.

## Root cause

Upstream routes server-side itemize to the logfile only, never to the client display channel:

- `sender.c:211` — `itemizing = am_server ? logfile_format_has_i : stdout_format_has_i`
- `log.c:816` — stdout itemize gated on `!am_server`
- `log.c:822-836` — `maybe_log_item()` routes server-side itemize to FLOG only

`maybe_emit_itemize` ignored this and unconditionally sent a MSG_INFO frame in the server role.

## Fix

- `maybe_emit_itemize` returns early for the server role (`!client_mode`); only the client (terminal owner) prints, via the FCLIENT callback.
- The now-unused `ServerWriter` parameter is dropped at both call sites in `transfer_loop.rs`.
- The receiver-side push itemize path (`receiver/mod.rs::emit_itemize`) and the client-pull stdout path are unchanged.

## Validation

On the Linux host against the upstream 3.5.0dev testsuite: `itemize`, `exclude`, `exclude-lsh`, `dirs`, `delete`, and `output-options` all PASS. New regression test `server_sender_emits_nothing_even_when_gate_passes` asserts the server role produces no itemize output even when the emit gate is fully satisfied.

### DIFF
--- a/crates/transfer/src/generator/protocol_io.rs
+++ b/crates/transfer/src/generator/protocol_io.rs
@@ -265,36 +265,43 @@ impl GeneratorContext {
         Ok(())
     }
 
-    /// Emits itemize output when conditions are met.
+    /// Emits the sender-side itemize line for one file when the gate allows.
     ///
-    /// In server mode (daemon/SSH), sends the formatted itemize string
-    /// (`"%i %n%L\n"`) as a MSG_INFO multiplexed message to the client.
-    /// In client mode, writes directly to the process stdout via the
-    /// itemize callback, matching upstream's `rwrite()` `FCLIENT` path.
+    /// Only the terminal-owning CLIENT process writes user-visible itemize
+    /// output (via the FCLIENT callback to stdout). A SERVER-sender is a no-op
+    /// here: upstream `sender.c:211` sets `itemizing = am_server ?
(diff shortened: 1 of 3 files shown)


oferchen:
Closing: investigation on the host against the **3.4.4** testsuite (identical to the CI `run_upstream_testsuite.sh` harness) shows this change does not close the residual `itemize`/`exclude`/`exclude-lsh` failures, and the `itemize`/`exclude` cases are *local* transfers where this server-only (`!client_mode`) suppression never executes. The real `itemize` root cause is alt-basis (`--copy-dest`/`--link-dest`/`--compare-dest`) itemize change-flag computation (upstream emits `cf`+spaces for a file copied from the alt-basis; oc-rsync emits `>f+++++++++`). Reopening under a correctly-scoped fix; the server-sender-suppression idea will be re-evaluated separately with wire validation.

PR #6044, opened 2026-06-21
fix(transfer): load per-directory merge filters during --delete pass

oferchen:
## Summary

The receiver `--delete` pass never loaded per-directory merge-filter scopes,
so per-directory merge files (`.rsync-filter`, `:C` dir-merge, `.cvsignore`)
were ignored when deciding which extraneous files to delete. Files that a
per-directory merge rule protects were wrongly deleted, and files such a rule
excludes from the transfer were wrongly kept. This is purely a receiver-side
caller bug in `delete_extraneous_files`.

## Root cause

`crates/transfer/src/receiver/directory/deletion.rs` cloned a single shared
`FilterChain` whose `scopes` stack was always empty because `enter_directory`
was never called on the receiver side. Deletion decisions therefore ran only
against the global `FilterSet`. Upstream `delete_in_dir`
(`generator.c:308`) calls `change_local_filter_dir` -> `push_local_filters`
to reload a directory's merge filters before scanning it for extraneous
entries.

## Why prior fixes missed it

Recent work (#5749/#5751/#5932/#5981/#6015/#6041) fixed the generator walk
and the decision-context synthetic-descendants logic. None of those touched
the receiver delete pass, which is a separate caller that owns its own
filter-chain clone and never entered any directory. The gap was on the
deletion-gate side, not the generator/decision-context side.

## Fix (per-worker owned chain clone)

The deletion scan is a parallel `map_blocking` whose closure is `Fn + Sync`,
so a shared `Arc<FilterChain>` cannot be mutated. `FilterChain` derives
`Clone` and its per-directory scopes are path-local (a scope for directory D
depends only on D's path and the merge files on the path to D, never on a
sibling worker). So:

- A filter-chain template is built once with `transfer_root` set to the
  destination root (re-anchors leading-`/` merge rules to the merge file's
  own directory).
- Each parallel worker clones the template into its own owned, mutable chain.
- Before listing a directory, the worker walks from the transfer root down
  to the scan target and calls `enter_directory` on each component, loading
  inheriting parent merge files plus the target's own. Scopes live in the
  owned clone and are dropped when the closure iteration ends, mirroring
  upstream's push/pop.
- `enter_directory` errors are classified through the existing
  `classify_scan_error` helper (EACCES/NotFound non-fatal, every other class
  fail-loud), never silently skipped.
- The per-entry deletion gate now evaluates `allows_deletion` against the
  per-worker chain with scopes loaded.

The generator walk and decision-context descendant logic are untouched.

## Test

Adds `per_dir_merge_filter_protects_during_deletion` to the receiver deletion
filter-chain test module. It sets up a destination subtree where a
`.rsync-filter` (`- *.deep`) protects an extraneous file, runs the delete
pass, and asserts the protected file survives while an unprotected extraneous
sibling is deleted. Deterministic (present/absent assertions, no mtime
dependence).

### DIFF
--- a/crates/transfer/src/receiver/directory/deletion.rs
+++ b/crates/transfer/src/receiver/directory/deletion.rs
@@ -98,14 +98,22 @@ impl ReceiverContext {
         // upstream: main.c:1367 - deletion_count >= max_delete
         let deletions_performed = Arc::new(AtomicU64::new(0));
 
-        // Share directory children map and filter chain across workers.
-        // The filter chain clone captures the global rules snapshot for
-        // allows_deletion() checks. Per-directory merge files are not
-        // re-read during deletion (upstream also only evaluates the
-        // pre-loaded filter list in delete_in_dir).
+        // Share directory children map across workers, plus a filter-chain
+        // template each worker clones into an owned, mutable chain.
+        //
+        // Each parallel worker re-reads the per-directory merge files for the
+        // directory it is about to scan via `enter_directory`, mirroring
+        // upstream's per-directory filter reload before deletion. The template
+        // carries the global rules plus the registered merge configs; the
+        // transfer root is set to the destination directory so leading-`/`
+        // merge rules re-anchor to the merge file's own directory.
+        // upstream: generator.c:308 change_local_filter_dir in delete_in_dir -
+        // push_local_filters() reloads `.rsync-filter`/`:C`/dir-merge files
+        // for each directory before it is scanned for extraneous entries.
         let dir_children = Arc::new(dir_children);
-        let filter_chain = Arc::new(self.filter_chain.clone());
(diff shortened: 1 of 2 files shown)


oferchen:
Closing: investigation determined the exclude-lsh divergence is sender-side file-list construction (per-directory merge rule anchoring/scoping in the filters FilterChain), not the receiver --delete pass this PR modifies. Reproduced in isolation: `rsync -av -f 'dir-merge .filt'` with foo/.filt containing `- /file1` drops foo/file1 upstream but keeps it in oc-rsync, with a fresh empty destination where --delete is a no-op. The fix belongs in crates/filters/src/chain/mod.rs (per-dir-merge re-anchoring + scoping), mirrored in the engine local_copy path for the local exclude test. Superseded by a filters-crate fix.

PR #6046, opened 2026-06-21
fix(engine): correct per-directory merge traversal semantics

oferchen:
## Summary

Fixes three per-directory merge (`dir-merge`) behaviours in the engine local-copy traversal so that the LOCAL legs of the upstream `exclude` testsuite pass. This is the local-copy path only; the receiver/network path is untouched. The bugs cause files that upstream excludes (`foo/file1`, `mid/one-in-one-out`, `bar/down/to/bar/baz/file5.deep`) to be wrongly kept and transferred during a plain copy.

## Root causes and fixes

1. **Anchored pattern in a per-dir merge anchors at transfer root, not the merge file's directory.** A leading-`/` rule loaded from `foo/.filt` (e.g. `- /file1`) is now rewritten to `/foo/file1`, mirroring `add_rule` prepending `dirbuf + module_dirlen` (`exclude.c:200-207`). Adds `FilterRule::with_pattern` and threads the directory's transfer-root-relative path into the merge loader.

2. **A `dir-merge`/`:C` directive registered while loading a directory's own merge file is only applied to subdirectories.** A growing active-rule loop now re-scans newly registered nested rules against the current directory, mirroring `push_local_filters` walking the growing `mergelist_cnt` (`exclude.c:787-789`). This makes `:C` in `mid/.filt` load `mid/.cvsignore` for `mid/` itself, and a same-directory `dir-merge .filt2` apply to that directory.

3. **Loaded nested dir-merge segments do not inherit into subdirectories.** Inheritable loaded segments now accumulate into child frames so a rule loaded at an ancestor depth keeps matching descendants, mirroring `push_local_filters` setting `lp->tail = NULL` while keeping `lp->head` (`exclude.c:801`). Non-inheritable (`n`/`C` modifier) segments are dropped from child frames.

## Code sites changed

- `crates/filters/src/rule.rs` — add `FilterRule::with_pattern` builder (preserves all other attributes).
- `crates/engine/src/local_copy/context.rs` — `LoadedDynamicSegment` (segment + inherit flag), `LoadedNestedDirMerge`, and `anchor_dir_merge_rule` (re-anchors a leading-`/` pattern at the merge file's directory).
- `crates/engine/src/local_copy/context_impl/transfer.rs` — thread `relative_dir` into `enter_directory*`; seed child frames with inheritable parent segments; growing active-rule loop loads same-directory nested dir-merges; `load_nested_dir_merge` helper.
- `crates/engine/src/local_copy/context_impl/options/filter.rs` — read the new tagged loaded-segment shape.
- `crates/engine/src/local_copy/executor/directory/recursive/mod.rs`, `crates/engine/src/local_copy/executor/cleanup.rs` — pass the directory's relative path to the merge loader.

## Tests

New engine integration test `crates/engine/tests/per_dir_merge_traversal.rs`:
- `anchored_rule_in_per_dir_merge_anchors_at_merge_file_directory` — `- /file1` in `foo/.filt` excludes `foo/file1` but not `foo/sub/file1`.
- `cvs_modifier_in_per_dir_merge_loads_cvsignore_for_current_dir` — `:C` in `mid/.filt` loads `mid/.cvsignore` and excludes `one-in-one-out`.
- `same_dir_nested_dir_merge_applies_to_current_dir_and_descendants` — `dir-merge .filt2` in `bar/.filt` applies `bar/.filt2` (`- *.deep`) to `bar/baz/file5.deep`.

Each test encodes the upstream parity rationale in its rustdoc.

### DIFF
--- a/crates/engine/src/local_copy/context.rs
+++ b/crates/engine/src/local_copy/context.rs
@@ -448,6 +448,59 @@ include!("context_impl/transfer.rs");
 include!("context_impl/delta_transfer.rs");
 include!("context_impl/reporting.rs");
 
+/// Outcome of loading one nested `dir-merge` file in a directory: the compiled
+/// rule segment (absent when the file had no concrete rules), any
+/// `exclude-if-present` markers, and any further nested dir-merge rules it
+/// registered.
+struct LoadedNestedDirMerge {
+    segment: Option<FilterSegment>,
+    markers: Vec<ExcludeIfPresentRule>,
+    nested: Vec<NestedDirMerge>,
+}
+
+/// Rewrites an anchored per-dir merge rule so its leading `/` anchor binds to
+/// the merge file's directory (`relative_dir`) rather than the transfer root.
+///
+/// upstream: exclude.c:200-207 `add_rule` - for a rule loaded via
+/// `parse_filter_file` with `XFLG_ANCHORED2ABS`, a leading-`/` pattern gets the
+/// directory prefix `dirbuf + module_dirlen` (length `dirbuf_len - module_dirlen
+/// - 1`) prepended, so `- /file1` in `foo/.filt` matches `foo/file1`. Unanchored
+/// patterns and rules with no directory context are returned unchanged.
+fn anchor_dir_merge_rule(rule: FilterRule, relative_dir: Option<&Path>) -> FilterRule {
+    let Some(dir) = relative_dir else {
+        return rule;
+    };
+    if dir.as_os_str().is_empty() {
+        return rule;
+    }
+    let pattern = rule.pattern();
+    let Some(rest) = pattern.strip_prefix('/') else {
+        return rule;
+    };
+    // `dir` is the transfer-root-relative directory of the merge file; build the
+    // anchored pattern `/<dir>/<rest>` using forward slashes (filter patterns are
+    // always `/`-delimited regardless of platform separator).
+    let dir_str = dir.to_string_lossy().replace('\\', "/");
+    let dir_str = dir_str.trim_matches('/');
+    let new_pattern = if rest.is_empty() {
+        format!("/{dir_str}")
+    } else {
+        format!("/{dir_str}/{rest}")
+    };
+    rule.with_pattern(new_pattern)
+}
+
+/// A filter segment loaded from a runtime-registered `dir-merge` file, paired
+/// with whether the originating rule inherits into subdirectories.
+#[derive(Clone, Debug)]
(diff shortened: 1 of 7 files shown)


oferchen:
## Status update: reverted the deletion-pass commits; traversal fix retained

I pushed two follow-up commits earlier today that tried to extend this PR to also fix the `--delete` pass for `exclude.test` / `exclude-lsh.test`. Host validation against the real upstream testsuite showed they were the wrong shape, so I have force-reset this branch back to the traversal fix (`c663cd1ba` + the lint follow-up). The branch now contains only the original, tested per-directory merge traversal correction.

### Why the deletion commits came out

The two reverted commits made the deletion evaluator ignore source-side per-directory merge layers. Validated on the Linux host against `testsuite/exclude.test`, that traded one bug for a worse one:

- Before: oc-rsync **under**-deleted (`foo/file1`, `bar/.../baz/file5.deep` kept though upstream removes them).
- With those commits: oc-rsync **over**-deleted, removing destination files upstream keeps (`.filt` / `.filt2` per-dir merge files and `nodel.deep`). Over-deletion is data loss, so it is strictly worse and not acceptable.

### What the real fix needs (for a follow-up)

Reading upstream `delete.c` / `exclude.c`, the generator's delete pass:

1. Maintains a depth-indexed local-filter stack (`change_local_filter_dir`), and on entering each **destination** directory **re-reads** that directory's per-dir merge files from the receiving tree (`push_local_filters` -> `set_filter_dir(dest dir)` -> `parse_filter_file`). Rules are not inherited from the sender's flist walk; they are reloaded per directory.
2. Protects the merge files themselves via `FILTRULE_EXCLUDE_SELF`, which is why `.filt` / `.filt2` are never deleted.
3. Decides which rules apply through a side/elide mechanism (sender-side vs receiver-side folded into `ent->elide`, compared against `cur_elide_value`) rather than a dedicated "deletion" flag. `--delete-excluded` flips unsided non-perdir rules to elided so excluded files become deletable.
4. Deletes a destination entry only when it is absent from the sender's already-filtered file list.

That is a focused but non-trivial change to the engine local-copy deletion path (effectively reproducing the destination-side per-directory merge reload with depth-stack scoping). It is the correct direction toward consolidating on the canonical filter chain, and it should land as its own change rather than be bolted on here.

The `exclude` / `exclude-lsh` upstream-testsuite legs remain red pending that work; they are not part of the required check set. This PR's traversal fix is a net improvement on its own and its required checks stand on their own.

PR #6049, opened 2026-06-21
fix(itemize): suppress server-side itemize emission to match upstream

oferchen:
## Problem

The upstream testsuite `exclude-lsh.test` fails on its `-aiiO --update` leg. With oc-rsync as the remote `--server --sender` (invoked via the test's `lsh.sh` remote shell) and upstream rsync as the local client-receiver, oc-rsync emits an extra sender-direction itemize line for every transferred file:

```
  >f+++++++++ extra-src
+ <f+++++++++ extra-src      # oc-rsync over-emits
  >f..t...... src-newness
+ <f..t...... src-newness    # oc-rsync over-emits
```

Upstream prints only the receiver `>f...` rows.

## Root cause

Upstream `log.c:812 log_item()` gates the per-file itemize emission to the client on `!am_server`:

```c
if (code != FLOG && stdout_format && !am_server)
    log_formatted(FCLIENT, stdout_format, s_or_r, file, NULL, iflags, hlink);
```

A server process (sender **or** receiver) never renders its own itemize row; the client always computes and prints it (the generator ships the `iflags` over the wire, not the rendered text). oc-rsync's `maybe_emit_itemize` instead multiplexed the line as a `MSG_INFO` frame whenever the process was not in client mode, so the server-sender's `<f...` lines leaked to the client and duplicated its `>f...` rows under `-ii`.

## Fix

`maybe_emit_itemize` now returns early when the process is a server (`!client_mode`, oc-rsync's equivalent of `am_server`), mirroring upstream's `!am_server` gate. The client-mode callback path that local transfers and the client-receiver/client-sender use is unchanged.

A regression test (`server_mode_does_not_emit_itemize`) asserts the server role emits to neither the itemize callback nor the wire.

## Verification (Linux host, upstream testsuite vs upstream rsync)

- `exclude-lsh` — now **PASS** (was FAIL)
- No regression: `exclude`, `itemize`, `dirs`, `output-options`, `00-hello`, `delete`, `daemon` all PASS
- `cargo nextest run -p transfer -E 'test(itemize_emit_gate)'` — 4 passed

### DIFF
--- a/crates/transfer/src/generator/protocol_io.rs
+++ b/crates/transfer/src/generator/protocol_io.rs
@@ -267,10 +267,14 @@ impl GeneratorContext {
 
     /// Emits itemize output when conditions are met.
     ///
-    /// In server mode (daemon/SSH), sends the formatted itemize string
-    /// (`"%i %n%L\n"`) as a MSG_INFO multiplexed message to the client.
-    /// In client mode, writes directly to the process stdout via the
-    /// itemize callback, matching upstream's `rwrite()` `FCLIENT` path.
+    /// Emits the per-file itemize line, but ONLY in client mode. Upstream
+    /// `log.c:812 log_item()` gates the `FCLIENT` itemize emission on
+    /// `!am_server`: a server process (sender OR receiver) never prints its own
+    /// itemize row - the client always computes and prints it (the generator
+    /// ships the `iflags` over the wire, not the rendered text). Emitting from
+    /// the server duplicated the client's row, producing a spurious sender
+    /// `<f...` line alongside the receiver `>f...` line under `-ii` over a
+    /// remote shell (upstream `exclude-lsh.test` `-aiiO --update` leg).
     ///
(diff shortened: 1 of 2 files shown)


oferchen:
Closing: CI ground truth on upstream rsync 3.4.4 shows this change has no effect on the failing tests it claimed to address.

- Master (34cc8898) upstream-testsuite: 49 passed / 2 failed (exclude, exclude-lsh).
- This branch (master + this commit): identical — 49 passed / 2 failed (exclude, exclude-lsh).

The itemize test passes on both; exclude/exclude-lsh fail on both. The premise that server-side itemize emission was double-emitting and causing the exclude-lsh failure is not supported — the failures are a 3.4.4-specific divergence in the filter/delete path, unrelated to itemize. The earlier host verification used upstream 3.4.3 (where exclude-lsh passes), which masked this.

Reopening the real investigation against upstream 3.4.4 separately.

PR #6050, opened 2026-06-21
fix(cli): accept repeated -i/--itemize-changes like upstream

oferchen:
## Problem

Upstream rsync treats `-i`/`--itemize-changes` as a repeatable counter (`-ii` raises itemize detail). oc-rsync declared the arg with `ArgAction::SetTrue`, so clap rejected any repeated occurrence with `the argument '--itemize-changes' cannot be used multiple times`. Invocations like `-rii` or `-aii` exited 1 before performing any work — a drop-in incompatibility with upstream.

## Fix

- Switch `--itemize-changes` from `ArgAction::SetTrue` to `ArgAction::Count`.
- Read it via `get_count() > 0` instead of `get_flag()`.
- The mutual `overrides_with` between `itemize-changes` and `no-itemize-changes` is preserved, so last-wins toggling is unchanged.

## Tests

Adds `parse_args_accepts_repeated_itemize_changes` asserting repeated `-i` parses successfully. Existing override tests (`parse_args_prefers_last_itemize_toggle`) continue to validate last-wins semantics under the new action.

### DIFF
--- a/crates/cli/src/frontend/arguments/parser/mod.rs
+++ b/crates/cli/src/frontend/arguments/parser/mod.rs
@@ -601,7 +601,7 @@ where
         } else {
             ProgressSetting::Unspecified
         };
-    let itemize_changes_flag = matches.get_flag("itemize-changes");
+    let itemize_changes_flag = matches.get_count("itemize-changes") > 0;
     let no_itemize_changes_flag = matches.get_flag("no-itemize-changes");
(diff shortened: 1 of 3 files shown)


viktoriavikulya (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

PR #6071, opened 2026-06-23
fix(client): apply --prune-empty-dirs over remote shell

oferchen:
## Summary

`--prune-empty-dirs` (`-m`) was silently ignored for remote-shell (SSH) pulls. Empty directories whose contents are all excluded survived in the destination, where upstream rsync prunes them.

## Root cause

Upstream treats `prune_empty_dirs` as a receiver-side global: `flist.c::flist_sort_and_clean` (`clean_flist`) drops empty-directory chains from the file list on the receiver, for every transport. It is deliberately NOT part of the compact server flag-letter string, because that string is parsed by the remote *sender*, which must ship all directories.

On the remote-shell pull path the local process is the receiver, and its prune pass (`receiver::file_list::prune::prune_empty_dirs_pass`, gated on `config.flags.prune_empty_dirs`) only ran when the flag was set on its `ServerConfig`. `apply_common_server_flags` propagated itemize and several other receiver-local flags but never `prune_empty_dirs`, so the gate was always false for SSH pulls. Local transfers set the flag through a different path and were unaffected.

## Fix

Propagate `config.prune_empty_dirs()` into the receiver `ServerConfig` in `apply_common_server_flags`, alongside the existing itemize propagation. This makes the local receiver's prune pass run on remote-shell pulls, matching upstream `clean_flist` semantics. The flag is receiver-local; nothing changes on the wire.

## Tests

- `apply_common_server_flags_sets_prune_empty_dirs` - flag propagates when `-m` is set.
- `apply_common_server_flags_prune_empty_dirs_default_false` - default stays off.

Covers the upstream `exclude-lsh.test` `--prune-empty-dirs` leg that fails over remote shell.

### DIFF
--- a/crates/core/src/client/remote/flags.rs
+++ b/crates/core/src/client/remote/flags.rs
@@ -302,6 +302,14 @@ pub(crate) fn apply_common_server_flags(config: &ClientConfig, server_config: &m
     // as --log-format=%i, but the local ServerConfig also needs the flag set so
     // the generator's maybe_emit_itemize() produces client-side output via callback.
     server_config.flags.info_flags.itemize = config.itemize_changes();
+    // upstream: flist.c::flist_sort_and_clean - prune_empty_dirs is a global the
+    // RECEIVER acts on (clean_flist drops empty-dir chains) regardless of
+    // transport. It is deliberately not part of build_server_flag_string because
+    // that compact string is parsed by the remote SENDER, which must ship all
+    // dirs. Propagate it here so the local receiver's flist prune pass
+    // (receiver::file_list::prune::prune_empty_dirs_pass) runs on remote-shell
+    // pulls; otherwise empty directories survive that upstream would have pruned.
+    server_config.flags.prune_empty_dirs = config.prune_empty_dirs();
     // upstream: flist.c::iconv_for_local and options.c::recv_iconv_settings -
     // when --iconv is configured, the local process must transcode file-list
     // entries between the local and remote charsets. Without this bridge the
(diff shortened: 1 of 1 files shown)


oferchen:
Closing: this is a regression, not a fix.

`exclude-lsh` passes on master (5 consecutive Upstream Testsuite runs green, incl. HEAD 7810e466) and fails on this PR. The only behavioral change here is one line propagating `prune_empty_dirs` to the remote-shell receiver's ServerConfig, which turns on the receiver-side flist prune pass for SSH pulls.

Root cause of the regression: oc-rsync's `prune_empty_dirs_pass` is **more aggressive** than upstream rsync's `clean_flist`. exclude-lsh.test:121-123 asserts that `--prune-empty-dirs` (no filters) produces the same tree as `-av --exclude foo/down/ foo/too/ foo/and/ new/`. With the prune pass enabled over the remote-shell path, oc's output drops directories the expected tree keeps (CI artifact diff: `to/` missing `foo/down`, `mid/for/foo/and`, `new/lose`, `new/keep/this`). A local `--prune-empty-dirs` comparison confirms upstream prunes only the genuinely-empty chains while oc's pass over-prunes.

The flag-propagation gap is technically real, but enabling it is net-negative until `prune_empty_dirs_pass` is corrected to match upstream `clean_flist` semantics. Since master is green on exclude-lsh, there is no failing-test pressure forcing this; the underlying prune-pass divergence should be a separate, properly-scoped investigation. The change is receiver-local (no wire impact), so no protocol concern.

PR #6088, opened 2026-06-24
ci: correct daemon known-failure rationale (source-only operand, not IPv6)

oferchen:
## Summary

Corrects a stale rationale comment in `tools/ci/upstream_testsuite_known_failures.conf` for the `daemon` test. The comment claimed the failure was an exit-10 IPv6-loopback bind issue on GitHub Actions runners. That is no longer accurate.

## Actual cause (verified)
The upstream `daemon.test` runs several legs under `bash -e`, so the first failing leg aborts the script. Legs 1-2 (module listing over rsh, trailing-tab) now pass. Leg 3 (`rsync -r localhost::test-hidden` - a source with no destination) fails: oc-rsync does not treat a single source-only daemon operand as implied `--list-only`. Upstream implies it via `options.c:2194-2195` (`if (argc < 2 && !read_batch && !am_server) list_only |= 1;`), but oc-rsync rejects the lone operand at `crates/core/src/client/remote/daemon_transfer/mod.rs:71-76` ("need at least one source and one destination", exit 1).

The harness classifies known failures by test **name**, not exit code, so the entry already marks the test XFAIL correctly - only the comment was wrong.

## Scope
Comment-only change. The `"daemon"` known-failure entry itself is unchanged. The real fix (imply `--list-only` for a single remote/daemon source operand) is tracked separately.

### DIFF
--- a/tools/ci/upstream_testsuite_known_failures.conf
+++ b/tools/ci/upstream_testsuite_known_failures.conf
@@ -15,18 +15,23 @@
 
 KNOWN_FAILURES=(
 
-  # "daemon" - CI-environment-only failure on GitHub Actions hosted runners.
-  #   The upstream daemon.test exits with code 10 (daemon startup) because
-  #   GitHub Actions runners disable IPv6 on the loopback interface; the
-  #   default dual-stack IPv6-first bind path the daemon takes when handed
-  #   port=0 cannot acquire `::1`, surfaces only as an IPv4 listener, and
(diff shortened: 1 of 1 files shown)


oferchen:
Superseded by #6083, which already rewrote the daemon known-failure rationale with more detail (covering both the daemon-over-rsh listing fix for legs 1-2 and the leg-3 source-only implied --list-only cause). Master's comment is accurate and more complete; closing this redundant change.

PR #6116, opened 2026-06-26
test(cli): fix list-only column positions for 14-wide size field (missed in #6109)

oferchen:
#6109 narrowed the `--list-only` size column 15->14 but two list_only tests still asserted the old 15-wide byte offsets (`&line[11..26]`, byte 26 = space, `[27..46]` timestamp, `[47..]` name), failing on the macOS cell (which runs the full cli suite). With the 14-wide size field, every offset after the size shifts down by one. Updated the hardcoded positions + `15->14` length assertions. Test-only; greens master so the in-flight PRs (#6110/#6113/#6114/#6115) stop inheriting the failure.


oferchen:
Superseded by #6117, which folds this list_only fix together with the progress_live.rs fix so a single PR clears both master-RED cli test files (each PR alone would still trip the other's RED).

PR #6131, opened 2026-06-26
fix(engine): exit 23/link_stat for missing local source argument

oferchen:
## Problem

A local source argument that does not exist exits 24 (`RERR_VANISHED`) with `file has vanished: "..."`. Upstream rsync exits 23 (`RERR_PARTIAL`) with `link_stat "%s" failed: %s` for a source argument that fails its initial `link_stat`, reserving `RERR_VANISHED` for files that disappear *mid-transfer*.

```
# upstream:
$ rsync /tmp/nope dst/ ; echo $?
rsync: [sender] link_stat "/tmp/nope" failed: No such file or directory (2)
... (code 23) ...
23
```

The missing-source case was routed through the shared `io_error` helper, whose NotFound branch correctly maps to 24 for genuine mid-run vanishes (`generator`/recursive paths). Changing that helper globally would regress those.

## Fix

Add a dedicated `LocalCopyErrorKind::LinkStatFailed` used **only** at the initial source-stat site (the `NotFoundError` arm in the local-copy executor source orchestration). It maps to `PartialTransfer` (23) with the upstream `link_stat` wording and the `[sender]` role. The shared `io_error` NotFound->Vanished path is untouched, so mid-transfer vanishes still exit 24.

## Tests

- `engine`: `link_stat_failed_is_partial_not_vanished` - exit 23, `RERR_PARTIAL`, not classified as vanished/io.
- `core`: `map_link_stat_failed_uses_partial_not_vanished` - maps to `PartialTransfer` with `link_stat "..." failed` wording, not `file has vanished`.
- Existing `io_error_notfound_uses_vanished_code` stays valid (mid-run vanish unchanged).

Built + clippy clean (engine, core).

### DIFF
--- a/crates/core/src/client/error.rs
+++ b/crates/core/src/client/error.rs
@@ -197,6 +197,17 @@ pub(crate) fn map_local_copy_error(error: LocalCopyError) -> ClientError {
             path,
             source,
         } => io_error(action, &path, source),
+        LocalCopyErrorKind::LinkStatFailed { path, source } => {
+            // upstream: flist.c send_file_list() - a source argument that fails
+            // its initial link_stat is reported by the sender as
+            // `link_stat "%s" failed: %s` and exits RERR_PARTIAL (23). This is
+            // deliberately not routed through io_error(), whose NotFound branch
+            // maps to RERR_VANISHED (24) for files that vanish mid-transfer.
+            let code = ExitCode::PartialTransfer;
+            let text = format!("link_stat \"{}\" failed: {source}", path.display());
+            let message = rsync_error!(code.as_i32(), text).with_role(Role::Sender);
+            ClientError::with_code(code, message)
+        }
         LocalCopyErrorKind::Timeout { duration } => {
             let code = ExitCode::Timeout;
             let text = format!(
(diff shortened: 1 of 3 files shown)


oferchen:
Closing after systematic debugging of the CI failures (Windows stable, interop, musl).

**Root cause:** the fix is mis-scoped. `process_single_source`'s `NotFoundError` arm is NOT exclusively the initial command-line source stat — both command-line operands AND `--files-from` entries become `plan.sources()` and flow through the same arm. The caller (`copy_sources`) distinguishes skip-and-continue from fatal purely via `is_vanished_error()`/`is_io_error()`.

Upstream genuinely treats the two cases differently:
- a **command-line** source arg that fails `link_stat` → exit 23 (`RERR_PARTIAL`, `link_stat "%s" failed`) — verified live against rsync 3.4.3.
- a **`--files-from`** entry that's missing → exit 24 (`RERR_VANISHED`, `file has vanished`), skipped while the remaining files still transfer.

The new `LinkStatFailed` kind is neither vanished nor io_error, so the caller's `else` branch aborts the whole transfer — which broke `tools/ci/run_interop.sh:test_file_vanished` (stable files no longer transferred) and conflicts with the `files_from_vanished.rs` unit tests that correctly assert the 24/skip-and-continue path.

**Correct fix (future):** thread source *provenance* (command-line vs `--files-from`) down to `process_single_source` and only emit the exit-23 `link_stat` error for a command-line operand; keep the vanished-skip-continue (24) path for `--files-from` entries. That is a non-trivial, design-level change, not the surgical polish this PR attempted. Tracked in the task notes for #473.

PR #6172, opened 2026-06-28
fix(transfer): record short final basis block at remainder length

oferchen:
## Problem

`generate_delta_from_signature` (sender-side delta reconstruction from a received signature) rebuilt **every** basis block at the full `block_length`, including the final partial block.

When a basis file's length is not a whole multiple of `block_length`, the receiver's last block covers only `SumHead.remainder` bytes. But the matcher guards each candidate window with `block.len() == block_length` (`matching/src/index/builder.rs`), so a full-length source window could match against the short final block and emit a `COPY { index, len }` token that over-reads the basis tail — corrupting the reconstructed file.

This is a latent data-corruption bug: it only triggers when the final basis block is short **and** a full-length source window happens to collide on the short block's rolling+strong checksums.

## Fix

Thread `SumHead.remainder` through `DeltaGeneratorConfig` and record the final block at its true length (`remainder` when `index == block_count - 1`), mirroring upstream `match.c`, which sizes the last block as `len - (block_count-1)*block_length`. Full-length source windows can no longer false-match the short final block.

## Changes

- `delta_config.rs`: add `remainder: u32` field + `with_remainder()` builder.
- `generator/delta.rs`: size the final reconstructed block at `remainder`; compute `total_bytes` from `(block_count-1)*block_length + remainder`.
- `generator/transfer/transfer_loop.rs`: pass `sum_head.remainder` into the config.

## Test

`final_short_block_excluded_from_full_window_match` builds a wire signature whose final block is shorter than `block_length` and asserts no `COPY` token references the short block's basis region, with a positive control that a full earlier block still matches.

Verified non-vacuous: reverting the `block_len` computation makes the test fail with tail-corruption detected.

### DIFF
--- a/crates/transfer/src/delta_config.rs
+++ b/crates/transfer/src/delta_config.rs
@@ -35,6 +35,14 @@ pub struct DeltaGeneratorConfig<'a> {
     /// rolling and a strong checksum.
     pub sig_blocks: Vec<protocol::wire::signature::SignatureBlock>,
 
+    /// Size in bytes of the final partial block (0 when the basis length is a
+    /// whole multiple of `block_length`). Carried on the wire in the
+    /// `SumHead.remainder` field. The final block must be recorded at this
+    /// length, not at `block_length`, so a full-length source window cannot
+    /// falsely match the short final block - see
+    /// [`crate::generate_delta_from_signature`].
+    pub remainder: u32,
+
     /// Length of the strong checksum in bytes. Common values are 16 (MD5,
     /// MD4) or 20 (SHA-1); must be non-zero and bounded by the digest size
     /// of the negotiated checksum algorithm.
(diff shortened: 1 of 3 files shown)


oferchen:
Closing: this change is a regression, not a fix.

generate_delta_from_signature records the final basis block, and the matching index builder (crates/matching/src/index/builder.rs:34-52) intentionally skips partial-length blocks, returning None from from_signature when no full-length block remains (builder.rs:107). Recording the final block at remainder length turns it into a partial block; any basis whose only block is that final partial one then fails index creation with "failed to create signature index" (delta.rs:195), aborting every such delta transfer (existing/backup/compress-delta/inplace/append all regressed in interop).

The original behaviour is correct: the receiver computes the final block's rolling checksum over remainder bytes, while the sender only probes block_length-sized windows. A block_length-window rolling sum can never equal a remainder-byte rolling sum, so the short final block cannot be falsely matched - there is no over-read and no corruption. The scenario the regression test constructed does not occur in production. No code change is warranted.

PR #6181, opened 2026-06-28
feat(daemon): wire io_uring socket read/write (default-on, kill-switchable)

oferchen:
## Summary

Routes the daemon transfer socket through fast_io's io_uring `RECV`/`SEND`
factories (`net_reader::for_socket` + `socket_writer_from_fd`) when the kernel
supports and permits it, with a clean fallback ladder by effectiveness and a
runtime kill-switch. Promotes `iouring-send-zc` into the default feature set.

### Effectiveness-ordered, platform + permission-aware selection
- **Send**: io_uring `SEND_ZC` (≥4 KiB, kernel ≥6.0) → io_uring `SEND` (≥5.6) → `std::write`
- **Recv**: io_uring `RECV` (≥5.6) → `std::read`
- **Platform**: `cfg(unix)`; the io_uring backend is Linux-only (stub elsewhere → std)
- **Permission**: `is_io_uring_available()` attempts `io_uring_setup(2)` and returns
  `SyscallBlocked` (→ std) under rootless containers, `io_uring_disabled` sysctl, or a
  seccomp policy without the io_uring bucket. The daemon's own seccomp allowlist already
  includes `io_uring_{setup,enter,register}`, so the path is sandbox-consistent.

### Safety
- Owning adapters keep each cloned `TcpStream` alive for the borrowed fd.
- `OC_RSYNC_IOURING_NET=0/false/no/off` forces the standard blocking-socket path
  (byte-for-byte upstream's model) for bisection.
- The SEND path is POLLOUT-gated with a linked timeout (`batching.rs`), so a
  back-pressured send during the bidirectional teardown surfaces `WouldBlock` and
  re-arms instead of reproducing the prior `submit_and_wait` hang.

## Host validation (aarch64, kernel 7.0, upstream rsync 3.4.3)
- Release build: **passes** (real Linux io_uring types; writer is `Send`).
- Smoke (128 MiB + 30 files, `--stats`, push **and** pull): **rc=0, byte-for-byte OK**, no hang.
- io_uring engagement confirmed via strace: `io_uring_enter` **3424 (on) vs 513 (off)**
  — the kill-switch cleanly disables the socket path (remaining 513 are the disk
  IoUringDiskBatch path).

Throughput micro-benching across a kernel matrix is tracked as NZC-8/NRX-8 follow-ups.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ default = [
     "parallel",
     "copy_file_range",
     "io_uring",
+    "iouring-send-zc",
     "iocp",
     "async",
 ]
--- a/crates/daemon/src/daemon/sections/module_access/transfer.rs
+++ b/crates/daemon/src/daemon/sections/module_access/transfer.rs
@@ -451,6 +451,80 @@ struct TransferStreams {
     supports_tcp_shutdown: bool,
 }
 
+/// Socket buffer size for the io_uring read/write paths, matching the
+/// `BufReader` capacity the transfer engine wraps the reader in.
+#[cfg(unix)]
+const SOCKET_IO_BUF: usize = 64 * 1024;
+
+/// Resolves the io_uring socket-I/O policy from the `OC_RSYNC_IOURING_NET`
+/// kill-switch, cached once via [`OnceLock`].
+///
+/// Default-on: returns [`IoUringPolicy::Auto`] (engage io_uring `RECV`/`SEND`
+/// when the kernel supports it, fall back otherwise). Setting the env var to
+/// `0`/`false`/`no`/`off` forces [`IoUringPolicy::Disabled`] - the standard
+/// blocking-socket path - so a regression can be bisected without a rebuild.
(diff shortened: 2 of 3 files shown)


oferchen:
Holding this until NRX-3 (multishot recv + registered buffers) lands. Host benchmarks show the current single-shot blocking RECV/SEND path regresses or ties on the socket — never wins (loopback push ~9% slower; +20ms RTT push ~24% slower) — so default-on is not yet justified. The wiring, kill-switch, and host safety are all proven; the default decision is deferred to the multishot-recv work that could actually make it win. Marked draft.

oferchen:
Closing: host benchmarks confirmed io_uring on the socket path delivers much lower performance benefit than expected - it regresses or ties versus std blocking sockets (loopback push ~9% slower; +20ms RTT push ~24% slower) and never wins, because single-shot blocking RECV/SEND adds per-op cost without batching. The wiring, kill-switch, and host safety are all proven and preserved on the branch `feat/iouring-net-daemon-wiring`; if revisited, it would need multishot recv + registered buffers (NRX-3) to have a chance of winning. Deferred.

PR #6239, opened 2026-07-01
fix(daemon): allow non-root exec-hook modules by exempting them from worker capability-drop

oferchen:
## Problem

A non-root daemon serving a module configured with `early exec` / `pre-xfer exec` / `post-xfer exec` hooks fails on **Linux** (connection closed, exit code 12, client sees "0 bytes received"). It passes as root, and passes on macOS non-root.

## Root cause

The only root-vs-non-root behavioral fork that runs for a hook module (no uid/gid/chroot configured) is `drop_worker_capabilities` in `capabilities.rs`. Root skips the capability drop; a non-root worker runs it, stripping the capability set the worker needs to `fork`/`execve` the hook helper. The worker thread's swallowed error then drops the socket.

`PR_SET_NO_NEW_PRIVS` (set globally at startup) is **not** the cause: it only blocks privilege *escalation* on `execve` (setuid/caps), never a plain hook `execve`, so no relaxation is needed there and it stays global.

## Fix

Mirrors the existing Landlock skip in `engage_landlock_sandbox` (`sandbox.rs`), which already exempts hook modules because Landlock rulesets are inherited across `exec`. A shared `module_has_exec_hooks` predicate is extracted and reused at all three hardening layers:

1. **`drop_worker_capabilities`** - skip the per-worker capability drop for hook modules (fixes the observed failure). Non-hook modules keep the full drop.
2. **`engage_landlock_sandbox`** - reuse the shared predicate (now also covers `early exec`, previously only `pre`/`post`).
3. **worker seccomp allowlist** - add `execve`/`execveat` **only** when the module has hooks (`allow_exec` threaded from `module_has_exec_hooks`), so hook modules work with seccomp ON without weakening the steady-state allowlist for normal modules.

Upstream rsync 3.4.4 has no capability/seccomp/Landlock/prctl code at all - its hooks run with the daemon's inherited privileges. These layers are oc-only defense-in-depth, so exempting operator-configured hook modules preserves upstream behavioural parity.

## Invariants preserved

- **Non-hook modules: security posture unchanged** - still capability-dropped, still seccomp-restricted (exec-free allowlist), still Landlock-sandboxed.
- **Non-Linux + root: zero behavior change** - all new logic is `#[cfg(target_os = "linux")]` and non-root-gated.
- The relaxation is tightly scoped to operator-configured hook modules, exactly the same trust model as the pre-existing Landlock skip.

## Tests

- `module_has_exec_hooks_matches_each_hook_directive` - predicate true for each of early/pre/post, false otherwise.
- `capability_drop_skipped_only_for_hook_modules` - drop gated by the predicate; panic-free on both paths.
- `exec_syscalls_gated_on_hook_modules` + `allowlist_is_a_concrete_set` - `execve`/`execveat` present **only** when `allow_exec == true`; steady-state allowlist never permits exec.

## Verification

- Local (macOS): `cargo build -p daemon --all-targets`, `cargo fmt -p daemon --check`, `cargo clippy` (no findings in changed files) all pass.
- The Linux-only hardening code path (`capabilities.rs`, seccomp allowlist internals) is `#[cfg(target_os = "linux")]` and cannot be compiled or the end-to-end non-root repro run on macOS. **The end-to-end non-root Linux repro must be verified on the Linux validation host and CI.**

### DIFF
--- a/crates/daemon/src/daemon/sections/capabilities.rs
+++ b/crates/daemon/src/daemon/sections/capabilities.rs
@@ -39,6 +39,30 @@ const CAP_FEATURE_AVAILABLE: bool = true;
 #[allow(dead_code)]
 const CAP_FEATURE_AVAILABLE: bool = false;
 
+/// Returns true when the operator configured this module with any exec hook
+/// (`early exec`, `pre-xfer exec`, or `post-xfer exec`).
+///
+/// Single source of truth for the "operator opted into running hooks"
+/// predicate. When true, the worker will `fork(2)`/`execve(2)` a helper
+/// script during the connection, so the hardening layers that would break
+/// that exec (worker capability drop, and - via the same condition -
+/// Landlock, which is inherited across `exec`) are relaxed for this module
+/// only. Modules without hooks keep the full defense-in-depth posture.
+///
+/// upstream: rsync 3.4.4 has no capability/seccomp/Landlock/prctl code at
+/// all - the exec hooks (`clientserver.c` `pre_exec_hook` / `post_exec_hook`
+/// / `early_exec`) run with the daemon's inherited privileges. This oc-only
+/// defense-in-depth must therefore exempt operator-configured hook modules to
+/// preserve upstream behavioural parity: a non-root daemon dropping worker
+/// capabilities would otherwise break the hook `execve`, which upstream never
+/// does. Mirrors the Landlock skip at `sandbox.rs` `engage_landlock_sandbox`,
+/// which reuses this same predicate.
+fn module_has_exec_hooks(module: &ModuleRuntime) -> bool {
+    module.early_exec.is_some()
+        || module.pre_xfer_exec.is_some()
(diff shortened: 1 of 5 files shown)


oferchen:
Closing: the 3.5.0dev tracker CI run (28563711344, Linux non-root) shows **daemon-exec PASS** and **daemon-groupmap-wild PASS** — the controlled-CI equivalent of a -j1 host re-run. Combined with daemon-exec passing as root on the host and on macOS non-root, the original -j16 host failure was a **parallelism flake, not a real bug**. This PR's worker-capability-drop exemption therefore fixes a non-bug and relaxes daemon hardening for no benefit, so it should not land.

The one genuinely-correct sliver — adding execve/execveat to the worker seccomp allowlist for hook modules (they'd be SIGSYS-killed if seccomp were enabled) — is a real but separate, low-priority latent fix; it can be reopened as a small standalone PR if we decide to support exec-hook modules under seccomp.

PR #6248, opened 2026-07-02
fix(engine): require --mkpath for missing dest parents in file-to-file copy

oferchen:
## Summary

A plain file-to-file copy into a destination whose parent directory chain does not exist was silently auto-creating that chain. Upstream rsync creates a missing destination path only when `--mkpath` (`mkpath_dest_arg`) is given; otherwise it fails and creates nothing.

Root cause: the local-copy executor's parent-creation gate was

```rust
let allow_creation = self.implied_dirs_enabled() || self.mkpath_enabled();
```

`implied_dirs` defaults to on, so every file-to-file copy auto-created deep parents. In upstream, `implied_dirs` only governs implied DIR entries under `-R` / `--relative`; it never governs plain file-to-file parents.

## Fix

Gate the `implied_dirs` contribution on relative mode, matching upstream:

```rust
let allow_creation = (self.implied_dirs_enabled() && self.relative_paths_enabled())
    || self.mkpath_enabled();
```

- Without `--mkpath` and without `-R`: a missing deep dest parent now fails and creates nothing (matches upstream `main.c:736`, which only calls `make_path()` when `mkpath_dest_arg` is set).
- With `--mkpath`: the chain is created and the copy succeeds.
- `-R` / `--relative` implied-dir creation is preserved: the gate keeps `allow_creation` true under relative mode, so implied intermediate directories are still materialized.

## Upstream reference

- `main.c:735-749` - `if (mkpath_dest_arg && statret < 0 && (cp || file_total > 1)) { make_path(...) }`. The missing destination path is created only under `--mkpath`.
- `implied_dirs` is consulted only alongside `FLAG_IMPLIED_DIR` / `relative_paths` (`generator.c:1251,1329,1422,2114`; `flist.c:2468`; `options.c:2976` only forwards `--no-implied-dirs` when `relative_paths && !implied_dirs`).

## Tests

New `crates/engine/tests/mkpath_dest_parent_gate.rs`:

1. file-to-file into a missing deep parent WITHOUT `--mkpath` fails and creates nothing.
2. same copy WITH `--mkpath` succeeds with matching contents.
3. `-R` still creates the implied dir chain without `--mkpath` (regression guard).

## Validation

- `cargo fmt --all -- --check`: clean.
- `cargo clippy -p engine --all-targets --all-features --no-deps --locked -- -D warnings`: clean.

### DIFF
--- a/crates/engine/src/local_copy/context_impl/options/dirs.rs
+++ b/crates/engine/src/local_copy/context_impl/options/dirs.rs
@@ -50,6 +50,19 @@ impl<'a> CopyContext<'a> {
         self.options.omit_link_times_enabled()
     }
 
+    /// Returns `true` when `parent` is a proper descendant of the destination
+    /// root - i.e. an in-tree parent produced by tree reconstruction during a
+    /// recursive/relative transfer, not the destination argument's own path.
+    ///
+    /// upstream: main.c:735-749 - only the destination ARGUMENT's missing path
+    /// (the root itself, `parent == destination_root`, or an ancestor of it) is
+    /// gated on `--mkpath`; parents strictly under an existing destination root
+    /// are always created for the tree (generator.c recv_generator).
+    fn parent_is_within_destination_root(&self, parent: &Path) -> bool {
+        let root = self.destination_root();
+        parent != root && parent.starts_with(root)
+    }
+
     fn parent_relative_to_destination<'p>(&self, parent: &'p Path) -> Option<&'p Path> {
         parent
             .strip_prefix(self.destination_root())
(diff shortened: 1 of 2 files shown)


oferchen:
Deferring this one. Two successive CI runs show the destination-parent-creation change destabilizes the core recursive-copy path: the per-file gate broke normal `-a src/ dst/` (itemize/delete/batch-mode), and the reworked `parent_is_within_destination_root` predicate still blocks legitimate single-level dest creation (`create_missing_destination_directory` nextest fails; batch-mode/delete still red).

Root cause and correct approach for a future PR: oc unconditionally uses `create_dir_all` for a missing destination parent, silently materializing a whole chain. Upstream creates directories one level at a time as recursion descends (each level's parent already exists) and only creates a missing *chain* up front under `--mkpath` (main.c:735 `make_path`). The faithful fix is therefore to use `std::fs::create_dir` (single level; errors if the chain is missing, matching upstream's behaviour without `--mkpath`) and switch to `create_dir_all` only when `--mkpath` is set — no path-predicate needed. That change must be validated against the upstream testsuite (itemize/delete/batch-mode/mkpath), which isn't runnable in this environment, so it's deferred rather than landed fly-blind. The divergence is low severity (oc is over-permissive; no data loss).

PR #6259, opened 2026-07-02
fix(transfer): apply --modify-window in generator dry-run itemize (same_time parity)

oferchen:
## Problem

A `-ain --modify-window=2` dry-run (`-n`) did not absorb a 1s mtime difference: the file was itemized as `>f...` (transfer) instead of unchanged. PR #6252 added `--modify-window` to the receiver quick-check, but the dry-run wire path never reached it.

## Root cause

`build_files_to_transfer` short-circuited on `skip_dest_writes()` (`dry_run || list_only`), returning a bare `ITEM_TRANSFER` for **every** candidate and bypassing `quick_check_matches` (which honours `--modify-window`).

Upstream runs `quick_check_ok()` -> `mtime_differs()` -> `same_time()` (`util1.c:1478`) for a dry-run as well; only `dry_run > 1` (missing parent dir, `generator.c:1290`) short-circuits it. A within-window mtime is therefore treated as unchanged and itemized with `iflags=0`, not `>f`.

## Fix

- Restrict the receiver short-circuit to **list-only**; let `-n` fall through to the normal stat + quick-check path so within-window files are itemized as unchanged.
- Under `dry_run`, suppress only the filesystem mutations that upstream gates on `do_xfers`/`!dry_run`: metadata/ACL/xattr apply on a quick-check match (`generator.c:1814`) and alt-dest copy/link (`generator.c:1033`). The itemize stream is unchanged; the destination is never touched.
- Apply the same symmetric window compare to the itemize **time bit** (`generator.c:526 mtime_differs`) via the shared `same_time()` helper, so a within-window mtime no longer sets `ITEM_REPORT_TIME` (`t`). `window == 0` keeps exact whole-second equality.

## Verification

Byte-for-byte against upstream rsync 3.4.4 over a loopback daemon:

| case | oc-rsync | upstream 3.4.4 |
|---|---|---|
| 1s drift, `--modify-window=2` | *(unchanged, no row)* | *(unchanged, no row)* |
| 3s drift, `--modify-window=2` | `>f..t...... file.txt` | `>f..t...... file.txt` |
| 1s drift, `--modify-window=0` | `>f..t...... file.txt` | `>f..t...... file.txt` |
| identical, `--modify-window=0` | *(unchanged)* | *(unchanged)* |

New integration test `modify_window_dry_run_itemize` drives the real binary against an `oc-rsync --daemon` and pins the window boundary. The real (non-dry-run) transfer path is unaffected: a within-window file is still skipped and `-a` still normalises its mtime.

- `cargo fmt --all`: clean
- `cargo clippy -p transfer --all-targets --all-features --no-deps --locked -- -D warnings`: clean
- new + existing `--modify-window` and `itemize` unit/integration tests: pass

### DIFF
--- a/crates/transfer/src/receiver/quick_check.rs
+++ b/crates/transfer/src/receiver/quick_check.rs
@@ -45,8 +45,9 @@ pub(super) fn is_hardlink_follower(entry: &FileEntry) -> bool {
 ///   `f2_sec > f1_sec ? f2_sec - f1_sec <= modify_window : f1_sec - f2_sec <=
 ///   modify_window`, i.e. `|a_sec - b_sec| <= window`.
 ///
-/// upstream: util1.c:1478 same_time() (via generator.c:quick_check_ok()).
-fn same_time(a_sec: i64, b_sec: i64, window: u64) -> bool {
+/// upstream: util1.c:1478 same_time() (via generator.c:quick_check_ok() and
+/// generator.c:526 itemize() -> mtime_differs()).
+pub(super) fn same_time(a_sec: i64, b_sec: i64, window: u64) -> bool {
     if window == 0 {
         return a_sec == b_sec;
     }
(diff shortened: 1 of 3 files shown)


oferchen:
Validated on the Linux host against the actual upstream 3.5.0dev testsuite (`runtests.py --rsync-bin <oc> compare`): with this branch built, `compare` still **FAILs**. Root cause: `compare_test.py` drives `-ain --modify-window=2 src/ dst/` as a **local** transfer, so oc uses the local-copy engine, not the wire receiver/generator this branch touches. The dry-run itemize the test checks is produced by the local-copy path. Marking draft; the modify-window fix needs to land in `crates/engine/src/local_copy/` (dry-run itemize) to flip `compare`. The receiver-side change here is a separate remote-path concern and its daemon-based tests are CI-fragile.

oferchen:
Superseded by #6266 (compare fixed in the local-copy dry-run itemize engine, lxhost-verified PASS). The wire-receiver approach here targeted the wrong engine for this local test.

PR #6272, opened 2026-07-03
fix(transfer): handle receiver data-discard without panic

oferchen:
## Summary

Fixes the `recv-discard-nullderef` 3.5.0dev conformance test. In a daemon-pull
scenario (client sender to daemon receiver) where the receiver's output
`mkstemp()` fails - e.g. a read-only destination directory - the receiver must
discard the incoming delta and the run must exit 23 (`RERR_PARTIAL`), not 0.

oc-rsync already drained the delta cleanly (no null-deref / crash), but it
reported the failure through the wrong channel and dropped the exit code at two
points:

1. **Receiver side** (`crates/transfer/src/pipeline/receiver.rs`,
   `receiver/transfer/pipeline.rs`, `writer/msg_info.rs`): the failed-open
   error was emitted as `MSG_INFO`. Upstream emits it as `FERROR_XFER`
   (`receiver.c:297` - `rsyserr(FERROR_XFER, errno, "mkstemp %s failed", ...)`),
   which sets the peer's `got_xfer_error` on receipt (`log.c:311`). Warnings now
   carry their `MessageCode`, and fatal transfer errors ride `MSG_ERROR_XFER`.
   The message text now matches upstream (`[receiver] mkstemp "..." failed`)
   instead of the misattributed `[sender] send_files failed to open`.

2. **Sender/generator side** (`crates/transfer/src/generator/transfer/orchestrator.rs`,
   `reader/multiplex.rs`, `reader/server.rs`): the client sender now folds a
   received `MSG_ERROR_XFER` count into `io_error` (`IOERR_GENERAL`), mirroring
   upstream `got_xfer_error -> _exit(RERR_PARTIAL)` (`main.c:1635`).

3. **CLI exit code** (`crates/cli/src/frontend/execution/drive/summary.rs`): the
   completed-transfer branch of `execute_transfer` returned a hardcoded `0`,
   discarding `summary.io_error_exit_code()`. It now honours the summary's exit
   code (upstream `log.c:log_exit()` maps io_error to `RERR_*`).

## Transport / trigger

Daemon pull, pipelined receiver. Trigger: destination dir chmod 0555, so the
receiver's output `mkstemp()` returns EACCES and the delta (which contains a
block MATCH against the shared leading block) is discarded.

## Validation (Linux host, aarch64)

- `recv-discard-nullderef`: FAIL (exit 0) on master -> PASS after fix
  (client exits 23).
- `cargo nextest run --workspace`: see run summary below (only the known
  pre-existing xtask cross-compiler env artifacts fail identically on master).

## Test

Added `output_open_failure_drains_as_error_xfer_warning` in
`pipeline/receiver.rs`: asserts a denied output open surfaces a
`MessageCode::ErrorXfer` warning (not `MSG_INFO`), encoding why the code matters
(the peer only sets `got_xfer_error` on `FERROR_XFER`).

## lxhost run evidence

- Target test: `recv-discard-nullderef` -> `PASS` (client exits 23).
- `cargo nextest run --workspace --build-jobs 2 -j 2`:
  `29815 passed (5 slow), 1 failed, 161 skipped`. The single failure is the
  pre-existing xtask env artifact
  `xtask::commands::package::tests::cross_compiler_resolution_falls_back_to_zig`
  (no cross toolchain on the host); it is unrelated to this change and fails
  identically on master.

### DIFF
--- a/crates/cli/src/frontend/execution/drive/summary.rs
+++ b/crates/cli/src/frontend/execution/drive/summary.rs
@@ -196,7 +196,12 @@ where
                     writeln!(writer, "warning: failed to append to log file: {error}")
                 });
             }
-            0
+            // upstream: log.c:log_exit() maps the accumulated io_error /
+            // got_xfer_error into RERR_* exit codes. A transfer can complete
+            // its summary yet still owe a non-zero code (e.g. a receiver that
+            // discarded a file because its output mkstemp() failed reports exit
+            // 23 via MSG_ERROR_XFER). Honour it here instead of forcing 0.
+            summary.io_error_exit_code().unwrap_or(0)
         }
         Err(error) => {
             if let Some(observer) = live_progress
--- a/crates/transfer/src/generator/transfer/orchestrator.rs
+++ b/crates/transfer/src/generator/transfer/orchestrator.rs
@@ -287,6 +287,16 @@ impl GeneratorContext {
             .advance_to(TransferPhase::Complete)
             .map_err(crate::fsm_error)?;
 
+        // upstream: log.c:311 - each MSG_ERROR_XFER the peer sends sets
+        // got_xfer_error on receipt; main.c:1635 then _exit(RERR_PARTIAL).
+        // The receiver emits MSG_ERROR_XFER when it cannot open a file's output
+        // (e.g. mkstemp() denied by a read-only destination dir) and discards
+        // the delta. Fold that into io_error so this sender/generator reports
+        // exit 23 instead of a false success.
+        if reader.xfer_error_count() > 0 {
+            self.add_io_error(super::super::io_error_flags::IOERR_GENERAL);
+        }
+
         // upstream: handle_stats() reports stats.total_size, the sum of all
         // flist file sizes (main.c:351 write_varlong30(f, stats.total_size, 3)).
         let total_size = self.file_list.iter().map(|e| e.size()).sum();
--- a/crates/transfer/src/pipeline/receiver.rs
+++ b/crates/transfer/src/pipeline/receiver.rs
@@ -17,6 +17,8 @@ use std::io;
 use std::path::PathBuf;
 use std::thread::JoinHandle;
 
+use protocol::MessageCode;
+
 use crate::pipeline::spsc::{self, TryRecvError};
 
 use crate::delta_apply::ChecksumVerifier;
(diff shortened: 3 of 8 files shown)


oferchen:
lxhost validation (aarch64, upstream 3.5.0dev testsuite): `runtests.py recv-discard-nullderef` = **PASS** on this branch (fcbd27c7e). Full-workspace nextest validated by CI `nextest (stable)`.

oferchen:
**HOLD — real regression found.** lxhost verification (aarch64): `exclude-lsh` fails **4/4 on this branch** but passes **4/4 on master** (both freshly built with `cargo build --release -j 2`). The exit-code propagation change over-fires on exclude-lsh (filter+delete over lsh), making it exit non-zero where upstream/master exit 0. Target `recv-discard-nullderef` still PASSes. The propagation needs narrowing so exclude-lsh stays exit-0 while recv-discard-nullderef stays exit-23. Not merging until fixed.

oferchen:
## Narrowed exit-code propagation: fix `exclude-lsh` regression

### Root cause

The exit-23 over-fire on `exclude-lsh` was **not** the `MSG_ERROR_XFER` emission or the `xfer_error_count` fold. Live tracing on the Linux host of the failing leg

```
oc-rsync -av --existing --include='*/' --exclude='*' lh:.../from/ .../chk/
```

showed `xfer_error_count=0` on the generator and the exit-23 coming from the local receiver's stats: `io_error=0x1 (IOERR_GENERAL)`, driven by **10 spurious "failed to create directory ... Permission denied" metadata errors**.

The mechanism: the CLI change (honour `summary.io_error_exit_code()` instead of hardcoded 0) correctly surfaces a receiver `IOERR_GENERAL` that master silently swallowed. But that `IOERR_GENERAL` was itself bogus. In the non-incremental batch receiver (`receiver/directory/creation.rs::create_directories`, the remote-shell pull path), directories skipped under `--existing` were inserted into the same `failed_dir_paths` set as genuine `mkdir` EACCES failures. The `dir_creation_errors` builder then turned **every** entry in that set - including the benign `--existing` skips - into a "failed to create directory ... Permission denied" error, folding `IOERR_GENERAL` into the receiver stats.

Local mode never hit this (it uses a different skip path: "skipping non-existent destination file"). Upstream skips these dirs entirely under `--existing` and exits 0.

### Upstream citation

`generator.c:1374-1378` - under `--existing` (`ignore_non_existing`) a missing directory sets `skip_dir` / `FLAG_MISSING_DIR` and is skipped; `io_error` is never touched. `receiver.c` reserves `io_error` (and hence `RERR_PARTIAL`) for genuine transfer failures. The incremental receiver path (`create_directory_incremental`) already modelled this correctly via `FailedDirectories::mark_failed` + `Ok(None)`; only the batch path conflated skip with failure.

### The narrowing

Split the two conditions in `create_directories`: a new `skipped_existing_dirs` set holds `--existing` skips (suppresses itemize + metadata, like before) while `failed_dir_paths` holds only real `mkdir` EACCES failures. `dir_creation_errors` derives solely from `failed_dir_paths`, so benign skips no longer set `IOERR_GENERAL`. The genuine-failure paths (real `mkdir` EACCES on the batch path; the file-level `mkstemp` discard path that `recv-discard-nullderef` exercises) are untouched and still yield exit 23. One-file change plus a regression test.

### Verification (Linux host, fresh `cargo build --release`)

Before (branch tip `fcbd27c7e`): `exclude-lsh` FAIL 4/4, exit 23 on the `--existing` leg (master passes).

After (`b53796b60`), 3x each:

```
run1: PASS exclude-lsh   run1: PASS recv-discard-nullderef
run2: PASS exclude-lsh   run2: PASS recv-discard-nullderef
run3: PASS exclude-lsh   run3: PASS recv-discard-nullderef
```

Neighbour spot-check (exit-code propagation touches broadly): `exclude` PASS, `delete` PASS, `delete-deep` PASS.

Full nextest `--workspace --no-fail-fast`:

```
Summary [640.4s] 29900 tests run: 29896 passed, 4 failed, 161 skipped
```

The 4 failures are the known host env artifacts only (`cross_compiler_resolution_*`, `execute_reports_missing_cross_compiler`, `tarball_resolution_skips_targets_without_cross_tooling`) - missing zig/cross toolchain on the host, unrelated to this change.

PR #6297, opened 2026-07-03
fix(daemon): drain reads during delta phase to prevent transfer deadlock

oferchen:
## Summary

Fixes the confirmed P1 daemon-transport delta deadlock (#503) using **Approach C** from `docs/design/daemon-delta-deadlock-fix.md`, co-landed with the section 4.4 timeout-clear and a #504 slot-release regression test. Daemon-TCP scoped; SSH/stdio, local, the shared multiplex writer, and the goodbye handshake are untouched.

## Root cause

The daemon splits one TCP socket into two blocking `try_clone()` fds (`streams.rs:78` read / `:87` write). The receiver delta path (`transfer/pipeline.rs`) writes a batch of file requests, flushes once, then blocks reading the response - with no interleaved drain of incoming frames. Once both ~128 KB kernel socket buffers fill, neither direction makes progress: a full-duplex write-write deadlock. Over SSH/stdio the peer runs in a separate process with independent pipe buffers, so the same engine code does not wedge - the fault is specific to the single-socket daemon path with bidirectional delta flow.

A leaked 10-second accept-time socket timeout (`listener.rs:242`) fires on the wedged read and surfaces the deadlock as `code 23 / failed to fill whole buffer`; with a long module `timeout` it hangs indefinitely instead.

## Fix

**Primary (Approach C):** a new `DrainingReader` (`transfer/draining_reader.rs`) wraps the daemon read-clone fd. A background thread continuously reads the socket into an unbounded queue while the receiver loop consumes from it, so the peer's send buffer is always drained and neither direction wedges - the same anti-deadlock guarantee upstream `perform_io` provides (`io.c:882-889`). The queue is unbounded so the drain thread never blocks on a full queue and stops draining the socket (which would re-arm the wedge); in-memory growth is bounded by the receiver's outstanding-request window, mirroring upstream's `iobuf.in`. It is a transparent byte pipe - every wire byte and the multiplex framing are unchanged. The drain thread is daemon-TCP-only (`setup_transfer_streams` TCP branch); stdio/local get no drain. It is stopped and joined before the TCP goodbye drain reads the socket via another clone (`orchestration.rs`), with `Drop` as a backstop on every exit path.

**Secondary (section 4.4, co-landed):** `apply_module_timeout` (`module_spec.rs`) now clears the socket read/write timeout to `None` when the module sets no `timeout`, matching upstream's `io_timeout = 0` default (`io.c:179`). This must ship with the primary fix - alone it would turn the deadlock's abort into a hang; with the drain thread guaranteeing the phase cannot deadlock, it is safe and removes the spurious code-23 abort.

**#504 relationship:** the connection-slot RAII (`ModuleConnectionGuard`) is already correct on all exit paths. The deadlock previously kept the thread from unwinding so the guard never dropped, exhausting the module's slots (the downstream slot-leak symptom). Fixing the deadlock resolves it. A regression test pins that an aborted transfer releases its slot so the module keeps accepting connections.

## Files

- `crates/daemon/src/daemon/sections/module_access/transfer/draining_reader.rs` (new) - the drain reader/thread + tests.
- `crates/daemon/src/daemon/sections/module_access/transfer/streams.rs` - wrap the read clone on the TCP path; carry the stop handle.
- `crates/daemon/src/daemon/sections/module_access/transfer/orchestration.rs` - stop/join the drain thread before the goodbye drain.
- `crates/daemon/src/daemon/sections/module_parsing/module_spec.rs` - clear the leaked timeout when no module `timeout`.
- `crates/daemon/src/daemon/module_state/tests.rs` - #504 slot-release regression test.

## Verification (Linux validation host, aarch64)

Built `--bin oc-rsync` (release). A/B on the design-doc dbg503 recipe (128 x 4 MiB, seeded basis, backdated dst, mid-file `dd conv=notrunc` mutation, re-sync over a no-chroot daemon module):

| Case | master | this branch |
|------|--------|-------------|
| dbg503 128x4MiB delta push | **rc=23 (deadlock)** | **rc=0, trees match** |

- **Upstream parity:** same 128x4MiB delta workload, oc client -> upstream rsync 3.4.x daemon: rc=0, trees match.
- **Non-regression (all pass):** plain local `-a` copy; SSH-loopback `--no-whole-file` delta; daemon `--whole-file`; daemon quick-check skip.
- **#504 stress:** 6 forced aborted daemon transfers on a `max connections = 4` module, then a legitimate transfer is still accepted (rc=0) - the slot is not wedged.
- fmt + clippy (`-D warnings`, `--locked`) clean; new DrainingReader byte-fidelity/anti-deadlock unit tests + slot-release regression test pass.

## Residual risk

Low, daemon-TCP scoped. The drain thread lifecycle (spawn at stream setup, stop+join before goodbye, Drop backstop) is the one new concurrency seam; it is covered by unit tests and the join ordering is asserted against the goodbye drain. Wire bytes are byte-identical (transparent byte pipe), so upstream parity holds.

### DIFF
--- a/crates/daemon/src/daemon/module_state/tests.rs
+++ b/crates/daemon/src/daemon/module_state/tests.rs
@@ -2,6 +2,7 @@ use std::io;
 use std::net::{IpAddr, Ipv4Addr};
 use std::num::{NonZeroU32, NonZeroU64};
 use std::path::PathBuf;
+use std::sync::atomic::Ordering;
 
 use super::*;
 // Host pattern types are defined in the parent daemon module (via include!() of config_helpers.rs).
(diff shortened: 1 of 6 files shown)


oferchen:
### Linux validation host results (aarch64)

Built `--bin oc-rsync` (release) for both master and this branch.

**dbg503 repro (design-doc recipe: 128 x 4 MiB, seeded basis, backdated dst, mid-file `dd conv=notrunc` mutation, re-sync over a no-chroot daemon module):**

| | master | this branch |
|---|---|---|
| dbg503 128x4MiB delta push | rc=23 (deadlock, `failed to fill whole buffer`) | **rc=0, trees match** |

**Upstream parity:** identical 128x4MiB delta workload, oc client -> upstream rsync 3.4.x daemon -> rc=0, trees match.

**Non-regression (all pass):** plain local `-a`; SSH-loopback `--no-whole-file` delta; daemon `--whole-file`; daemon quick-check skip. Daemon gzip delta upload rc=0, trees match.

**#504 stress:** 6 forced aborted daemon transfers on a `max connections = 4` module, then a legitimate transfer is still accepted (rc=0) - the slot is not wedged.

**Tests (lxhost):** targeted DrainingReader byte-fidelity / anti-deadlock + `apply_module_timeout` + #504 slot-release: 5 passed. Full daemon crate lib: 1665 passed, 0 failed. fmt + clippy (`-D warnings`, `--locked`) clean. The full-workspace nextest is the authoritative CI gate.

PR #6305, opened 2026-07-03
feat(transfer): basis read-ahead prefetch (fadvise WILLNEED, default-off)

oferchen:
## Summary

The decoupled receiver pipeline is I/O-wait-bound: the main thread blocks **serially** on basis-file disk reads (`MapFile::open` -> `map_ptr` -> `folio_wait_bit_common`) during delta-token processing. Measurements show the disk at only ~66% util while ~33-40% of receiver wall-clock is spent waiting on those basis reads.

This PR issues `posix_fadvise(POSIX_FADV_WILLNEED)` on upcoming files' **resolved** basis paths ahead of time, so the kernel warms the page cache before the pipeline reaches each basis. This is the kernel-hint strategy: zero userspace memory, byte-identical output.

## Behavior

- **Default-OFF and feature-gated** behind `basis-readahead`. The default build compiles in only `NullPrefetcher` (empty impl), spawns no threads, and is byte-for-byte unchanged.
- **Runtime-activated** for bench: even when the feature is compiled in, the active prefetcher is only selected when `OC_RSYNC_BASIS_READAHEAD=1`. No CLI surface.
- **Correctness disable-list:** prefetch is forced to `NullPrefetcher` under `--inplace` and `--append` (writing one file mutates a later file's basis) and on the redo pass (redo requests carry no basis).
- **Byte-identical wire even when ON:** `fadvise(WILLNEED)` is a pure page-cache timing hint on local basis reads. It never changes the bytes read, so wire output and the reconstructed destination are identical. Stated in the module doc.

## Design

- `fast_io::willneed::hint_basis_willneed` - safe wrapper over `posix_fadvise` (Linux/Android; no-op on BSD/macOS/Windows, matching platform `libc` availability). The single `unsafe` block lives in the permitted `fast_io` crate; `transfer` stays `#![deny(unsafe_code)]`.
- `transfer::basis_prefetch` - `BasisPrefetcher` trait, `NullPrefetcher` default, and `FadviseWillneedPrefetcher` (Unix + feature) draining a bounded `std::sync::mpsc::sync_channel` on a background worker (depth 4 = backpressure; `Drop` closes + joins). Uses std rather than adding a dependency.
- Wired into `run_pipeline_loop_decoupled`: after signatures are computed (`!is_redo_pass`), the resolved `BasisFileResult.basis_path` for the look-ahead window is handed to the prefetcher. The send/response/token/disk-commit paths are untouched.

## Verification

- `cargo clippy -p fast_io` / `-p transfer` (default) / `-p transfer --features basis-readahead` - all clean with `-D warnings`.
- `cargo nextest -p fast_io -E 'test(willneed)|test(hint)'` - 4 passed.
- `cargo nextest -p transfer [--features basis-readahead] -E 'test(basis_prefetch)|test(prefetch)'` - 6 passed (feature on), 4 passed (default, incl. feature-off Null selector).
- `cargo check -p bin --features basis-readahead` - workspace passthrough resolves.

## Follow-up

An lxhost A/B bench and the default-on decision are deliberately left as follow-up. This PR does **not** flip the default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,13 @@ copy_file_range = ["fast_io/copy_file_range"]
 # `docs/design/mmap-vs-sqpoll-conflict-resolution.md` for the rationale.
 mmap-free-basis = ["engine/mmap-free-basis", "fast_io/mmap-free-basis"]
 
+# EXPERIMENTAL (default-off): basis read-ahead prefetch for the decoupled
+# receiver. Forwards to `transfer/basis-readahead`, which warms upcoming delta
+# basis files into the page cache via posix_fadvise(WILLNEED). Kept out of
+# default features; a bench build enables it and activates it at runtime with
+# OC_RSYNC_BASIS_READAHEAD=1. Pure timing hint - wire output is byte-identical.
+basis-readahead = ["transfer/basis-readahead"]
+
 # ============================================================================
 # OpenSSL Features
 # ============================================================================
--- a/crates/fast_io/src/lib.rs
+++ b/crates/fast_io/src/lib.rs
@@ -194,6 +194,9 @@ pub mod win_tmpfile;
 /// macOS-optimized file writer using `F_NOCACHE` and `writev`.
 pub mod macos_io;
 
+/// Kernel read-ahead hint (`POSIX_FADV_WILLNEED`) for basis-file prefetch.
+pub mod willneed;
+
 /// Memory-mapped file reader for efficient large file access.
 #[cfg(unix)]
 pub mod mmap_reader;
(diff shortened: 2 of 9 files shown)


oferchen:
Closing: the lxhost bench (SSH-loopback PULL, 400x8MiB >RAM cold delta, cold-cache A/B) showed NO measurable win. The fadvise(WILLNEED) hint fires correctly and output is byte-identical, but disk %util sat at ~22% (not the ~66% the profile suggested) — the transfer is bound by the single-stream SSH pipe + delta/checksum CPU, not basis-read latency, so the kernel hint has nothing to accelerate (wall-clock delta -1.5% = noise). The per-file basis-read stall the profile measured is not on the end-to-end critical path. Would only help on a rotational/network-backed/bandwidth-contended basis device; no such host to validate. Shelving rather than shipping an unmerited default-off feature.

PR #6357, opened 2026-07-08
fix(protocol): drop "none" from client-side negotiation lists

oferchen:
## Summary

Upstream `get_default_nno_list` (compat.c:485) omits the `none` entry (num == 0) from the **client's** default checksum and compression lists, advertising it only on the server side:

```c
if (nni->num == 0 && !am_server && !dup_markup)
    continue;
```

oc-rsync serialized `none` unconditionally on both sides, so an oc-rsync **client** sent a trailing ` none` token that upstream clients never emit:

| Side | Upstream client | oc-rsync client (before) |
|------|-----------------|--------------------------|
| checksum | `xxh128 xxh3 xxh64 md5 md4 sha1` | `xxh128 xxh3 xxh64 md5 md4 sha1 none` |
| compress | `zstd ... zlibx zlib` | `zstd ... zlibx zlib none` |

The stray token is harmless to the selection algorithm (`none` is always last, so a stronger algorithm always matches first) but is a byte-level divergence from upstream's negotiation frames, visible in wire captures of an oc-rsync client talking to an upstream server.

## Change

Filter `none` from the advertised lists when `is_server` is false, mirroring upstream exactly. `none` is the sole `num == 0` name in both the checksum and compression tables, so filtering by name is exact. Local selection still uses the full `SUPPORTED_CHECKSUMS` / `supported_compressions()` list, so negotiated results are unchanged. The server side continues to advertise `none`.

## Testing

- New regression test `test_client_drops_none_server_keeps_it` asserts the exact bytes written for both client (`none` absent) and server (`none` present) sides.
- `cargo fmt`, `cargo check -p protocol --all-features --tests`, and `cargo clippy -p protocol --all-features --all-targets -- -D warnings` all pass.

### DIFF
--- a/crates/protocol/src/negotiation/capabilities/negotiate.rs
+++ b/crates/protocol/src/negotiation/capabilities/negotiate.rs
@@ -268,15 +268,30 @@ pub fn negotiate_capabilities_with_override(
     // When --checksum-choice overrides the selection, advertise only that
     // algorithm (upstream options.c replaces valid_checksums with the user's
     // choice so negotiate_the_strings sees a single-entry list).
+    //
+    // upstream: compat.c:485 get_default_nno_list drops the "none" entry
+    // (num == 0) from the *client* default list; only the server advertises it
+    // (`if (nni->num == 0 && !am_server && !dup_markup) continue;`). "none" is
+    // the sole num == 0 name in both the checksum and compression tables, so
+    // filtering by name mirrors upstream exactly.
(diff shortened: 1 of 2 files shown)


oferchen:
Duplicate of #6360, which implements the same client-side `none`-drop in negotiate_capabilities with stronger tests (pins the exact expected wire bytes as independent ground truth rather than recomputing them with the implementation's own filter) and fuller upstream citations. Closing in favor of #6360.

PR #6445, opened 2026-07-10
fix(protocol): use legacy longint for proto<30 end-of-transfer stats

oferchen:
## Problem (#204, #205)

Every proto-29 transfer that reached the end-of-transfer stats exchange
deadlocked. Symptoms tracked as #204 (`-z` zlib) and #205 (1 MiB delta),
but the root cause is shared and affects **all** proto-29 file transfers,
not just those two options.

## Root cause

The sender's end-of-transfer sequence is
`send_files -> handle_stats -> read_final_goodbye`. At protocol >= 29,
`handle_stats` exchanges **five** longint values (total_read,
total_written, total_size, flist_buildtime, flist_xfertime). Upstream
encodes each with `write_varlong30`, which is protocol-gated (io.h):

```c
static inline void write_varlong30(int f, int64 x, uchar min_bytes) {
    if (protocol_version < 30) write_longint(f, x);      // fixed 4 bytes
    else                       write_varlong(f, x, min_bytes);
}
```

`TransferStats::write_to`/`read_from` called our `write_varlong30`
unconditionally, and that helper never restored the `protocol_version < 30`
branch. So at proto 29 the five stats were emitted as compact varlong
(15 bytes) instead of fixed 4-byte longint (20 bytes). A 2.6.9 peer reads
them with `read_longint`, consumed the first three plus 3 stray bytes,
then blocked mid-read of `flist_buildtime`. Its final `-1` goodbye was
never sent, so the sender's `read_final_goodbye` blocked too - a
deadlock, seen as "connection unexpectedly closed" under the CI timeout.

Captured on the wire (oc daemon-sender, 100-byte file): proto-29 emitted a
15-byte stats frame then blocked on `recvfrom`; proto-30 emitted its
varlong stats and completed. Confirmed against rsync 2.6.9.

## Fix

`TransferStats::write_to`/`read_from` now branch on
`ProtocolVersion::uses_varint_encoding()` (>= 30) and use `write_longint`/
`read_longint` for proto < 30, exactly mirroring upstream `write_varlong30`
/`read_varlong30`. Protocol >= 30 uses `write_varlong(_, 3)` as before, so
the proto 30/31/32 wire is byte-for-byte unchanged.

## Verification (rsync 2.6.9, aarch64)

- proto-29 `-avz` pull and 1 MiB delta pull: rc=0, trees byte-identical
  (previously hung).
- proto 30/31/32 `-z` and delta pull (oc<->oc): no regression, rc=0,
  byte-identical.
- Full `rp28_e_1_run.sh` (oc daemon, 2.6.9 client): **15/15 pass**.
- Full `rp28_f_1_run.sh` (oc client, 2.6.9 daemon): **18/18 pass**.
- `cargo nextest -p protocol -E 'test(transfer_stats)'`: 21/21 pass,
  including new golden-byte tests asserting the proto-29 4-byte layout.
- fmt + clippy (`-p protocol`) clean.

Both interop cells that documented this exact stall (RP28.e.2, RP28.f.2)
are promoted from `continue-on-error` to blocking.

### DIFF
--- a/.github/workflows/_interop.yml
+++ b/.github/workflows/_interop.yml
@@ -295,10 +295,10 @@ jobs:
       # build it inline via the RP28.b.1 helper to keep the cell
       # self-sufficient. The harness exits 77 when either binary is
       # missing so the cell becomes a no-op rather than a false failure.
-      # Non-blocking via `continue-on-error: true` until RP28.k promotes
-      # pre-30 parity to a required check.
+      # Blocking: the pre-30 daemon-sender end-of-transfer stats deadlock is
+      # fixed (proto < 30 now emits fixed 4-byte longint stats, matching
+      # `write_varlong30`), so the full D1-D10 matrix passes against 2.6.9.
       - name: Run rsync 2.6.9 daemon-mode client interop (RP28.e.2)
-        continue-on-error: true
         run: |
           set -euo pipefail
           RSYNC_269="target/interop/upstream-install/2.6.9/bin/rsync"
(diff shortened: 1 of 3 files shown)


oferchen:
Superseded: the proto-29 stats-encoding fix (transfer.rs proto<30 longint gate) landed in #6438 with the full v28/v29 golden updates. The unique guard-relaxation (promoting RP28.e.2/RP28.f.2 to blocking) is extracted onto ci/promote-rp28-blocking, rebased on master so it carries no redundant/failing golden delta. Closing to avoid a duplicate transfer.rs change.

PR #6545, opened 2026-07-12
fix(matching): reframe parallel delta literals to match sequential

oferchen:
## Summary

The opt-in parallel delta scan (`DeltaGenerator::generate_chunked`) split the source into disjoint ranges scanned independently. Each range restarted the `block_len + CHUNK_SIZE` literal-flush cadence at its own start, so a literal run longer than one cadence that crossed a **range** boundary (not a match boundary) was re-segmented into different length-prefix chunks than the single-pass sequential scan - same data, same total/literal counts, different wire bytes. Interior boundaries also landed mid-block, degrading a straddling matched block to literals.

This makes the parallel path **byte-identical** to the sequential scan for the eligible (duplicate-free, block-aligned-match) case:

1. **Block-aligned ranges.** Every interior boundary is floored to a basis block edge, so no matched block straddles a boundary; on a duplicate-free basis the `Copy` set and order then equal the sequential scan.
2. **Literal re-segmentation.** `resegment_literals` coalesces every maximal run of adjacent `Literal` tokens and re-splits it at the shared cadence from the run start.

### DRY by construction
The literal-flush cadence is now a single source of truth: `literal_flush_cadence(block_len)` is called by **both** the sequential threshold flush and the parallel post-pass, so the two paths cannot drift. The sequential scan and the wire encoder are unchanged.

## Decision: kept opt-in (default NOT flipped)

The mandatory in-process seal (below) proves byte-identity for every eligible case, but a residual remains: **shifted** duplicate-free content (e.g. bytes inserted mid-file) can still leave a match straddling an aligned boundary, producing one extra literal block per straddled boundary. Reconstruction stays byte-exact, but the wire bytes differ from the sequential/upstream serial output. Per the wire-fidelity rule (any residual divergence -> do not flip), `--parallel-delta-scan` **remains opt-in**. A follow-up seam-match-repair pass at range boundaries would be required before flipping the default.

## Seal table (in-process, exact production serializer `generate`/`generate_chunked` -> `script_to_wire_bytes`)

| Case | Chunks | Result |
|------|--------|--------|
| Dup-free, boundary-aligned single-byte edits | 2/4/8 | wire-identical to sequential |
| Dup-free, 200 KiB literal run straddling `n/2` | 2/4/8 | wire-identical to sequential |
| Dup-free, 1 MiB literal run straddling `n/2` | 2/4/8 | wire-identical to sequential |
| Exact-cadence-multiple literal run (unit) | - | re-split identical, no empty tail |
| Copy-at-boundary / adjacent copies / sub-cadence (unit) | - | preserved / unchanged |
| Duplicate-heavy basis | 4 | **diverges** (gate reverts to sequential - unchanged) |
| Shifted dup-free content (13 B inserted mid-file) | 4 | **diverges** (residual; keeps default off) |

All cases also reconstruct byte-exact via `apply_delta`. `matching` lib (224 tests) + `parallel_delta_wire_parity` (4 tests) pass; `cargo fmt` and `clippy -p matching --all-features --tests -D warnings` clean.

## #6543-equivalent test flipped to equality
The referenced `parallel_delta_dup_free_long_literal_seam_diverges` test was not present at this HEAD; the long-literal-seam case is now covered by `parallel_delta_dup_free_long_literal_is_wire_identical`, which asserts `chunked_wire == seq_wire` (plus reconstruction) at chunks 2/4/8 for 200 KiB and 1 MiB straddling runs.

## Not done (deferred with the flip)
End-to-end real-binary seal (80 MiB push vs upstream 3.4.4 receiver at 2/4/8 threads) and the default flip / `--no-parallel-delta-scan` opt-out are deferred: they gated the flip, which is not happening due to the shifted-content residual. The in-process seal exercises the exact production token serializer.

### DIFF
--- a/crates/matching/src/generator.rs
+++ b/crates/matching/src/generator.rs
@@ -34,6 +34,22 @@ const DEFAULT_BUFFER_LEN: usize = 128 * 1024;
 /// upstream: rsync.h:158, match.c:339
 const CHUNK_SIZE: usize = 32 * 1024;
 
+/// The single source of truth for the literal-flush cadence.
+///
+/// The sequential scan ([`DeltaGenerator::generate`]) flushes its pending
+/// literal accumulator into one `Literal` token every time it reaches this many
+/// bytes (`pending_literals.len() >= literal_flush_cadence(block_len)`), so a
+/// match-free region is framed as a run of fixed-size literal tokens (the final
+/// partial run aside). The parallel post-pass ([`resegment_literals`]) re-splits
+/// its coalesced literal runs at exactly this same cadence, so both paths frame
+/// literals identically by construction rather than by coincidence.
+///
+/// upstream: match.c:339-340 (`if (a_bytes >= CHUNK_SIZE + blength)`).
+#[inline]
+const fn literal_flush_cadence(block_len: usize) -> usize {
+    block_len + CHUNK_SIZE
+}
+
 /// Minimum bytes per parallel range in [`DeltaGenerator::generate_chunked`].
 ///
 /// Ranges below this size are not worth a rayon task: the per-range boundary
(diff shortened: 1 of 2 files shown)


oferchen:
Superseded by #6546, which subsumes this PR's literal-cadence reframing and additionally makes shifted duplicate-free content wire-identical (Approach A: overlap stripes + greedy Copy-run merge), closing the residual this PR documented as still-divergent.

PR #6602, opened 2026-07-14
fix: honor daemon dont-compress '*' whole-stream store

oferchen:
## Summary

Honor upstream rsync's whole-stream compression "match all" special case: when a
daemon module is configured with `dont compress = *`, the sender must initialise
the zlib stream at level 0 (store) for the entire transfer instead of compressing
per block.

## Upstream mechanism

`token.c:init_set_compression()` builds the sender's dont-compress match list from
`lp_dont_compress(module_id)`. When a bare `*` token appears in that list, upstream
collapses it to a whole-stream store (`token.c:206-211`):

```c
if (t - start == 1+1 && *start == '*') {
    *match_list = '\0';
    suftree = NULL;
    per_file_default_level = skip_compression_level;   /* Z_NO_COMPRESSION for zlib */
    break;
}
```

`per_file_default_level` is fed into `deflateInit2()` (`token.c:378`), so the whole
zlib stream is stored. This is zlib-only: zstd uses `do_compression_level`
unconditionally (`token.c:748`) and is unaffected. The per-file suffix matching
(`set_compression()`) is compiled out (`#if 0`) in 3.4.4, so no per-file level
change happens. A non-daemon client's `--skip-compress` sets `f = ""`
(`token.c:182`) and never triggers the store.

Verified against rsync 3.4.4 (zlib, 2 MB compressible file, daemon sender):

| Scenario | Bytes on wire |
|---|---|
| `dont compress = *` (daemon) | ~2,000,918 (stored) |
| normal module `-z` | ~2,036 (compressed) |
| client `--skip-compress='*'` | ~2,036 (compressed, discarded) |
| zstd + `dont compress = *` | ~190 (compressed, ignored) |

## Divergence

oc parsed a bare `*` as a literal suffix pattern that matches nothing, so
`dont compress = *` compressed the whole stream at level 6 instead of storing it.

## Fix

- Detect the bare-`*` match-all case when building the daemon server config and
  set a new `ConnectionConfig::dont_compress_match_all` flag (instead of an inert
  literal-`*` skip list). Scoped to the daemon path, so client `--skip-compress`
  is untouched, matching upstream's `f = ""`.
- In the generator transfer loop, when the flag is set and the negotiated codec is
  zlib/zlibx, select `CompressionLevel::None` (level 0 store) for the whole-stream
  encoder. zstd/lz4 keep the configured level, mirroring upstream.

No changes to the compression-level representation. The default skip-compress path
and all other codecs remain wire-identical.

## Verification

- New unit tests: `whole_stream_compression_level` (zlib match-all -> None; zstd/lz4
  and non-match-all pass through) and `dont_compress_is_match_all` detection.
- End-to-end against rsync 3.4.4 as client pulling from an oc-rsync daemon: zlib
  `dont compress = *` -> 2,000,696 bytes (stored, matches upstream); normal module
  and zstd unchanged; transferred content byte-identical.
- `cargo fmt`, `cargo clippy -p transfer -p daemon --all-features --no-deps -D warnings`
  clean.

### DIFF
--- a/crates/daemon/src/daemon/sections/module_access/client_args/server_config.rs
+++ b/crates/daemon/src/daemon/sections/module_access/client_args/server_config.rs
@@ -244,7 +244,12 @@ fn build_server_config(
             // upstream: loadparm.c - `dont compress` parameter specifies suffixes
             // that should skip per-file compression during transfer.
             if let Some(dont_compress) = module.dont_compress.as_deref() {
-                if let Some(list) = parse_daemon_dont_compress(dont_compress) {
+                if dont_compress_is_match_all(dont_compress) {
+                    // upstream: token.c:206-211 - a bare `*` collapses the match
+                    // list to a whole-stream store (zlib level 0) and clears the
+                    // per-file suffix tree, so no `skip_compress` list is built.
+                    cfg.connection.dont_compress_match_all = true;
+                } else if let Some(list) = parse_daemon_dont_compress(dont_compress) {
                     cfg.skip_compress = Some(list);
                 }
             }
--- a/crates/daemon/src/daemon/sections/module_access/helpers.rs
+++ b/crates/daemon/src/daemon/sections/module_access/helpers.rs
@@ -174,6 +174,22 @@ pub(crate) fn format_bandwidth_rate(value: NonZeroU64) -> String {
     }
 }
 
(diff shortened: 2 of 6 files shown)


Katomoto (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

PR #6621, opened 2026-07-15
feat(transfer): wire MSG_DELETED for server-side --delete output

oferchen:
## Summary

Wires the `MSG_DELETED` (multiplex tag 101) message end-to-end so a server
generator performing `--delete` forwards each deletion to the client over the
wire, exactly as upstream rsync does. Previously a server-side `--delete` pass
rendered its `deleting <path>` lines into a thread-local buffer that is never
drained on the remote side, so a `-v --delete` push produced no client-visible
deletion output, and the `-i` push routed the row through `MSG_INFO` rather than
the dedicated `MSG_DELETED` tag. A client receiving `MSG_DELETED` from an
upstream server silently dropped the frame.

## Upstream behaviour mirrored

- `rsync.h:300` - `MSG_DELETED = 101`.
- `log.c:866-869` `log_delete()` - `am_server && protocol_version >= 29 &&
  len < MAXPATHLEN` sends `send_msg(MSG_DELETED, fname, len, ...)`; a directory
  bumps `len` to include its trailing NUL (`log.c:867-868`). Otherwise the line
  is rendered locally (`log.c:872-874`, `"deleting %n"` or the itemize format).
- `io.c:1614-1621` - the reader treats a trailing NUL as a directory marker,
  then calls `log_delete()`, which on the non-server side formats and gates the
  line on `INFO_GTE(DEL, 1)` / `stdout_format`.
- `options.c:251` - verbose level 1 sets `DEL=1`, so `-v` enables the plain
  form.

## Changes

- `protocol`: `MessageCode::Deleted` (101) already existed and the multiplex
  frame layer is generic over the tag, so no production change was needed. Added
  a golden byte test pinning the exact wire bytes (header tag `0x6C` plus
  payload, with the directory trailing-NUL convention) and the send/recv round
  trip.
- `transfer` writer: added `MsgInfoSender::send_msg_deleted`, emitting a
  `MSG_DELETED` frame on a multiplexed server stream (no-op otherwise).
- `transfer` receiver delete pass: a single `emit_delete_notification` helper
  routes each deletion the way `log_delete()` does - a server generator at
  protocol >= 29 forwards the raw name (directory + trailing NUL) and prints
  nothing locally; every other side renders `deleting %n` directly as before.
  Applied to all three emit sites (immediate parallel pass, `--delete-delay`
  executor, and the `--max-delete` capped serial path).
- `transfer` reader: the client-sender demux now renders received `MSG_DELETED`
  frames, formatting `deleting <path>` / `*deleting   <path>` (directories get a
  trailing slash) and gating on the client's own info=del / itemize verbosity.
  The render state is captured on the setup thread and plumbed through
  `ServerReader`, so a server or client-receiver reader leaves it unset and
  drops the frame, matching upstream's non-rendering `am_server` path.

## Tests

- Golden byte test for the `MSG_DELETED` frame (file payload, directory
  trailing-NUL payload, header tag, send/recv round trip).
- Client render unit tests: plain and itemize forms for files and directories,
  itemize-wins-over-plain precedence, suppression when neither gate is active,
  and drop-without-render-state.
- Server-vs-local routing unit tests: a server generator forwards `MSG_DELETED`
  and prints nothing locally; a local/client receiver renders directly with no
  frame; protocol < 29 falls back to the local render path.

All new tests pass; the existing deletion, delete-timing, and filter-chain
suites remain green. `cargo fmt`, workspace `clippy -D warnings`, protocol and
transfer targeted tests all clean.

### DIFF
--- a/crates/protocol/tests/golden_msg_deleted_wire.rs
+++ b/crates/protocol/tests/golden_msg_deleted_wire.rs
@@ -1,0 +1,72 @@
+//! Golden byte tests for the `MSG_DELETED` (tag 101) multiplex frame.
+//!
+//! A server generator forwards each `--delete` victim to the client as a
+//! `MSG_DELETED` frame carrying the raw deletion-root-relative name; a directory
+//! carries a trailing NUL so the reader can tell it from a regular file. These
+//! tests pin the exact wire bytes (4-byte LE header + payload) so the encoding
+//! stays byte-compatible with upstream rsync.
+//!
+//! # Upstream Reference
+//!
+//! - `rsync.h:300` - `MSG_DELETED = 101`.
+//! - `io.c:mplex_write()` - header = `(MPLEX_BASE + code) << 24 | len`, so the
+//!   tag byte is `7 + 101 = 108 = 0x6C`.
+//! - `log.c:866-869` - `send_msg(MSG_DELETED, fname, len, ...)`; a directory
+//!   bumps `len` to include its trailing NUL (`log.c:867-868`).
+//! - `io.c:1616` - the reader treats a trailing NUL as a directory marker.
+
+use std::io::Cursor;
+
+use protocol::{MessageCode, MessageFrame, MessageHeader, recv_msg_into, send_msg};
+
+#[test]
+fn golden_msg_deleted_header_tag_is_108() {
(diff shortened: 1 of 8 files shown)


oferchen:
Superseding with a clean single-commit PR; reopening fresh.

PR #6641, opened 2026-07-16
fix(protocol): honor negotiated symlink-target iconv gating and strict-failure semantics

oferchen:
## Summary

The `--iconv` filename/symlink-target path diverged from upstream rsync 3.4.4
(protocol 32) in four ways. Symlink-target transcoding was not gated on the
negotiated `CF_SYMLINK_ICONV` capability, and three strict-failure paths kept
lossy `?`/U+FFFD-mangled data instead of upstream's drop-or-empty + `io_error`
semantics. Against a peer that lacks the capability (e.g. proto-30 / rsync
3.0.x) this silently transcoded symlink targets that upstream sends raw, and on
unconvertible input it produced different bytes and a different exit code than
upstream. All four are fixed to match the upstream C byte-for-byte.

## Divergences fixed

1. **Symlink-target iconv must honor negotiated `CF_SYMLINK_ICONV`.**
   Upstream computes `sender_symlink_iconv = iconv_opt && (am_server ?
   strchr(client_info,'s') : (compat_flags & CF_SYMLINK_ICONV))`
   (`compat.c:765-767`) and gates the symlink-target conversion separately from
   the filename conversion on both send (`flist.c:1642`) and receive
   (`flist.c:1156`). oc attached the converter to the symlink-target
   writer/reader whenever `--iconv` was set, with no capability check. A single
   threaded `symlink_iconv` boolean (derived from the negotiated compat flags)
   now gates the encode/decode; without the capability, targets pass through as
   raw local bytes (`flist.c:1181` `read_sbuf` else branch / `flist.c:1659`).

2. **Sender drops a symlink entry whose target is unconvertible under strict
   iconv.** Upstream `send_file1` sets `io_error |= IOERR_GENERAL` and
   `return NULL` on the symlink target's `ic_send` failure
   (`flist.c:1642-1651`). oc used a lossy `?`-mangled target and kept the entry.
   The build-time drop pass now also strict-tests the symlink target (only when
   `symlink_iconv` gating says the target is being converted) and records
   `IOERR_GENERAL` (exit 23), keeping sender/receiver ndx aligned.

3. **Receiver empties an unconvertible received symlink target + sets
   `io_error`.** Upstream `recv_file_entry` warns, sets `io_error |=
   IOERR_GENERAL`, and empties the target (`bp[0]='\0'`, `outbuf.len = 0`) on
   `ic_recv` failure (`flist.c:1169-1177`). oc kept a `?`/U+FFFD target and only
   emitted a trace. It now empties the target and records `io_error`.

4. **Receiver empties an unconvertible received filename + sets `io_error`.**
   Upstream prints `FERROR_UTF8`, sets `io_error |= IOERR_GENERAL`, and sets
   `outbuf.len = 0` so the name becomes empty (`flist.c:757-764`). oc kept a
   lossy `?`/U+FFFD non-empty name. It now empties the name to match upstream;
   the entry stays in the list so ndx alignment is preserved.

## Upstream references verified (rsync-3.4.4)

- `compat.c:765-767` - `sender_symlink_iconv` negotiation gate.
- `flist.c:1642-1651` - `send_file1()` strict symlink-target failure -> drop.
- `flist.c:1156,1169-1178` - `recv_file_entry()` symlink-target gate + empty.
- `flist.c:757-764` - `recv_file_entry()` filename strict failure -> empty.

## Tests added

Protocol (`crates/protocol`):
- `read_symlink_target_without_negotiated_flag_preserves_wire_bytes` - no
  `CF_SYMLINK_ICONV` -> received target stays raw wire bytes, no `io_error`.
- `write_symlink_target_without_negotiated_flag_passes_raw_local_bytes` - no
  `CF_SYMLINK_ICONV` -> sent target stays raw local bytes.
- `read_symlink_target_strict_failure_empties_target_and_sets_io_error`.
- `read_entry_strict_filename_failure_empties_name_and_sets_io_error`.
- `symlink_message_shape_matches_upstream` - diagnostic wording.
- Existing transcode tests updated to require the negotiated flag
  (`with_symlink_iconv(true)`), pinning the corrected gating.

Transfer (`crates/transfer`), sender drop-decision unit tests:
- `symlink_with_unconvertible_target_dropped_when_negotiated`.
- `symlink_with_unconvertible_target_kept_when_not_negotiated`.
- `unconvertible_name_always_dropped`.
- `convertible_symlink_survives_with_gate_on`.

`cargo fmt --all --check` and `cargo clippy --all-targets --all-features
--no-deps -D warnings` are clean for the touched crates.

### DIFF
--- a/crates/protocol/src/flist/read/extras.rs
+++ b/crates/protocol/src/flist/read/extras.rs
@@ -22,10 +22,15 @@ impl FileListReader {
     /// Wire format: varint30(len) + raw bytes
     ///
     /// upstream: flist.c:recv_file_entry() lines 920-935
+    ///
+    /// `name` is the entry's already-read wire filename, used only to render the
+    /// `cannot convert symlink data for` diagnostic on a strict conversion
+    /// failure (mirroring upstream's `full_fname(thisname)`).
     pub(super) fn read_symlink_target<R: Read + ?Sized>(
-        &self,
+        &mut self,
         reader: &mut R,
         mode: u32,
+        name: &[u8],
     ) -> io::Result<Option<PathBuf>> {
         let is_symlink = mode & 0o170000 == 0o120000;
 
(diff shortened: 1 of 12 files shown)


viktoriavikulya (commented):
The happy path is clean. What happens if the third-party API returns 429 here? Might be worth handling that case explicitly.

PR #6777, opened 2026-07-18
docs(cli): tidy frontend execution and command_builder docs

oferchen:
Comment and rustdoc-only cleanup across the CLI frontend `execution/`,
`command_builder/`, and root-level modules. No logic, signature, or
control-flow changes.

- Add missing module (`//!`) docs where a true module file lacked one.
- Fix stale doc claims verified against the code: `numeric.rs` no longer
  lists a nonexistent `--human-readable` parser; `size.rs` drops the
  unsupported exa (`E`) suffix; `preflight.rs` corrects the ACL/xattr
  check-site line numbers; `password.rs` no longer points at a `lib.rs`
  parser that does not exist.
- Remove a restatement comment and tidy trailing punctuation.

All `upstream:` citations preserved verbatim (before/after counts identical).

### DIFF
--- a/crates/cli/src/frontend/command_builder/mod.rs
+++ b/crates/cli/src/frontend/command_builder/mod.rs
@@ -1,3 +1,5 @@
+//! Assembles the clap Command from staged option-group sections.
+
 mod sections;
 
 pub(super) use clap::{Arg, ArgAction, Command as ClapCommand, builder::OsStringValueParser};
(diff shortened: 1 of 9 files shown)


oferchen:
Superseded by #6775 (already merged) and the targeted fix in the PR above. Merging this as-is would regress #6775 (it removes backtick spans and reverts a cleaned-up comment). Its two genuine net-new fixes (size.rs exa-suffix doc + the dead --compress-level .help) are salvaged in the fix PR instead.

PR #6944, opened 2026-07-25
fix(core): stop writing the remote dest under --only-write-batch

oferchen:
## Summary
- --only-write-batch on a remote push updated the remote destination; upstream writes nothing (verified: upstream 0 files, oc 2)
- Emit the --only-write-batch=X placeholder in the sender's server args so the remote receiver runs dry (options.c:2850, main.c:1839)
- Divert the token stream into the batch instead of teeing it to the wire (sender.c:217 f_xfer = write_batch < 0 ? batch_fd : f_out), which also fixes the exit-12 stall once the remote stops reading tokens
- --write-batch (non-only) still transfers and records unchanged

## Test plan
- [ ] Remote push with --only-write-batch leaves the destination empty and exits 0
- [ ] A valid batch file is still produced
- [ ] --write-batch still transfers AND records (no regression to the shared path)
- [ ] Server args include --only-write-batch=X for the sender
- [ ] cargo fmt clean, clippy -D warnings clean, cargo check --workspace clean

### DIFF
--- a/crates/core/src/client/config/client/arguments.rs
+++ b/crates/core/src/client/config/client/arguments.rs
@@ -66,6 +66,20 @@ impl ClientConfig {
         self.batch_config.as_ref()
     }
 
+    /// Reports whether `--only-write-batch` is in effect (upstream's
+    /// `write_batch < 0`).
+    ///
+    /// The batch is still recorded in full, but no destination is updated:
+    /// `main.c:1839` turns the flag into `dry_run = 1` and `sender.c:217`
+    /// points the token stream at the batch file instead of the wire.
+    #[doc(alias = "--only-write-batch")]
+    #[must_use]
+    pub fn only_write_batch(&self) -> bool {
+        self.batch_config
+            .as_ref()
+            .is_some_and(|cfg| !cfg.should_transfer())
+    }
+
     /// Returns `true` when the local client should emit the sender-side `<`
     /// itemize direction arrow (push over a remote shell or daemon).
     ///
--- a/crates/core/src/client/remote/embedded_ssh_transfer.rs
+++ b/crates/core/src/client/remote/embedded_ssh_transfer.rs
@@ -700,6 +700,7 @@ fn build_server_config_for_generator(
         server_config.file_selection.files_from_path = Some(path);
         server_config.file_selection.from0 = plan.sender_from0;
     }
+    flags::apply_only_write_batch_for_sender(config, &mut server_config);
 
     flags::apply_common_server_flags(config, &mut server_config);
     Ok(server_config)
--- a/crates/core/src/client/remote/flags.rs
+++ b/crates/core/src/client/remote/flags.rs
@@ -396,6 +396,31 @@ fn split_pattern_modifiers(raw: &str) -> (String, bool, bool) {
     (raw[start..end].to_owned(), anchored, directory_only)
 }
 
+/// Carries `--only-write-batch` onto a remote-shell PUSH sender's config.
+///
+/// Upstream binds `f_xfer` once per `send_files()` run from the global
+/// `write_batch` (`sender.c:217`); oc's sender reads the same decision off its
+/// in-process `ServerConfig`, which is parsed from the compact flag string and
+/// so never sees this long-form-only option. Without it the sender would keep
+/// streaming tokens at a remote receiver that `--only-write-batch=X` has just
+/// put into `dry_run` and which therefore reads none of them.
+///
+/// Scoped to the remote-shell sender on purpose: `server_options()` emits the
+/// placeholder inside its `am_sender` block, so a pull sender-server never sees
+/// it, and the daemon path deliberately strips client-only batch flags from the
+/// module argument list - its receiver stays live and must keep reading tokens.
+///
+/// # Upstream Reference
+///
+/// - `options.c:2850-2851` - `if (write_batch < 0) "--only-write-batch=X"`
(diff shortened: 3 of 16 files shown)


oferchen:
## Root cause of the 120s `nextest` hang

It was a real deadlock, not a flake, and not a buffer-capacity problem.

`ReceiverContext` has two entry points. `run_pipelined` dispatches
`list_only` -> `only_write_batch` -> `dry_run` -> transfer.
`run_pipelined_incremental` only dispatched `list_only` -> `dry_run` -> transfer;
it never learned about `only_write_batch`.

Because `--only-write-batch=X` sets **both** `only_write_batch` and `dry_run` on
the server receiver (`server/run.rs`, mirroring `main.c:1839`), the incremental
path fell through to `run_dry_run_loop`. That loop writes a bare `NDX + iflags`
request and **no sum head**. The local sender, however, is *not* dry-run under
`write_batch < 0` - `do_xfers` stays 1 (`main.c:1839`) - so it goes down the real
transfer path and blocks reading the sum head it is owed
(`sender.c:442-443`, `write_ndx_and_attrs(f_out); write_sum_head(f_xfer)`).

Sender blocked on a read; receiver blocked on the echo read. Neither writes
again. Lockstep desync, both sides idle forever.

Blocked-process evidence (Linux host, before the fix):

```
pid 1591903 : oc-rsync -a --only-write-batch=... --rsh ... batchhost:/tmp/owb-d/dest/
   tid=1591903 state=S wchan=anon_pipe_read
pid 1591905 : oc-rsync --server -logDtpre.iLsfxCIvu --only-write-batch=X . /tmp/owb-d/dest/
   tid=1591905 state=S wchan=anon_pipe_read
```

Both in `anon_pipe_read`, i.e. two blocking reads - not a writer stalled on a
full pipe. That rules out the pipe/socket-buffer hypothesis (and it would have
pointed the wrong way anyway: macOS pipes start smaller than Linux's 64 KiB).

## Why CI failed and local runs passed

Nothing to do with the OS. It is feature unification.

The root package declares `transfer = { path = "crates/transfer", default-features = false }`,
and `incremental-flist` is in `transfer`'s **default** feature set. So:

| build | `transfer/incremental-flist` | receiver path | result |
|---|---|---|---|
| `cargo test -p bin ...` | off | `run_pipelined` | passes |
| `cargo nextest run -p bin --all-features` | off | `run_pipelined` | passes |
| `cargo nextest run --workspace --all-features` (CI) | **on** | `run_pipelined_incremental` | **hangs** |

Selecting `transfer` as a workspace member with `--all-features` turns
`incremental-flist` on for the whole unified build, including the `oc-rsync`
binary the test drives.

Cross-checks that confirm this:

* On a **GitHub `ubuntu-latest` x86_64 runner** (kernel 6.17), a build made with
  `cargo build --all-features --bin oc-rsync` (root package only, so
  `incremental-flist` off) completes the same push in ~1s:
  `RESULT only-write-batch: exit=0  dest files=0`. Same OS, same arch, same
  runner class as the failing job - no hang.
* On an idle Linux host, `cargo nextest run -p bin --all-features` passed; 60/60
  scripted pushes passed under saturating CPU load. Only once the binary was
  rebuilt with `cargo build --workspace --all-features` did it hang.
* macOS CI never ran this test at all - the macOS cell runs
  `-p core -p engine -p cli -p metadata -p apple-fs`, not the `bin` integration
  tests.

## Before / after, same nextest invocation

`cargo nextest run -p bin -p transfer --all-features` (i.e. `incremental-flist` on):

```
before:  test only_write_batch_push_leaves_the_destination_untouched ... FAILED
         thread ... panicked at tests/only_write_batch_remote_push.rs:182:42:
         push did not finish within the timeout
         test result: FAILED. finished in 120.02s

after:   Summary [0.060s] 3 tests run: 3 passed
```

Byte-identical to the CI failure (same line 182, same 120s).

## The fix

`run_pipelined_incremental` now dispatches `only_write_batch` before `dry_run`,
exactly as `run_pipelined` already did. Ten lines, receiver-side only; the
sender's divert semantics are unchanged (tokens to the batch, NDX/attrs on the
wire).

## Validation (Linux host, binary built with `--workspace --all-features`)

* `--only-write-batch` push: exit 0, 0 destination files, 40263-byte batch, 5/5 runs.
* `--read-batch` replay of that batch reproduces a tree that `diff -r` matches
  against the source.
* `--write-batch` (non-only) unchanged: exit 0, all files transferred, destination
  matches the source, batch recorded.
* oc push with `--only-write-batch` into a **real upstream 3.4.4 receiver**
  (`--rsync-path` pointing at rsync 3.4.4): exit 0, 0 destination files.
* `cargo fmt --all -- --check` clean; `cargo clippy -p core -p transfer
  --all-features --all-targets --no-deps -- -D warnings` clean.

## Separate pre-existing gap, not fixed here

While validating I found an unrelated fidelity gap in the same incremental
receiver: `create_directory_incremental` has no `skip_dest_writes()` guard,
unlike `create_directories` (`directory/creation.rs:136`). So on an
`incremental-flist` build the receiver mkdirs destination directories even when
no writes are allowed:

```
oc        --dry-run push: exit=0 dirs_created=1 files=0
upstream  --dry-run push: exit=0 dirs_created=0 files=0
```

This affects plain `--dry-run` independently of batching, and upstream
(`receiver.c:693`) creates nothing. Left out of this PR deliberately: the guard
has to be placed so `-n` itemize rows and `directories_created` stats are
preserved, which is a behaviour change worth its own change and its own test.

PR #6985, opened 2026-07-27
fix(batch): record the real sum_head for a delta --write-batch

### LINKED ISSUE #6978: fix(batch): upstream cannot replay an oc-rsync delta --write-batch file
## Summary

`oc-rsync --write-batch` produces a batch file that upstream rsync cannot replay
when the recorded transfer used delta encoding. Upstream either aborts with a
protocol error or crashes with SIGSEGV. Whole-file batches replay correctly, and
oc-rsync reads its own delta batches correctly, so the defect is confined to the
delta body that `--write-batch` records.

## Reproduction

Binaries used (banner-verified, not path-verified):

```
oc-rsync v0.6.4 (revision #7ea2c56f9) protocol version 32
rsync  version 3.4.4  protocol version 32
rsync  version 3.4.1  protocol version 32
```

```sh
mkdir -p w/{src,dest,final}
dd if=/dev/zero bs=1K count=100 2>/dev/null | tr '\0' 'A' > w/src/large.dat
cp w/src/large.dat w/dest/large.dat
cp w/src/large.dat w/final/large.dat
printf 'MODIFIED' | dd of=w/src/large.dat bs=1 seek=50000 conv=notrunc 2>/dev/null

oc-rsync -a --no-whole-file --ignore-times --write-batch=w/b w/src/ w/dest/
rsync -vv --read-batch=w/b w/final/
```

Observed:

```
delta-transmission enabled
Invalid block index 0 (count=0) [receiver]
rsync error: protocol incompatibility (code 2) at receiver.c(417) [receiver=3.4.4]
```

With a different fixture (100 KB of zeros patched with 100 random bytes at
offset 50000, `-av --no-whole-file --ignore-times`) upstream 3.4.1 and 3.4.4
both terminate with `Segmentation fault: 11` instead of a diagnostic.

## Analysis

The batch header and file list are byte-identical to upstream's for the same
transfer. The streams diverge at the sum_head that follows the file list:

| field     | upstream 3.4.4 | oc-rsync |
|-----------|----------------|----------|
| count     | 147            | 0        |
| blength   | 700            | 0        |
| s2length  | 2              | 16       |
| remainder | 200            | 0        |

oc-rsync records a whole-file-shaped sum_head (`count=0`, `blength=0`,
`s2length=16`) even though the transfer ran with `--no-whole-file` and produced
block matches, and the token body that follows is not upstream's delta encoding
either (it contains two multi-hundred-byte runs of `0xff` that upstream's
stream does not have). Upstream's receiver then reads a match token whose block
index cannot be valid for `count=0` and fails at `receiver.c:417`, or walks off
the end of the (zero-length) sum table and crashes.

## Scope

* Not reproducible with whole-file batches: `oc-rsync -av --write-batch=b src/ dest/`
  followed by `rsync --read-batch=b final/` succeeds and produces the right tree.
* oc-rsync replays its own delta batch correctly, so only cross-tool replay is
  affected.
* Unrelated to the cross-version batch header check: upstream 3.0.9 and 3.1.3
  reject a protocol-32 batch header from *any* writer, including upstream 3.4.4
  itself.

## Impact on CI

`tests/batch_interop_test.sh` covers this as the `oc-rsync -> upstream <version>`
leg. The leg is currently carried as an expected failure in
`tools/ci/known_failures.conf` so the script can enforce every other leg; the
harness reports an unexpected PASS as a failure, so the exemption will turn the
job red as soon as this is fixed and must be deleted in the same change.

oferchen:
Closes #6978.

## Problem

`--write-batch` on a delta transfer recorded a batch upstream cannot replay. Upstream 3.4.4 aborts:

```
Invalid block index 0 (count=0) [receiver]
rsync error: protocol incompatibility (code 2) at receiver.c(417) [receiver=3.4.4]
```

and upstream 3.4.1 is worse - it accepts the stream and reconstructs the wrong bytes.

## Root cause

`begin_batch_file_delta()` emitted a fixed, whole-file-shaped `sum_head` **before** the executor knew what the body would contain:

```rust
const FILE_SUM_LENGTH: i32 = 16;
let count: i32 = 0;
let blength: i32 = 0;
let s2length: i32 = FILE_SUM_LENGTH;
let remainder: i32 = 0;
```

The delta executor then appended block-match tokens against a basis the header claimed did not exist. For a 100 KiB basis:

| field | upstream 3.4.4 | oc before | oc after |
|---|---|---|---|
| count | 147 | **0** | 147 |
| blength | 700 | **0** | 700 |
| s2length | 2 | 16 | 16 |
| remainder | 200 | **0** | 200 |

`count`, `blength` and `remainder` now match upstream byte for byte. `s2length` still reports 16 because that is the strong-sum width oc's local basis signature genuinely uses; the batch carries no block sums, upstream's receiver never reads the field on the replay path, and narrowing it to upstream's phase-1 value would mean matching blocks on 2-byte sums with no phase-2 redo to catch a collision. Reporting a width the signature does not have would be the same class of defect as the one being fixed.

The token stream itself was already correct: on a fixture with distinct block content, oc's delta body is **byte-identical** to upstream's.

**This is not the same defect as #6967.** That one was about *where* the trailer is written on the remote tee paths (`core/client/run/batch.rs`, `transfer/generator/`). This one is in the local-copy batch composer (`engine/local_copy/context_impl/options/batch.rs`), which synthesises the stream rather than teeing it. They share a theme - the batch was composed by parts that never checked each other - but no code and no root cause.

## Why it shipped

oc's reader was tolerant in exactly the complementary way:

```rust
let block_length = if block_length_wire > 0 { .. } else { choose_block_length(entry_size) };
let remainder    = if remainder_wire   > 0 { .. } else { block_length };
```

A zeroed header was silently replaced with a guess, and `count` was never checked at all - `dispatch.rs` even computed `block_count as u32 - 1`, which underflows at `count = 0`. oc read its own malformed batches perfectly, so oc-to-oc testing could never surface this.

The CI harness could not surface it either: `tests/batch_interop_test.sh` replayed the batch **bare**, with none of the options it was recorded under. A delta batch must be replayed with its recorded option set - that is precisely what upstream's generated `<batch>.sh` companion re-runs (`batch.c:write_batch_shell_file`). A bare replay **segfaults 3.4.1 and 3.4.4 on a batch upstream wrote itself**, verified with `--no-inc-recursive` so the bare invocation is accepted at all, so the leg could never have passed regardless of the writer. That is corrected here.

## Fix

**One owner for the invariant.** The geometry, its field-range rules (`io.c:read_sum_head`) and the block-index rule (`receiver.c:414`) now live in `protocol::wire::SumHead`, the layer `engine`, `batch` and `transfer` all share. `SumHead::block_span()` is the only way a block index becomes a byte range, so a header that does not describe its body fails there instead of being worked around.

`transfer`'s existing `SumHead` keeps its public API but delegates encode/decode to the shared type instead of carrying a second copy of the rules - and it now checks on write what it checks on read.

**Writer.** `begin_batch_file_delta()` reserves the 16 bytes with upstream's all-zero `null_sum`; the delta executor records the basis geometry through `record_batch_delta_geometry()` as it matches; `finalize_batch_file_delta()` patches the reserved slot. Every match token is validated against the recorded head as it is written, so a drift fails while writing rather than shipping a batch that only upstream rejects. A whole-file body records no geometry and keeps the all-zero head - which also fixes `s2length`, previously written as 16 where `io.c:write_sum_head()` sends 0.

**Reader.** The guessed block length is gone (`choose_block_length` deleted - it had no remaining production caller), and every copy token resolves through `block_span()`. oc now rejects the pre-fix batch with upstream's own diagnostic:

```
Invalid block index 0 (count=0) while replaying '.../r.~batch-tmp'
```

## Verification

Binaries banner-verified, never by path:

```
oc-rsync v0.6.4 (revision #7ea2c56f9) protocol version 32
rsync  version 3.4.4  protocol version 32
rsync  version 3.4.1  protocol version 32
```

Both fixtures from #6978 (100 KiB of `'A'` patched at 50000, and 100 KiB of zeros patched with 100 random bytes) recorded with `-a --no-whole-file --ignore-times --write-batch` and replayed under genuine 3.4.1 **and** 3.4.4: **exit 0, byte-identical destination** in all four combinations. Whole-file batches still replay clean under 3.4.1, 3.4.4 and oc itself.

`tests/batch_interop_test.sh` against real 3.4.1 and 3.4.4 - all seven legs pass:

```
oc-rsync roundtrip (write-batch -> read-batch): PASS
oc-rsync -> upstream 3.4.1: PASS
oc-rsync -> upstream 3.4.4: PASS
upstream 3.4.1 -> oc-rsync: PASS
upstream 3.4.4 -> oc-rsync: PASS
upstream 3.4.1 -z -> oc-rsync (compressed batch): PASS
upstream 3.4.4 -z -> oc-rsync (compressed batch): PASS
```

The same harness run against a pre-fix binary reports 2 failures, so the leg genuinely detects this defect.

New tests fail against the pre-fix writer (checked by reverting the patch step):

- `crates/engine/tests/write_batch_delta_sum_head.rs` - the recorded header must carry the basis geometry, must never be whole-file-shaped, and must survive the strict reader.
- `crates/protocol/src/wire/delta/sum_head.rs` - encoding pinned to bytes observed from upstream 3.4.4, the `receiver.c:414` index rule, and per-field rejection of every impossible geometry.

`cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --all-features --no-deps -D warnings` are clean; `protocol`, `batch` and `transfer` are fully green.

Two `engine` permission tests (`execute_preserves_read_only_permissions`, `archive_preserves_mixed_permissions_across_files`) fail on this macOS host, and fail identically on a pristine `origin/master` worktree - pre-existing and unrelated.

## Notes

- The harness edit overlaps #6980, which makes these legs able to fail. If #6980 lands first this needs a trivial rebase; if this lands first, #6980 gets a leg that already passes.
- The 3.0.9 / 3.1.3 legs remain a genuine upstream limitation (`The protocol version in the batch file is too new (32 > 30)`, `compat.c(163)`), reproducible with a batch written by upstream 3.4.4 itself. Unrelated to this change and already scoped in `known_failures.conf`.
- Residual byte divergences from upstream in a delta batch, all pre-existing and accepted by upstream on replay: the 4-byte checksum seed, three bytes in the post-flist region, and the trailing stats values.

### DIFF
--- a/crates/batch/src/replay/delta.rs
+++ b/crates/batch/src/replay/delta.rs
@@ -6,7 +6,6 @@
 //!
 //! - [`apply_delta_ops`] writes the reconstructed file from a basis + ops.
 //! - [`write_literals_to_file`] handles the no-basis (literal-only) path.
-//! - [`choose_block_length`] mirrors upstream `match.c:choose_block_size()`.
 //! - [`default_xfer_sum_len`] returns the per-file transfer checksum length.
 
 use std::fs::{File, OpenOptions};
@@ -63,16 +62,29 @@ pub(super) fn write_literals_to_file(
     Ok(())
 }
 
+/// Widens a wire block index for [`protocol::wire::SumHead::block_span`],
+/// which mirrors upstream's signed `i < 0 || i >= sum.count` test.
+///
+/// A `u32` past [`i32::MAX`] is an index upstream would have read back as
+/// negative, so it maps to -1 and is rejected rather than truncated.
+pub(super) const fn block_index_as_i32(block_index: u32) -> i32 {
+    if block_index > i32::MAX as u32 {
+        -1
+    } else {
+        block_index as i32
+    }
+}
+
 /// Applies delta operations to reconstruct a file from a basis file.
 ///
 /// Reads copy and literal tokens from `delta_ops` and writes the
 /// reconstructed output to `dest_path`. Copy tokens reference blocks in
 /// `basis_path` at offsets computed as `block_index * block_length`.
 ///
-/// `block_count` is the number of blocks in the basis file's signature.
-/// `remainder` is the size of the last block (which may be shorter than
-/// `block_length`). For the last block (index == block_count - 1), the copy
-/// uses `remainder` bytes instead of `block_length`.
+/// `sum_head` is the block geometry the batch advertised for this file. Every
+/// copy token is resolved through it, so a token that references a block the
+/// header does not describe fails here rather than reconstructing the wrong
+/// bytes - the same check upstream performs at `receiver.c:414`.
 ///
 /// upstream: receiver.c:recv_files() / match.c - block_length for all blocks
 /// except the last, which uses remainder.
@@ -80,14 +92,13 @@ pub(super) fn write_literals_to_file(
 /// # Errors
 ///
 /// Returns [`BatchError::Io`] if the basis file cannot be opened, the output
-/// file cannot be created, or any read/write/seek operation fails.
+/// file cannot be created, any read/write/seek operation fails, or a copy
+/// token references a block outside `sum_head`.
 pub fn apply_delta_ops(
     basis_path: &Path,
     dest_path: &Path,
     delta_ops: Vec<protocol::wire::DeltaOp>,
-    block_length: usize,
-    block_count: u32,
-    remainder: usize,
+    sum_head: protocol::wire::SumHead,
 ) -> BatchResult<()> {
     let basis_file = File::open(basis_path).map_err(|e| {
         BatchError::Io(std::io::Error::new(
@@ -133,7 +144,19 @@ pub fn apply_delta_ops(
                 block_index,
                 length,
             } => {
-                let offset = u64::from(block_index) * (block_length as u64);
+                // Token-format block matches encode length=0 because the
+                // receiver derives the block span from the advertised
+                // geometry. Resolving through the sum_head is what rejects a
+                // header that does not describe this body.
+                // upstream: receiver.c:414-422
+                let (offset, span) = sum_head
+                    .block_span(block_index_as_i32(block_index))
+                    .map_err(|error| {
+                        BatchError::Io(std::io::Error::new(
+                            std::io::ErrorKind::InvalidData,
+                            format!("{error} while replaying '{}'", dest_path.display()),
+                        ))
+                    })?;
 
                 basis.seek(SeekFrom::Start(offset)).map_err(|e| {
                     BatchError::Io(std::io::Error::new(
@@ -142,17 +165,10 @@ pub fn apply_delta_ops(
                     ))
                 })?;
 
-                // Token-format block matches encode length=0 because the
-                // receiver derives block size from the signature. Use
-                // block_length for all blocks except the last, which uses
-                // remainder (the last block is typically shorter).
-                // upstream: receiver.c - block size for last block is remainder.
                 let effective_length = if length > 0 {
                     length as usize
-                } else if block_count > 0 && block_index == block_count - 1 {
-                    remainder
                 } else {
-                    block_length
+                    span as usize
                 };
                 let mut remaining = effective_length;
                 while remaining > 0 {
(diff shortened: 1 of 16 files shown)


oferchen:
Rebased onto current master (post-#6980/#6983/#6986/#6987) and reconciled with #6980's harness restructure:

- The cross-tool replay now uses the recorded option set (`-a --no-whole-file --ignore-times`), inside #6980's `record_fail`/`record_pass` XFAIL machinery - both properties survive.
- Deleted the now-obsolete `batch:oc-to-upstream` exemption for upstream > 3.1.3 from `tools/ci/known_failures.conf`, exactly as the issue required ("must be deleted in the same change"); with #6980's unexpected-pass rule the entry would otherwise turn the job red. The protocol-32 header rejection for <= 3.1.3 stays.
- Combined harness re-verified locally against banner-checked 3.4.1/3.4.4: exit 0 with this branch's binary (7 PASS); exit 1 against the pre-fix binary with both oc-to-upstream legs failing hard. 3.0.9/3.1.3 legs XFAIL through the retained exemption.

PR #6991, opened 2026-07-28
feat(rsync_io): add feature-gated quic transport skeleton

oferchen:
## What

Measurement spike for a QUIC transport: quinn + rustls (ring provider) behind an off-by-default `quic` feature on `rsync_io`, wired exactly like the existing `embedded-ssh` optional transport. It establishes a loopback connection with an in-process self-signed certificate (ALPN `rsync`), opens one bidirectional stream, and exposes it as blocking `Read` + `Write` with the tokio current-thread runtime confined to the module. A deterministic loopback round-trip integration test covers both directions.

This is NOT the transport itself: no certificate-file loading, no CLI flags, no daemon integration, no config surface. The point is to measure what QUIC costs the build before committing to a design.

**Not for merge until the transport design is approved.**

## Feature-gating guarantee

`quic` is not in any default feature set and nothing forwards to it from `core` or the root manifest. Measured proof that a default build is untouched:

- `cargo tree -p rsync_io -e normal | grep -c quinn` -> **0**
- `cargo tree --workspace -e normal | grep -c quinn` -> **0**
- `cargo tree --workspace --all-features -e normal | grep -ci aws-lc` -> **0** (the rustls aws-lc-rs default provider stays out even with the feature on; both quinn and rustls are pinned `default-features = false` + ring)

The default `oc-rsync` binary is unchanged by construction: the `bin` crate does not depend on `rsync_io` directly (only via `core`, which is untouched), so the feature is unreachable from the binary and no root passthrough feature is added.

## Measurements (macOS aarch64, rustc 1.88.0, n=1)

| Metric | Without `quic` | With `quic` | Delta | Command |
|---|---|---|---|---|
| Clean release build, wall | 34.95 s | 38.88 s | +3.9 s (+11%) | `/usr/bin/time -p cargo build -p rsync_io --release [--features quic]` (fresh `CARGO_TARGET_DIR`) |
| Clean release build, user CPU | 33.8 s | 69.5 s | +35.7 s (~2x) | same |
| `librsync_io.rlib` | 1,469,168 B | 2,100,480 B | +631,312 B | `ls -l target/release/librsync_io.rlib` |
| Total release target dir | 74,564 KiB | 175,228 KiB | +100,664 KiB | `du -sk` |
| Crate graph of `rsync_io` (`--no-default-features`) | 32 | 69 | +37 | `cargo tree -p rsync_io -e normal --no-default-features [--features quic] --prefix none \| sort -u \| wc -l` |
| Crates new to the whole workspace default graph | - | - | **10** | set-diff vs `cargo tree --workspace -e normal` |

The 10 genuinely new crates: `quinn`, `quinn-proto`, `quinn-udp`, `rustls`, `rustls-webpki`, `rustls-pki-types`, `rcgen`, `lru-slab`, `rand_pcg`, `yasna`. The other 27 of the +37 (tokio, ring, zeroize, ...) are already in the workspace graph via `embedded-ssh`/russh. Binary-size delta on `oc-rsync` could not be measured directly because the feature is unreachable from the binary (see above); rlib sizes are reported instead.

Pinned versions and their MSRV (all within the 1.88.0 toolchain pin): quinn 0.11.11 / quinn-proto 0.11.16 / quinn-udp 0.5.15 (`rust-version = 1.85`), rustls 0.23.42 (1.71), rustls-webpki 0.103.13 (1.71), rustls-pki-types 1.15.1 (1.60), rcgen 0.14.8 (1.88).

## CI coverage

No workflow file is touched. The feature is exercised automatically by the two workspace-level `--all-features` cells in `ci.yml`: `fmt + clippy` (`cargo clippy --locked --workspace --all-targets --all-features`) and `nextest (stable)` (`cargo nextest run --locked --profile ci --workspace --all-features`), both on ubuntu. Note: the Windows, macOS, and musl cells build per-package or fixed feature sets that do not reach `rsync_io/quic`, so cross-platform cells compile `rsync_io` only with default features; QUIC build/test coverage in this PR is Linux gnu.

## Design note recorded for the follow-up

The confined current-thread runtime works but its driver only makes progress while a `block_on` call is in flight; the spike needed an explicit `finish()` delivery barrier and a graceful `close()` terminal operation to make teardown deterministic (a peer that stops polling otherwise leaves the other side waiting for its idle timeout). The production design will evaluate driving `quinn-proto` sans-IO from a plain thread instead.


oferchen:
Superseded by #6999. The QUIC-2 concurrency-model evaluation selected the sans-IO driver-thread design over this feature-gated confined-runtime skeleton; #6999 carries the complete quic feature (module, driver, tests, deps) in that final form as standalone new files. Merging this skeleton would conflict with #6999 (both create crates/rsync_io/src/quic/mod.rs and declare the quic feature) and reintroduce the confined runtime that #6999 intentionally omits. Closing so #6999 is the single source.

PR #6998, opened 2026-07-28
docs(design): QUIC transport certificate, ALPN, and scheme policy note

oferchen:
QUIC-4 decision note fixing the policy surface for the native QUIC transport, ahead of the build-out tasks. Decision note only - no code, CLI flags, or config directives are introduced; the note is awaiting project-owner review before any of it becomes code.

Framing: a native QUIC transport is an oc extension that adds reach - it carries the unmodified daemon wire protocol over QUIC instead of TCP. Upstream deliberately does no in-daemon TLS (it fronts the daemon with an SSL proxy plus the rsync-ssl client helper), so where upstream has a name we mirror it and where it has none the new surface is justified as an extension.

The five decisions, each with recommendation:

- A. Server identity - `quic cert file` / `quic key file` global directives (matching the existing space-separated, whitespace-normalised directive style); zero-config default is a self-signed rcgen cert persisted beside the config so identity is stable across restarts, with an ephemeral per-boot fallback (loud warning) when the state dir is read-only.
- B. Client verification - system roots by default, `--quic-ca <pem>` for private CAs, and TOFU known-hosts (`quic_known_hosts`, SPKI pin) as the primary self-signed path mirroring SSH; `--quic-known-key=SHA256:...` as the single escape hatch; no blanket insecure flag.
- C. ALPN - a single fixed token `rsync` (the in-band handshake remains the sole version authority; a version-bearing ALPN token would be a second source of truth); a mismatch fails at the TLS layer and maps to exit 5.
- D. Scheme/flag - `quic://host[:port]/module` plus a `--quic` modifier; a QUIC-selected transfer to a non-QUIC endpoint fails loudly and never falls back to TCP (downgrade hazard); default `873/udp`; dedicated `quic = yes` / `quic port` / `quic listen` daemon directives so TCP and QUIC listeners can coexist.
- E. Out-of-scope ledger - 0-RTT, connection migration, mTLS/client certs, and MASQUE, each deferred with the trigger that would reopen it.

Surfaced precedent conflicts: default port 873/udp vs rsync-ssl's 874 (explained); in-daemon cert directives have no upstream analogue (justified as an extension, named after rsync-ssl's RSYNC_SSL_CERT/_KEY/_CA_CERT vocabulary); client verification exposed as CLI flags rather than rsync-ssl's env vars, matching oc-rsync's own conventions.

### DIFF
--- a/docs/design/quic-transport-policy.md
+++ b/docs/design/quic-transport-policy.md
@@ -1,0 +1,333 @@
+# QUIC Transport: Certificate Strategy, ALPN, and Scheme Policy
+
+Tracking task: QUIC-4. Status: decision note awaiting owner review. No code, no
+CLI flags, no config directives are introduced by this note - it fixes the
+policy surface so the build-out tasks (QUIC-5 through QUIC-11) implement one
+agreed design rather than discovering it.
+
+## 0. Framing: where QUIC sits relative to upstream
+
+Upstream rsync deliberately does **not** terminate TLS inside the daemon. Its
+own answer to "encrypt an rsync-daemon connection" is a front-end proxy
+(haproxy/nginx) plus the `rsync-ssl` client helper script; `rsyncd.conf(5)` is
+explicit that the daemon "only encrypts ... if you put rsync behind an SSL
+proxy" and documents no in-daemon certificate parameters. The client helper
+speaks to that proxy over stunnel/openssl/gnutls and exposes exactly four
+knobs, all environment variables: `RSYNC_SSL_CERT`, `RSYNC_SSL_KEY`,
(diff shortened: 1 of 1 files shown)


oferchen:
Decision points A-E reviewed and approved on 2026-07-28. This is the agreed policy surface for the native QUIC transport: (A) `quic cert file` / `quic key file` directives with a persisted self-signed default, (B) system roots + `--quic-ca` + TOFU known-hosts, no blanket insecure flag, (C) single ALPN token `rsync`, (D) `quic://` + `--quic` with no silent TCP fallback on `873/udp`, (E) 0-RTT / migration / mTLS / MASQUE deferred. QUIC-5/6 build to this spec.

PR #7017, opened 2026-07-28
perf(transfer): share the receiver file list via Arc instead of deep-cloning it

oferchen:
## Problem

The pull receiver kept **two full copies** of the file list resident, doubling
its peak flist RSS (measured ~460 MB at 1M files vs ~227 MB for a local copy):

1. `ReceiverContext::file_list` — the owned list, held for the whole session.
2. `Arc::new(self.file_list.clone())` at `receiver/transfer/pipeline.rs:210` —
   a **deep** clone of the entire `Vec<FileEntry>` handed to the disk-commit
   pipeline (`DiskCommitConfig.file_list`) and kept alive for the full
   disk-commit window.

Because each `FileEntry` owns a heap-allocated `name`, that clone duplicates the
whole flist. The arithmetic matches the profile: receiver ≈ 2× the flist.

## Fix

A lifetime/sharing change, not an algorithm change. The field becomes
`Arc<Vec<FileEntry>>` and the disk-commit config receives a pointer clone
(`Arc::clone(&self.file_list)`) instead of a deep copy, so a single flist
allocation is shared between the receiver and the commit thread.

- Reads are unaffected: `Arc<Vec<FileEntry>>` derefs to `[FileEntry]`, so
  indexing, `.iter()`, `.len()`, `.get()` compile unchanged. Only mutations and
  assignments were touched.
- Construction mutations (reception, sort/clean, prune, hardlink matching)
  go through `Arc::make_mut`, which returns the existing allocation without
  copying because the receiver uniquely owns the Arc at that point.

## Mutation-safety check

Every mutation of `self.file_list` was audited against the point where the Arc
is shared with the disk-commit (`pipeline.rs:210`):

- **Before the share (construction):** `receive_file_list`, `sanitize_file_list`,
  the sort/clean pass, `prune_empty_dirs_pass`, `match_hard_links`,
  `normalize_pre30_hardlinks`, id remapping, and the INC_RECURSE sub-list
  materialization (`ensure_all_segments_loaded`, which upstream and this code
  both complete *before* the transfer loop). All run while the receiver is the
  sole owner.
- **During the share (transfer loop):** none. `run_pipeline_loop_decoupled`
  borrows `&self` and holds only shared `&FileEntry` references into the list,
  so the borrow checker guarantees the list is read-only for the whole window
  the Arc is shared.
- **After the share (finalization):** `reclaim_oldest_segment` frees per-entry
  heap for completed INC_RECURSE segments. It runs in `exchange_phase_done`,
  after `run_pipelined` has called `pipelined_receiver.shutdown()` (joining the
  disk thread and dropping its Arc clone). It now mutates through
  `Arc::get_mut` and simply skips the (pure-RSS) reclaim if a clone somehow
  lingers, rather than deep-cloning the list — which would defeat the sharing.

No late mutation required relocation; the sharing window is provably
read-only.

## Correctness

This cannot change wire bytes, sort order, itemize rows, or stats — it only
changes how the one list is owned. The receiver disk-commit, incremental,
directory, hardlink, sanitize, and golden itemize test suites pass unchanged.

## Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p transfer --all-targets --all-features --no-deps -- -D warnings` — clean
- `cargo nextest run -p transfer --all-features` (receiver / pipeline / disk-commit /
  incremental / directory / itemize / hardlink / sanitize subsets) — all pass,
  including `itemize_rows_match_upstream_3_4_4_golden`.

The ~184 MB/1M RSS recovery (receiver 460 MB -> ~227 MB) is host-verified
separately in #208.

### DIFF
--- a/crates/transfer/src/receiver/context.rs
+++ b/crates/transfer/src/receiver/context.rs
@@ -45,7 +45,15 @@ pub struct ReceiverContext {
     /// Server configuration.
     pub(in crate::receiver) config: ServerConfig,
     /// List of files to receive.
-    pub(in crate::receiver) file_list: Vec<FileEntry>,
+    ///
+    /// Shared via `Arc` with the disk-commit pipeline (`DiskCommitConfig`) so a
+    /// pull transfer keeps a single resident copy of the flist rather than
+    /// deep-cloning it for the commit thread. The receiver uniquely owns the
+    /// `Arc` while it builds, sorts, and cleans the list (see
+    /// [`file_list_mut`](Self::file_list_mut)); it is shared only for the
+    /// read-only transfer window and reclaimed afterward once the disk thread
+    /// has released its clone.
+    pub(in crate::receiver) file_list: Arc<Vec<FileEntry>>,
     /// Negotiated checksum and compression algorithms from Protocol 30+ capability negotiation.
     /// None for protocols < 30 or when negotiation was skipped.
     pub(in crate::receiver) negotiated_algorithms: Option<NegotiationResult>,
@@ -430,7 +438,7 @@ impl ReceiverContext {
         Self {
             protocol: handshake.protocol,
             config,
-            file_list: Vec::new(),
+            file_list: Arc::new(Vec::new()),
             negotiated_algorithms: handshake.negotiated_algorithms,
             compat_flags: handshake.compat_flags,
             checksum_seed: handshake.checksum_seed,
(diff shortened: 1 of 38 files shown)


oferchen:
Closing: the ~184MB/1M receiver-RSS recovery this PR was built on does NOT materialize under measurement. Fair back-to-back A/B (constant sender, master-client vs branch-client, verified exit=0 + file counts): 100k identical (103484 vs 103448 KB), 1M the branch is ~30MB HIGHER not lower (486560 vs 517016 KB); cross-checks vs an upstream sender and branch-as-both are all identical. The deep clone removed here (pipeline.rs:210) is not resident at the RSS peak — the peak is set earlier during flist reception + sort_and_clean, and the clone happens later reusing freed heap, so VmHWM is unchanged. The change is correct and harmless (dropping a redundant deep clone), but a 38-file field-type migration is not justified by a non-existent RSS win. The real receiver-RSS structural win remains the inc_recurse O(N)->O(segment) work (#102/#105), which is unaffected. Reopen if a measured CPU/wall-time benefit for removing the clone is later shown.

PR #7075, opened 2026-07-29
refactor: remove inert consecutive-match algorithm

oferchen:
## Summary

Removes the inert "consecutive-match" delta algorithm. This was a private,
oc-only extension gated on a `CAP_CONSECUTIVE_MATCH` compatibility bit that
was advertised through an `-e` capability letter. That capability letter has
since been removed for byte-identity with upstream rsync, so nothing ever
sets the `CONSECUTIVE_MATCH` compat bit anymore and every code path guarded
by it is unreachable.

## Why the code was dead

- The `CONSECUTIVE_MATCH` bit (`0x0200_0000`) is deliberately excluded from
  `KNOWN_MASK`/`ALL_KNOWN` and was only ever OR-ed into advertised flags by
  the now-removed capability letter. The only remaining places that set the
  bit were the tests that exercised the algorithm.
- Negotiated flags are the AND of both peers' advertised sets, so with no
  advertiser the bit can never survive negotiation:
  - `effective_s2length()` always returned the full length,
  - the generator's gating threshold was always `1` (upstream behaviour),
  - the receiver's strong-sum halving never triggered.

## What was removed

- **protocol**: the `CompatibilityFlags::CONSECUTIVE_MATCH` constant, the
  `effective_s2length()` strong-sum halving helper, their re-exports, and the
  associated unit tests.
- **matching**: `DeltaGenerator::consecutive_match_needed` field, the
  `with_consecutive_match_needed()` builder, the `>= 2` gating branches in
  `generate` / `generate_chunked`, the entire `generate_gated()` scan, and the
  tests that only exercised the gated path.
- **transfer**: the `consecutive_match_needed()` helper and its call sites in
  the delta generator, and the receiver-side basis strong-sum halving that
  consumed `effective_s2length`.
- **transfer**: the now-unused `BasisFileConfig.compat_flags` field (its sole
  consumer was the halving) and its wiring in `context.rs` and `pipeline.rs`.
- **docs**: the stale fallback bullet in `intra-file-parallelism.md` that
  documented the removed gated-scan routing.

Net: 5 insertions, 687 deletions.

## Behaviour

Pure dead-code removal. Observably neutral - wire bytes, matched/literal
stats, stdout, and exit codes are unchanged, because the gate bit was never
negotiated on any live path. The separate seq-match coalescing / `extend_run`
index feature is untouched.

## Verification

- `cargo-fmt --all` clean.
- `cargo check -p matching -p transfer -p protocol --all-features` compiles
  clean with no warnings.

### DIFF
--- a/crates/matching/src/generator.rs
+++ b/crates/matching/src/generator.rs
@@ -224,17 +224,6 @@ fn merge_copy_runs(source: &[u8], mut runs: Vec<CopyRun>, block_len: usize) -> D
 #[derive(Clone, Debug)]
 pub struct DeltaGenerator {
     buffer_len: usize,
-    /// Consecutive-match gating threshold (zsync `seq_matches`).
-    ///
-    /// `1` (default) reproduces upstream's accept-on-first-match behaviour and
-    /// leaves the scan byte-for-byte unchanged. `2` engages the oc-only
-    /// consecutive-match extension: a block match is only trusted (emitted as a
-    /// Copy) when it belongs to a run of at least two consecutive matching
-    /// blocks; a lone match is demoted to a Literal of its own source bytes.
-    /// The threshold is set to `2` only when the mutually negotiated
-    /// `CAP_CONSECUTIVE_MATCH` bit is present, which is also the sole condition
-    /// under which the receiver halves the per-block strong-sum length.
-    consecutive_match_needed: u8,
     /// Enforces upstream's in-place ("updating basis file") monotonicity guard.
     ///
     /// When `true` the scan mirrors `match.c:211` (`updating_basis_file`): a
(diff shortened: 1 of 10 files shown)


oferchen:
Holding: pending a design decision on whether oc should carry an upstream-ignored oc-peer marker in the -e string (which would re-enable the consecutive-match oc<->oc optimization) vs strict -e byte-identity (in which case this deletion is correct). Do not merge until resolved.

oferchen:
Closing: the consecutive-match algorithm is a deliberate oc-rsync optimization (extension pack), not dead code. It must NOT be deleted. Instead the oc-peer marker negotiation removed in #7065 needs to be restored so the feature is reachable oc<->oc again (upstream ignores the marker; oc<->upstream stays byte-identical when the extension is off).

PR #7173, opened 2026-08-05
refactor(daemon): emit @ERROR lines through a typed formatter with round-trip test

oferchen:
## Summary

Every daemon refusal in the ASCII negotiation phase is an `@ERROR: <msg>\n` line the client treats as fatal (upstream: `clientserver.c:381-385`). The `@ERROR: ` prefix and the terminating newline were previously spelled out at every emit site with an ad-hoc `format!("@ERROR: ...")` plus a separate `\n`, so no single point guaranteed the exact bytes.

This introduces `AtError`, a typed formatter that owns the framing, and routes every negotiation-phase emit site through it.

## What changed

- New `crates/daemon/src/daemon/at_error.rs`: an `AtError` enum whose variants each map to one upstream refusal. Fixed-wording variants borrow the single-sourced `*_PAYLOAD` constants; dynamic variants carry only their operands. `to_wire()` renders the exact `@ERROR: <msg>\n` bytes and `Display` renders the newline-free line for logging/assertions. A freeform `Message` variant covers call-site-composed bodies (refused options, omitted greeting tokens, per-session OS error strings) and strips a redundant leading `@ERROR: ` so the prefix is applied exactly once.
- `send_error` now takes `&AtError` instead of a pre-formatted `&str`, so the sink is the single place that appends the newline.
- Routed all remaining hand-built sites through the formatter: the two raw-socket max-connections paths (sync accept loop + async refusal) and the async-session refusal that previously wrote `"@ERROR: ...\n"` by hand.

## Fidelity

The emitted bytes are unchanged and remain byte-identical to upstream. The wording pin test (`daemon_error_payloads_match_upstream_wording`) and the max-connections golden-byte test (`daemon_max_connections_wire_bytes_match_upstream`) still pass unmodified.

## Round-trip coverage

Adds `every_variant_formats_and_parses_back`: for each variant it asserts the golden `@ERROR: <body>\n` bytes, that the line ends in exactly one newline, and that feeding the bytes back through the client parse path (`protocol::parse_legacy_error_message`) recovers the original body verbatim - pinning the formatter and parser as true inverses.

## Out of scope

The multiplexed `ERROR: module is ...` rejections and the incomplete-build handshake placeholder travel inside `MSG_ERROR` frames without a trailing newline, so they are intentionally not routed through this line formatter.

## Verification

- `cargo fmt --all -- --check`: clean
- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings`: clean
- `cargo check --target x86_64-pc-windows-gnu -p daemon --all-features`: clean
- `cargo check --target x86_64-unknown-linux-musl -p daemon --all-features`: clean
- `cargo nextest run -p daemon --all-features`: 351 targeted tests pass (error/at_error/greeting/handshake/auth/max-connections/refuse); round-trip + golden gate tests pass
- `cargo build --bin oc-rsync --all-features`: builds

### DIFF
--- a/crates/daemon/src/daemon.rs
+++ b/crates/daemon/src/daemon.rs
@@ -69,8 +69,11 @@ use crate::{
     systemd,
 };
 
+mod at_error;
 mod help;
 pub(crate) mod tracing_stream;
+
+pub(crate) use self::at_error::AtError;
 
 /// Concurrent session tracking for the daemon accept loop.
 #[cfg(feature = "concurrent-sessions")]
@@ -724,11 +727,11 @@ fn refuse_async_connection_at_capacity(
     limit: Option<usize>,
 ) {
     let limit = limit.unwrap_or(0);
-    let payload = format!(
-        "{}\n",
-        MODULE_MAX_CONNECTIONS_PAYLOAD.replace("{limit}", &limit.to_string())
-    );
-    let _ = stream.write_all(payload.as_bytes());
+    let payload = AtError::MaxConnections {
+        limit: limit as i64,
+    }
+    .to_wire();
+    let _ = stream.write_all(&payload);
     let _ = stream.flush();
 }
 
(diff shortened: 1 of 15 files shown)


oferchen:
Closing as duplicate: task #167 is already delivered by PR #7174 (refactor/daemon-typed-at-error). This was a redundant backup run.

PR #7191, opened 2026-08-05
fix(engine): order -R multi-source operands to match upstream itemize

oferchen:
## Problem

With `--relative` (`-R`), a multi-source local copy visited its operands in
command-line order, while upstream rsync emits them in name order. Upstream's
`send_file_list()` accumulates every source operand into one file list and
sorts it globally with `f_name_cmp` (`flist_sort_and_clean`, flist.c:2544)
before the generator itemizes it, so top-level operands appear sorted
regardless of the order given on the command line.

An earlier change sorted multi-source operands for non-relative transfers but
deliberately gated out `-R`, because the relative case also emits implied
parent directories. As a result:

```
# rsync 3.4.4 (reference)          # oc-rsync (before)
$ rsync -aiR -n b/x a/y dst/       $ oc-rsync -aiR -n b/x a/y dst/
a/                                 b/
a/y/                               b/x/
a/y/f3                             b/x/f1
a/y/g                              a/
b/                                 a/y/
b/x/                               a/y/f3
b/x/f1                             a/y/g
```

The files land identically either way; only the observable itemize / verbose /
`--list-only` order diverged, which the output-fidelity contract forbids.

## Fix

Extend the operand sort to the `-R` case. A directory operand's subtree is
emitted contiguously and name-sorted within, and two operands whose relative
paths diverge sort as whole blocks, so ordering the operands by the exact key
upstream would give their flist row reproduces the global sort. The only
difference from the non-relative case is what that row is named:

- without `--relative`: the destination basename (`flist.c` send_file_name
  keeps the operand tail);
- under `-R`: the operand's full relative path (`send_implied_dirs` + the
  leaf), so `a/y` sorts before `b/x`, and `b/w` before `b/x` under a shared
  `b/` prefix.

The reorder is a pure output-ordering change: destinations are byte-identical,
the `--delete` keep-set is order-independent, and directory creation follows
the same sorted order upstream uses.

## Scope

Copy-contents (trailing-slash) operands still keep command-line order. Their
cross-operand content interleave cannot be reproduced by ordering the operands
and needs a separate local-copy flist-collect refactor (tracked separately). A
pre-existing, unrelated divergence remains when one operand is a strict
ancestor directory of another (`-aR a a/y`): the descendant operand's subtree
is listed twice; that is a dedup issue, not an ordering one, and is out of
scope here.

## Tests

`crates/engine/tests/multi_source_operand_order.rs`:

- `relative_multi_source_distinct_prefixes_match_upstream_order` - `-aR b/x
  a/y` itemizes `a, a/y, a/y/f3, b, b/x, b/x/f1` (the exact upstream 3.4.4
  order), failing before the fix.
- `relative_multi_source_shared_prefix_match_upstream_order` - shared-prefix
  operands sort `b/w` before `b/x`.
- `relative_operands_are_ordered_by_full_relative_path` - replaces the old
  scope-boundary test that pinned command-line order.

The existing non-relative operand-sort tests are unchanged and stay green.

## Verification

Against a real rsync 3.4.4 on Linux x86_64: `-aiR -n` multi-source itemize
with distinct prefixes and with a shared prefix now match the reference binary
byte-for-byte (both previously in argument order). `cargo fmt --check`;
`clippy --workspace --all-targets --all-features -D warnings` and default
features, both clean; `cargo check` for `x86_64-pc-windows-gnu` and
`x86_64-unknown-linux-musl` (lib+bins+tests); `nextest -p engine -p cli` on
default features (8699) and `--all-features` (8784), all passing.

### DIFF
--- a/crates/engine/src/local_copy/executor/sources/orchestration.rs
+++ b/crates/engine/src/local_copy/executor/sources/orchestration.rs
@@ -52,52 +52,79 @@ fn whole_unix_seconds() -> u64 {
 /// the same comparator the network generator applies to its flist
 /// (`protocol::flist::compare_file_entries`, transfer `generator/file_list`).
 ///
-/// Scope: applied only when every operand contributes a single NAMED top-level
-/// entry - a non-`--relative` transfer with no trailing-slash / copy-contents
-/// operand. A copy-contents operand merges its CONTENTS into the destination,
-/// which upstream interleaves across operands by content name; that cannot be
-/// reproduced by ordering the operands, so those transfers keep command-line
-/// order. `--relative` operands carry implied parent directories whose ordering
-/// is handled during emission, so they are left untouched. Reordering is a
-/// pure output-ordering change: the destination result is byte-identical, the
-/// `--delete` keep-set is order-independent, and dir creation simply follows
-/// the same sorted order upstream uses.
+/// Scope: applied whenever every operand contributes a NAMED top-level subtree,
+/// i.e. no operand is a trailing-slash / copy-contents operand. A copy-contents
+/// operand merges its CONTENTS into the destination, which upstream interleaves
+/// across operands by content name; that cannot be reproduced by ordering the
+/// operands, so those transfers keep command-line order (handled in #212).
+///
+/// Both the plain and `--relative` (`-R`) cases reduce to the same rule because
+/// upstream sorts the whole flist globally: a directory operand's subtree is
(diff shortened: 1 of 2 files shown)


oferchen:
Superseded by #7180, which merged the same -R multi-source operand-ordering fix (task landed, master evolved the test into the copy-contents ignored-spec). This branch is a parallel-developed duplicate and no longer applies cleanly.

PR #7211, opened 2026-08-05
fix(build): forward workspace embedded-ssh feature to cli, not just core

oferchen:
## Problem

`cargo nextest run --locked --profile ci --workspace` fails on
`mutually_exclusive_options::test_rsh_and_ssh_url_operand_reports_missing_embedded_ssh`
(crates/cli/tests, panics at the "ssh:// URLs require the built-in SSH client"
assertion).

Root cause is a **workspace feature-unification skew**. The root `Cargo.toml`
has `embedded-ssh` in its default set, but the feature forwarded only to
`core/embedded-ssh`, not `cli/embedded-ssh`. Under `--workspace`, cargo unifies
features per crate: `core/embedded-ssh` resolved ON (via the root default) while
`cli/embedded-ssh` stayed OFF. So the cli-crate test gated
`#[cfg(not(feature = "embedded-ssh"))]` still compiled and ran (cli saw the
feature absent), but `core` — with embedded-ssh present — emitted the `-e`
transport-conflict message instead of the "requires embedded-ssh" one, failing
the assertion. Running the same test as `-p cli` passes because there both
crates resolve the feature OFF; only the unified `--workspace` build exposes the
skew.

## Fix

Forward the root `embedded-ssh` feature to **both** `cli` and `core`
(`["cli/embedded-ssh", "core/embedded-ssh"]`), so a default/`--workspace` build
enables the feature consistently in every crate that gates on it. The
`not(feature = "embedded-ssh")` test then correctly compiles out in the
default/CI build (embedded-ssh IS present) and only runs under
`--no-default-features`, matching the intent. No source change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,8 +111,13 @@ openssl-vendored = ["checksums/openssl-vendored"]
 # ============================================================================
 
 # Embedded SSH transport via russh - removes the runtime dependency on an
-# external `ssh` client. Forwards through `core` to `rsync_io/embedded-ssh`.
-embedded-ssh = ["core/embedded-ssh"]
+# external `ssh` client. Forwards to BOTH `cli` and `core` so a workspace
+# feature-unified build (e.g. `cargo nextest --workspace`) enables the feature
+# consistently across every crate that gates on it. Forwarding to `core` alone
+# left `cli/embedded-ssh` OFF while `core/embedded-ssh` was ON under `--workspace`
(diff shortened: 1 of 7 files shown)


oferchen:
Superseded: master already forwards `embedded-ssh` to both `cli` and `core` (Cargo.toml:117). The workspace feature-unification fix this PR carried is already present on master; only a comment reword remained, not worth a separate PR.

PR #7226, opened 2026-08-06
fix(cli): one match_report total: emitter with upstream gate + counters

oferchen:
## Upstream ground truth

`match_report()` (`match.c:439-448`) is the *only* producer of the `total:` line, it is called from exactly one place - `sender.c:491`, after `send_files()` has finished every file - and it is gated on `DEBUG_GTE(DELTASUM, 1)` alone:

```c
void match_report(void)
{
	if (!DEBUG_GTE(DELTASUM, 1))
		return;

	rprintf(FINFO,
		"total: matches=%d  hash_hits=%d  false_alarms=%d data=%s\n",
		total_matches, total_hash_hits, total_false_alarms,
		big_num(stats.literal_data));
}
```

Note the double space after `matches=%d` and `hash_hits=%d`, the single space before `data=`, and that `--whole-file` does **not** suppress the line - it merely leaves the counters at their zero initial values. The per-file line at `match.c:428-431` is a different, `DELTASUM >= 2` diagnostic with the reversed field order `false_alarms / hash_hits / matches`.

MEASURED against rsync 3.4.4:

| invocation | client output |
|---|---|
| `rsync -rvv src/ dst/` (local, whole-file) | `total: matches=0  hash_hits=0  false_alarms=0 data=17` |
| `rsync -rvv --no-whole-file` (local, fresh dest) | `total: matches=0  hash_hits=0  false_alarms=0 data=17` |
| `rsync -rvv --no-whole-file` (local, real basis) | `total: matches=1  hash_hits=1  false_alarms=0 data=700` |
| `rsync -rvv --no-whole-file` (daemon **push**) | `total: matches=1  hash_hits=1  false_alarms=0 data=700` |
| `rsync -rvv --no-whole-file` (daemon **pull**) | *no line* - the sender is remote |

Position in every emitting case: after the per-file name list, before the blank line and the `sent ... received ...` trailer.

## What oc had

Two places claimed to implement `match_report()`, on disjoint paths, which is why a count-based guard never caught it:

1. **`crates/cli/src/frontend/progress/render.rs`** - local-copy path. Right text, right position, but it hardcoded `matches=0  hash_hits=0  false_alarms=0` and was additionally gated on whole-file mode, so a local `--no-whole-file` delta printed no line at all.
2. **`crates/matching/src/generator.rs`** - network sender path, commented `// upstream: match.c match_report() equivalent`, emitting once **per file** at `DELTASUM` level 1:

   ```
   delta: 2 tokens, 1400 total, 700 literal, 700 matched
   ```

   Wrong text, wrong cardinality, and wrong position: the scanner's `debug_log!` lands dead-last through the deferred diagnostic flush, *after* the summary trailer.

## What this PR does

**One emitter**, the renderer's - the only one that can place the line where upstream puts it.

- The local-copy delta loop now tallies real `matches` / `hash_hits` / `false_alarms` through a counted probe and folds them into run totals on `LocalCopySummary`, mirroring `match.c:433-435`. The renderer prints them.
- The gate loses its `whole_file` condition, matching `match.c:440`. A whole-file transfer still prints the line, with zeros.
- The invented per-file `delta: ...` line is deleted, leaving exactly one `total:` producer in the tree.

MEASURED after the change, byte-identical to upstream on all three local cases, including `total: matches=1  hash_hits=1  false_alarms=0 data=700`.

`matches` and `data` match upstream exactly. `hash_hits` / `false_alarms` are oc-native: oc has no `SUM2HASH` bucket table, so it counts zsync bithash-prefilter positives and the subset that fails the exact verify. Documented on `ProbeCounters`. This is a `-vv` debug diagnostic only - no wire, stats, or exit-code impact.

## Restoration note

The local counting machinery landed once before and was silently reverted by an unrelated stale-base merge that clobbered the same files; the multi-source flist ordering lost in that same revert was restored separately. It comes back here routed through the single renderer emitter rather than an engine-side `debug_log`, so there is still only one place that prints the line.

## Known remaining gap, stated plainly

A remote **push** now prints nothing at `DELTASUM` level 1, where upstream prints the `total:` line (on a push the client is the sender, so upstream's sender-side `match_report()` runs locally and reaches the terminal). A **pull** correctly prints nothing, MEASURED.

Closing the push gap needs the probe counters plumbed out of all three scan paths - the sequential scan, the consecutive-match gated scan, and the parallel chunked scan, whose overlapping stripes make a truthful `hash_hits` non-obvious. Shipping it now would mean a line that reads zero on two paths out of three, so it is deliberately left to its own change.

## Tests

- `level_2_local_delta_total_reports_real_match_counters` - a local `--no-whole-file` delta pins the exact MEASURED line, its position between the name list and the trailer, and that it is emitted exactly once. Fails both if a second emitter appears and if the counters regress to hardcoded zeros.
- `deltasum_emissions.rs` - pins `DELTASUM` level 1 to silence in the scanner (VERIFIED to fail when a level-1 emitter is reintroduced) and pins the level-2 per-file line to upstream's exact field order.
- `find_match_window_counted_*` - the counted probe must return the same block index as the uncounted one, so a debug diagnostic can never alter which bytes the delta reuses; and a clean miss must leave the counters untouched.
- `record_delta_probe_accumulates_across_files` - the totals sum across files rather than being overwritten per file, which is what `match.c:433-435` does.

The pre-existing `level_2_local_brackets_name_list_with_delta_and_total` guard was left unchanged: its assertions describe upstream's real whole-file output and still hold.

### DIFF
--- a/crates/cli/src/frontend/execution/drive/summary.rs
+++ b/crates/cli/src/frontend/execution/drive/summary.rs
@@ -159,9 +159,7 @@ where
     let is_local_transfer = !config.is_pull() && !is_sender;
     // upstream: main.c:650-653 forces whole_file=1 for a local transfer unless
     // the user passed --[no-]whole-file; the local-copy default is therefore
-    // whole-file. This decides the `delta-transmission` wording and gates the
-    // `total:` line (whose zero match counts are only provable in whole-file
-    // mode).
+    // whole-file. This decides the `delta-transmission` wording.
     let whole_file = config.whole_file();
     // upstream: generator.c:2291-2294 + match.c:439-446 - the delta-transmission
     // notice fires at DEBUG_GTE(FLIST, 1) and the match_report `total:` line at
@@ -175,7 +173,11 @@ where
     };
     let delta_notice = DeltaTransmissionSummary {
         notice: (is_local_transfer && debug_gte(DebugFlag::Flist, 1)).then_some(delta_state),
-        emit_total: is_local_transfer && whole_file && debug_gte(DebugFlag::Deltasum, 1),
+        // upstream: match_report() (match.c:440) is gated on DEBUG_GTE(DELTASUM, 1)
+        // alone - `--whole-file` does not suppress it, it just leaves the match
+        // counters at zero. MEASURED against rsync 3.4.4: both `-rvv` and
+        // `-rvv --no-whole-file` print the line.
+        emit_total: is_local_transfer && debug_gte(DebugFlag::Deltasum, 1),
     };
     // Capture the preserve-links state before `config` is consumed so the
     // `--list-only` renderer knows whether to append the ` -> <target>` arrow
--- a/crates/cli/src/frontend/progress/render.rs
+++ b/crates/cli/src/frontend/progress/render.rs
@@ -91,9 +91,9 @@ pub(crate) struct DeltaTransmissionSummary {
     /// a local transfer. upstream: generator.c:2291-2294.
     pub notice: Option<DeltaTransmissionState>,
     /// Whether to print the `match_report()` `total:` line after the name list.
-    /// `true` only at `DEBUG_GTE(DELTASUM, 1)` on a whole-file local transfer,
-    /// where every match count is provably zero and `data` is the literal bytes
-    /// copied (read from the summary). upstream: match.c:439-446.
+    /// `true` at `DEBUG_GTE(DELTASUM, 1)` on a local transfer, whole-file or
+    /// not; the counters and `data` are read from the summary.
+    /// upstream: match.c:439-446.
     pub emit_total: bool,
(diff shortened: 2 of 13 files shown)


oferchen:

Verified: output measured against real rsync 3.4.4 and pinned exactly (`total: matches=0  hash_hits=0  false_alarms=0 data=17`, and `matches=1 hash_hits=1` for a 700-byte delta case). CI 62 checks green.

NOT verified: host `oc-verify` did not complete — its session was killed during a host-congestion cleanup. 2 CI failures remain, both `Set up job` (runner infrastructure, not code).

Note: the original task premise ("two emitters") was wrong — there was only ever one production emitter. The real defect was hardcoded zero counters plus suppression under `--no-whole-file`.

PR #7228, opened 2026-08-06
refactor(transfer): one marker-aware NDX reader for every role

oferchen:
## Why

Upstream funnels every NDX read on a live transfer stream through one function,
`read_ndx_and_attrs()` (`rsync.c:322-431`). Its loop classifies the value before
any attribute byte is touched, and the branch order is load-bearing - most
notably `NDX_DEL_STATS` is drained *before* the `!inc_recurse || am_sender` gate,
so a peer that never negotiated INC_RECURSE still consumes the counters instead
of desyncing on them.

oc had four partial re-implementations of that rule, each missing a different
branch:

| Copy | Missing |
|------|---------|
| `receiver/file_list/on_demand.rs` `read_next_frame` | the `NDX_DEL_STATS` branch |
| `receiver/file_list/receive.rs` `receive_extra_file_lists` | `NDX_DONE` and `NDX_DEL_STATS` |
| `receiver/transfer/phases.rs` phase-boundary + goodbye reads | the sub-list branches; goodbye had its own del-stats loop |
| `generator/transfer/goodbye.rs` `read_ndx_skipping_del_stats` | the `rsync.c:343` rejection |

## What

`crates/transfer/src/receiver/ndx_stream.rs` holds the rule once:

- **`read_ndx_step()`** - one iteration of upstream's `read_loop`, in upstream's
  order: file index (`:332`), `NDX_DONE` (`:334`), `NDX_DEL_STATS` (`:336`), the
  `!inc_recurse || am_sender` gate (`:343`), `NDX_FLIST_EOF` (`:353`), sub-list
  segment (`:360`).
- **`read_marker_aware_ndx()`** - the whole loop, for call sites that read no
  attribute tail.
- **`read_ndx_and_attrs()`** - loop plus tail. The attribute read sits *inside*
  the loop because upstream's protocol-29 keep-alive branch (`rsync.c:386-391`)
  jumps back into `read_loop` after `iflags` is decoded; that branch stays
  unimplemented and is documented as a known incompleteness (it is reachable -
  rsync 3.0.9 emits the raw form at exactly protocol 29, `io.c:953-968`, and oc
  negotiates down to 28), with the `continue` seam marked so adding it is an
  insertion rather than a restructuring.

Those branches mutate state the classifier does not own - the receiver's
`file_list` / `ndx_segments` / `flist_eof`, or the generator's deletion counters.
A `FlistMarkerSink` trait inverts the dependency. `ReceiverContext` implements it
by delegating to the existing `receive_one_extra_segment` (segment reception is
reused, not reimplemented); a `NoLazyFlist` value stands in for the call sites
that must never grow a list, and the generator's goodbye borrows itself through
a small adapter that declares the sender role so upstream's `am_sender` term is
what rejects a marker.

`SenderAttrs::read_with_codec_xattr` is split into an NDX half and
`read_attrs_after_ndx()`. That split is what lets `NDX_DONE` stop consuming two
phantom `iflags` bytes the peer never sent (`rsync.c:334-335`).

## Scope

Deliberately **not** in this change: routing the pipelined per-file echo sites,
collapsing the two NDX codecs in the drivers, the protocol-29 keep-alive branch,
the non-regular-file guard (`rsync.c:420-428`), and the xname `sanitize_path()`
pass (`rsync.c:407-417`). The `dir_ndx` range check is not duplicated either -
`receive_one_extra_segment` already performs it (`flist.c:2622-2626`).

Nothing is written on the `NDX_FLIST_EOF` branch. Upstream's
`write_int(f_out, NDX_FLIST_EOF)` (`rsync.c:358`) never reaches a socket: the
sender is rejected by the `:343` gate first, and the receiver child runs with
`sock_f_out = -1; f_out = error_pipe[1]` (`main.c:1063-1067`). Upstream's
`start_flist_forward()` / `stop_flist_forward()` teeing (`io.c:1483-1489`) is
likewise not ported - it exists only because upstream's generator and receiver
are separate processes each parsing their own copy; oc runs both over one
`file_list`.

## Tests

New tests pin the branch *order*, not just behaviour:

- a sub-list marker interleaved between two per-file echoes is consumed and only
  file NDXs surface;
- `NDX_FLIST_EOF` mid-transfer sets `flist_eof` and does not surface;
- **sub-list `dir_ndx` values are non-monotonic and non-contiguous** - a measured
  upstream oracle produced `1, 9, 2, 10, 3, 11` (level-1 dirs interleaved with
  newly discovered subdirs appended to `dir_flist`), so no validation may assume
  ordering or contiguity;
- `NDX_DEL_STATS` is drained by a sink whose `inc_recurse` is off, which only
  passes while `rsync.c:336-342` precedes `rsync.c:343`; the sender drains it too
  and its counters reach the sink;
- `NDX_DONE` returns without consuming the attribute tail (asserted on the reader
  position, so the regression is visible even though the returned value looks
  correct);
- a sender whose `inc_recurse` is genuinely on still rejects every marker with
  upstream's wording, `Invalid file index: %d (%d - %d)`, tagged `RERR_PROTOCOL`.

The byte-for-byte upstream 3.4.4 daemon capture
(`real_upstream_multisegment_sublist_decodes_as_segments`) passes unchanged as
the no-regression gate on the classifier.

### DIFF
--- a/crates/transfer/src/generator/transfer/goodbye.rs
+++ b/crates/transfer/src/generator/transfer/goodbye.rs
@@ -14,7 +14,54 @@ use protocol::codec::{MonotonicNdxWriter, NDX_DEL_STATS, NDX_DONE, NdxCodec};
 use protocol::stats::DeleteStats;
 
 use super::super::{GeneratorContext, is_early_close_error};
+use crate::receiver::ndx_stream::{FlistMarkerSink, NdxFrame, StreamRole, read_marker_aware_ndx};
 use crate::role_trailer::error_location;
+
+/// The sender's view of the goodbye NDX stream.
+///
+/// Borrows the generator so `NDX_DEL_STATS` lands in its counters, and reports
+/// [`StreamRole::Sender`] so upstream's `am_sender` term (`rsync.c:343`) is what
+/// rejects any file-list marker. `inc_recurse` is reported truthfully rather
+/// than forced off, so the rejection is attributable to the role and not to a
+/// capability the transfer actually negotiated.
+///
+/// The primitive is defined under `receiver` because that is where the lazy
+/// file-list state it inverts lives; it holds no receiver-only types, and the
+/// generator already reaches across for shared wire types (see
+/// `generator/protocol_io.rs`'s use of `receiver::SumHead`).
+struct GoodbyeNdxSink<'a>(&'a mut GeneratorContext);
+
+impl FlistMarkerSink for GoodbyeNdxSink<'_> {
+    type FrameMark = ();
+
+    fn role(&self) -> StreamRole {
+        StreamRole::Sender
+    }
+
+    fn last_file_ndx(&self) -> i32 {
+        // upstream: rsync.c:345-348 - the last index of the newest flist, or -1
+        // when nothing has been sent.
+        self.0.file_list().len() as i32 - 1
+    }
+
+    fn begin_frame(&mut self) {}
+
+    fn on_del_stats(&mut self, stats: &DeleteStats) -> io::Result<()> {
+        // upstream: main.c:238-247 read_del_stats() adds to the global counters.
+        self.0.accumulate_delete_stats(stats);
+        debug_log!(
+            Flist,
+            2,
+            "consumed NDX_DEL_STATS during goodbye: {} deletions",
+            stats.total()
+        );
+        Ok(())
+    }
+
+    fn inc_recurse(&self) -> bool {
+        self.0.inc_recurse()
+    }
+}
 
 impl GeneratorContext {
     /// Handles the goodbye handshake at end of transfer.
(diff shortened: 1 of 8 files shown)


oferchen:

Verified: Linux interop passed. All 8 decomposed sub-tasks covered (primitive 471 lines + 466 lines of tests; `read_attrs_after_ndx` split; three duplicate copies of upstream's `read_ndx_and_attrs` loop collapsed onto one implementation). CI 45 checks green.

NOT verified: nextest (stable), Linux musl, and Windows (stable) — the last is blocked on a `Set up job` infra rerun. Byte-neutrality evidence (daemon pull/push + proto29 diffed against master) was NOT produced.

Risk note: this touches the receiver NDX read path that all four network cells route through. The byte-neutrality diff is the gate that matters here, and it is outstanding.

PR #7229, opened 2026-08-06
perf(transfer): own the in-flight FileEntry window in the receiver pipeline

oferchen:
Pure borrow-model refactor of the receiver pipeline. No behaviour change, no wire change.

`run_pipeline_loop_decoupled` took `&'a self` and held `&'a FileEntry` borrows into
`self.file_list` for the whole loop - both the input `files_to_transfer` and the
in-flight `pending_files_info` window. While those borrows are live `self.file_list`
cannot grow, which blocks any future work that needs to append to it mid-loop.

This changes the signature to `&mut self`, drops the `&'a FileEntry` from the input
tuple (`Vec<(usize, PathBuf, u32)>`), and lets the in-flight window own its
`FileEntry` - cloned as it enters the window and bounded by `pipeline.available_slots()`,
so the clone count is O(window), not O(total).

What is deliberately unchanged: the `no_transfer_rows` merge order, the flush and
keepalive cadence, the NDX write order, and the phase-2 redo codec threading
(one `ndx_write_codec`/`ndx_read_codec` pair through both the phase-1 and redo
calls - upstream keeps a single connection-wide `prev_positive`/`prev_negative`,
`io.c:2245`, and the redo re-requests through that same state).

## Verification

`cargo fmt --all -- --check` and `cargo check --workspace --all-features` are clean on
x86_64 Linux, and `cargo nextest run --workspace --all-features` runs there against this
branch.

Byte-neutrality was re-checked against a release binary built from the merge base
(`8c68fdb46`) over six daemon-pull scenarios: fresh and delta-seeded pulls on protocol 32
and protocol 29, plus an `--append-verify --ignore-times` run on each protocol that forces
the phase-2 verification redo. Both sides of each transfer used the binary under test. For
every scenario the destination tree (per-file md5, type, mode, size, mtime, symlink target)
and the normalized client stdout - itemize rows plus the whole `--stats` block, including
`Total bytes sent` / `Total bytes received` - are identical. Only the `bytes/sec` figure is
normalized.

```
===== scenario p32  (args: ) =====                                   IDENTICAL
===== scenario p32d (args: ) =====                                   IDENTICAL
===== scenario p29  (--protocol=29) =====                            IDENTICAL
===== scenario p29d (--protocol=29) =====                            IDENTICAL
===== scenario redo32 (--append-verify --ignore-times) =====         IDENTICAL
===== scenario redo29 (--protocol=29 --append-verify --ignore-times) IDENTICAL
BYTE_NEUTRALITY_FAIL=0
```

`daemon_pull_forced_verification_failure_recovers_via_redo` - the regression guard for the
redo NDX codec threading - passes on this branch (1 passed).

Known follow-up, filed separately: `build_files_to_transfer` still yields the
4-tuple and the caller discards the entry. Changing the producer would have widened
this diff, which is gated on byte-neutrality, so the shim stays for now.

### DIFF
--- a/crates/transfer/src/receiver/tests/push_metadata_itemize.rs
+++ b/crates/transfer/src/receiver/tests/push_metadata_itemize.rs
@@ -103,7 +103,7 @@ fn server_push_pipeline_forwards_metadata_only_record_and_drains_echo() {
     let sent = SharedBuf::default();
     let mut writer = crate::writer::ServerWriter::new_plain(sent.clone());
     let mut metadata_errors = Vec::new();
-    let files: Vec<(usize, &FileEntry, PathBuf, u32)> = Vec::new();
+    let files: Vec<(usize, PathBuf, u32)> = Vec::new();
     let mut ndx_write_codec = protocol::codec::MonotonicNdxWriter::new(ctx.protocol.as_u8());
     let mut ndx_read_codec = protocol::codec::create_ndx_codec(ctx.protocol.as_u8());
     let result = ctx
--- a/crates/transfer/src/receiver/transfer/pipeline.rs
+++ b/crates/transfer/src/receiver/transfer/pipeline.rs
@@ -97,16 +97,15 @@ impl ReceiverContext {
     /// `(files_transferred, bytes, literal, matched, redo_indices, delayed_updates)`.
     #[allow(clippy::too_many_arguments)]
     pub(in crate::receiver) fn run_pipeline_loop_decoupled<
-        'a,
         R: Read,
         W: Write + crate::writer::MsgInfoSender + ?Sized,
     >(
-        &'a self,
+        &mut self,
         reader: &mut crate::reader::ServerReader<R>,
         writer: &mut W,
         pipeline_config: PipelineConfig,
         setup: &PipelineSetup,
-        files_to_transfer: Vec<(usize, &'a FileEntry, PathBuf, u32)>,
+        files_to_transfer: Vec<(usize, PathBuf, u32)>,
         metadata_errors: &mut Vec<(PathBuf, String)>,
         is_redo_pass: bool,
         total_files: usize,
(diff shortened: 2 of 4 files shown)


oferchen:

Verified: diff reviewed line-by-line. The `no_transfer_rows` merge is textually unchanged apart from dropping the entry element; the clone is bounded by `pipeline.available_slots()` (O(window), not O(total)); the phase-2 redo NDX codec threading is untouched, so the #7221 fix survives. Rebased onto current master (39fe17d6c).

NOT verified: **the byte-neutrality evidence — which is this PR's entire justification — was never produced.** A pure borrow-model refactor is only safe if the wire is provably unchanged; CI green proves tests pass, not that. The host run was dispatched but its result is unknown.

Required before this leaves draft: proto32 daemon pull AND a proto29 transfer, pre-change vs post-change binaries, destination tree and client stdout diffed (normalize only bytes/sec). If the diff is non-empty, this PR is wrong regardless of CI.

Known follow-up: `build_files_to_transfer` still yields the 4-tuple and the caller discards the entry — a deliberate shim to keep this diff minimal, tracked separately.

oferchen:
## Byte-neutrality gate: **PASSED**

Measured on aarch64 Linux, default features (so `transfer/incremental-flist` is ON and the
incremental driver this PR modifies genuinely executes — a build with it off would make the
test vacuous).

Binaries: `origin/master` @ `39fe17d6c` (PRE) vs this branch @ `f01596f7b` (POST).
Workload: 8 dirs x nested `sub/` x 25 files = 200 files, daemon pull, run through each binary.

| cell | destination tree | client stdout |
|---|---|---|
| proto32 daemon pull | **IDENTICAL** (200 files) | **IDENTICAL** |
| proto29 daemon pull | **IDENTICAL** (200 files) | **IDENTICAL** |

All four transfers exited 0. stdout compared with only `bytes/sec` and `speedup` normalized —
every other line, including the full `--stats` block, matched byte for byte.

This is the evidence the PR was waiting on: a pure borrow-model refactor is only safe if the
observable output is provably unchanged, and CI green proves tests pass rather than that.
Taking it out of draft.

PR #7230, opened 2026-08-06
fix(receiver): print client-side transfer diagnostics locally

oferchen:
Started as "verify the phase-2 verification-redo recovers on the sibling network
paths" and uncovered a production bug that made a remote-shell pull **deadlock and
leave the destination corrupt**.

## The bug

A client receiver queued its per-file diagnostics - the phase-1 "failed
verification" warning and per-file transfer errors - and pushed every one of them
to the peer as a multiplexed frame. Upstream frames a diagnostic only when
`am_server` (log.c:330-346); a client writes it to its own stdout/stderr.

Framing it from a client receiver put the text in front of a `--server --sender`
whose stdout IS the wire. The peer rendered the payload raw into the multiplexed
stream, the client's demux desynced, its receive loop returned before consuming the
phase-2 redo resend, and the run deadlocked - the client blocked in `wait()` on a
child blocked writing a resend nobody was reading.

Over a daemon transport the same frame was swallowed instead. That is why the
daemon cells recovered and the remote-shell pull hung forever: a fix verified on
one transport did not cover its siblings.

## The fix

Route the queued diagnostics through the receiver's existing client-vs-server emit
seam (`emit_info_line` / `emit_error_line` / `emit_error_xfer_line`) and add the
missing `emit_warning_line`, so a server receiver frames MSG_WARNING (downgraded to
MSG_INFO below protocol 30, log.c:332-337) and a client writes to stderr. The queued
text gains the trailing newline upstream's `rprintf` format carries.

## Oracle

rsync 3.4.4 (protocol 32) was measured on all four network cells plus a local
control with this fixture. Every cell exits 0, ends byte-correct, and prints
`WARNING: <name> failed verification -- update retained (will try again).` to
stderr.

The daemon-pull redo regression test is generalized into all four cells, each run
under a deadline so a transport deadlock fails the test instead of hanging the suite.


oferchen:
Marking draft until verification completes: the four network cells still need an empirical re-run against the patched binary, plus oc-verify. The root cause and fix are strace-proven on both sides, but the post-fix measurement is outstanding.

oferchen:
## Four-cell validation: **core fix works, two output-fidelity gaps found**

Measured on aarch64 Linux against `061e87959`. Fixture: 200 KiB source, 100 KiB wrong zero
prefix, `--append-verify --ignore-times` (forces a phase-2 verification redo). SSH cells use a
local-shell stub, which the original investigation confirmed reproduces identically to real ssh.

### The deadlock fix is confirmed working

| cell | exit | destination byte-correct |
|---|---|---|
| daemon pull | 0 | YES |
| daemon push | 0 | YES |
| ssh pull | 0 | YES |
| ssh push | 0 | YES |

`--debug=all` on the ssh-push cell shows the redo executing as intended —
`recv_files(payload.bin)` appears twice (phase-1 attempt, then phase-2 redo),
`generate_files phase=2` runs, and the file lands correct. **No hang on any cell.**

### Two divergences from the upstream oracle remain

Oracle (rsync 3.4.4, measured on all five cells):
`WARNING: payload.bin failed verification -- update retained (will try again).`
— bare relative name, stderr, default verbosity.

| cell | default verbosity | name emitted |
|---|---|---|
| daemon pull | prints | `/abs/path/to/dst/payload.bin` — **absolute, should be bare** |
| daemon push | prints | absolute — same |
| ssh pull | prints | absolute — same |
| ssh push | **silent** — needs `-v`; `--debug=all` does not surface it either | `payload.bin` — **correct** |

Inverted: the framed server-receiver → client path gets the *name* right and the *verbosity*
wrong; the three local-emit paths get the verbosity right and the name wrong.

Tracked as #348 (absolute name) and #349 (ssh-push verbosity gate). Both are output fidelity,
not correctness — no transfer is wrong and nothing hangs.

**One consequence for this PR:** its four-cell test passes on ssh push, so it is not asserting
the warning at default verbosity with the exact upstream text. That assertion should be
tightened, or these two gaps stay invisible to CI.

Staying in draft.

oferchen:
## Correction to my previous comment — I had one of the two findings backwards

I reported that ssh push was defective for being silent at default verbosity. **That was wrong.**
I had propagated an unverified oracle claim without measuring the verbosity condition myself.
Measuring upstream directly reverses it.

### Upstream's actual rule — `receiver.c:1071-1091`

```c
enum logcode msgtype = redoing ? FERROR_XFER : FWARNING;
if (msgtype == FERROR_XFER || INFO_GTE(NAME, 1) || stdout_format_has_i) {
    ...
    rprintf(msgtype, "%s: %s failed verification -- update %s%s.\n",
            errstr, local_name ? f_name(file, NULL) : fname, keptstr, redostr);
}
```

A **phase-1** failure is `FWARNING`, printed only under `-v` or `-i`. A **phase-2** (`redoing`)
failure is `FERROR_XFER` and always prints. This fixture's redo succeeds, so only the phase-1
warning occurs — and upstream suppresses it at default verbosity.

Measured, rsync 3.4.4, local, same fixture:

```
-a  (default)   exit=0   stderr = EMPTY
-av             exit=0   stderr = WARNING: payload.bin failed verification -- update retained (will try again).
-a -i           exit=0   stderr = WARNING: payload.bin failed verification -- update retained (will try again).
```

### Corrected verdict

**oc's ssh-push path is CORRECT** — silent at default, bare `payload.bin` under `-v`. It matches
upstream exactly and is the reference the other cells should copy.

**The other three cells have two defects**, measured on daemon pull at every verbosity:

```
-a      WARNING: /tmp/.../dst/payload.bin failed verification ...
-av     WARNING: /tmp/.../dst/payload.bin ...     <- still absolute
-a -i   WARNING: /tmp/.../dst/payload.bin ...     <- still absolute
```

1. **Missing the verbosity gate** — prints at default where upstream is silent.
2. **Wrong name** — the joined absolute destination path instead of the flist-relative name.
   Upstream `change_dir()`s into the destination root, so relative is the correct rendering
   (visible as `[Receiver] change_dir(...)` in a `--debug=all` trace).

Also worth mirroring: a phase-2 failure must read `ERROR: <name> failed verification -- update
retained.` with **no** `(will try again)` suffix.

### Unchanged

The deadlock fix in this PR is confirmed working on all four cells — exit 0, byte-correct, no
hang, `--debug=all` shows the redo executing (`recv_files(payload.bin)` twice, `generate_files
phase=2`). This correction is about the diagnostic only.

**One consequence:** this PR's four-cell test passes on all four cells despite the above, so it
is asserting a substring rather than the oracle. It should assert EMPTY stderr at default
verbosity and the exact bare-name line under `-v` — otherwise these stay invisible.

Tracked as #348 (rewritten). #349 closed as invalid.

oferchen:
## ⚠ This PR introduces a regression — please do not merge as-is

An A/B against master with an identical fixture and command, differing only in the binary
(daemon pull, `-a --append-verify --ignore-times`, default verbosity):

```
origin/master   exit=0   stderr = EMPTY
this branch     exit=0   stderr = WARNING: /tmp/.../dst/payload.bin failed verification -- update retained (will try again).
```

Master is **correct** here. Upstream 3.4.4 is also silent at default verbosity, measured:

```
-a  (default)   stderr = EMPTY
-av             WARNING: payload.bin failed verification -- update retained (will try again).
-a -i           WARNING: payload.bin failed verification -- update retained (will try again).
```

That is `receiver.c:1071-1072`:

```c
enum logcode msgtype = redoing ? FERROR_XFER : FWARNING;
if (msgtype == FERROR_XFER || INFO_GTE(NAME, 1) || stdout_format_has_i) {
```

A phase-1 failure is `FWARNING`, printed only under `-v` or `-i`. This fixture's redo succeeds,
so it never escalates to the ungated `FERROR_XFER`.

### Two defects in the new emit path

Routing the queued diagnostics through `emit_warning_line` was the right idea, but the emit
site drops two of upstream's rules:

1. **No verbosity gate** — fires at default verbosity where upstream and master are both silent.
2. **Wrong name** — the joined absolute destination path instead of the flist-relative name.
   Upstream `change_dir()`s into the destination root, so relative is the correct rendering
   (`[Receiver] change_dir(...)` is visible under `--debug=all`).

Also worth mirroring while fixing: a phase-2 failure reads
`ERROR: <name> failed verification -- update retained.` with **no** `(will try again)` suffix.

### What is unaffected

**The deadlock fix is sound and confirmed.** All four cells exit 0 and end byte-correct, and
`--debug=all` shows the redo executing (`recv_files(payload.bin)` twice, `generate_files
phase=2`). The client-vs-server framing analysis in the original commit message is correct and
worth keeping — it fixed a real hang and destination corruption on ssh pull. Only the emit-site
detail is wrong.

### Why CI did not catch this

This PR's four-cell test asserts a substring of the warning at whatever verbosity it runs, so it
passes despite emitting where upstream is silent. It should assert the oracle instead: **empty
stderr at default verbosity**, and the exact bare-name line under `-v`.

Tracked as #348. A `verify-redo` check that encodes exactly this oracle across the transport
matrix is in #7232.

oferchen:
Superseded by #7236, which already contains both commits from this branch
(`87503f160` and `061e87959` — verified with `git merge-base --is-ancestor`).

Closing rather than stacking: #7236 targets master directly, and a stacked PR
here would reintroduce the base-retargeting hazard this repo has already been
bitten by. Consolidating keeps one reviewable unit and one merge.

For the record, the concern raised against this branch during review is
resolved. It was reported as introducing an ungated warning at default
verbosity. Per-stream measurement on both branches (stdout and stderr captured
separately, five cells, three verbosities) showed master already emitted that
line at every verbosity — on stdout, with an absolute path — via MSG_INFO being
re-framed by the peer. This branch changed only the stream (stdout to stderr,
matching upstream's `rwrite()` mapping of FWARNING at log.c:314-316) and a
missing newline. Both correct. It made a pre-existing defect visible rather
than creating one; the missing verbosity gate and the absolute name are fixed
in #7236.

PR #7238, opened 2026-08-06
fix(matching): match the basis short final block on the sender

oferchen:
## What

oc's sender never matched the basis's short final block, so that block was re-sent as literal data on every transfer. Measured against upstream rsync 3.4.4 over a daemon push (`--no-whole-file --block-size=700`), 204800-byte basis = 292 x 700 + 400, same upstream daemon as receiver in both runs so the sender is the only variable:

| sender | Literal | Matched |
|---|---|---|
| rsync 3.4.4 | 102,200 | 102,600 |
| oc-rsync (before) | 102,600 | 102,200 |

The 400-byte gap is exactly the short final block. This is not a redo defect: it reproduces with `--append`, `--append-verify`, phase 2 and MSG_REDO all absent, and identically with `--inplace`.

## Upstream rule

`hash_search()` reaches the short block through three cooperating pieces:

- `match.c:174` - `end = len + 1 - s->sums[s->count-1].len;` the scan bound comes from the LAST block's length, not from `blength`, so the scan continues past `len - blength`.
- `match.c:321-331` - `more = offset + k < len;` ... `else --k;` the rolling window shrinks once there is nothing left to feed it.
- `match.c:222-224` - `l = (int32)MIN((OFF_T)s->blength, len-offset); if (l != s->sums[i].len) continue;` a candidate is accepted only at a matching length.

Together those admit exactly **one** short match, at offset `len - tail_len`: every other shrunk window has a length no basis block carries. The EOF drain now probes that single position via the existing `find_tail_match()` primitive - the same one the local-copy path already used - so the rule has one implementation, not two.

## The coalescing hazard

The short block is emitted as its **own** `Copy` carrying its **true** length, never folded into a seq-match run. `flush_seq_match_run` derives a run as `run_len * block_len`; absorbing a shorter block would overstate the run, over-run the source cursor the wire layer feeds to `see_deflate_token`, and desync the receiver's compressed-token dictionary. `short_final_block_copy_is_standalone_and_carries_its_true_length` pins the exact token stream, because a test on byte totals alone would pass with a coalesced tail.

## Two adjacent corrections

**A false citation.** The in-code justification for declining the tail cited `token.c:665` as the upstream receiver rejecting short-block tokens. That line is `rprintf(FERROR, "inflate returned %d (%d bytes)\n", r, n);` - a generic inflate failure. Upstream sends such tokens routinely. Replaced with the three citations above.

**A basis smaller than one block.** `from_signature()` refused to build an index when no block was full-length, so such a file was sent entirely as literal. Upstream runs `build_hash_table()` + `hash_search()` for any `s->count > 0` (`match.c:match_sums()`) and matches the lone block. Measured with a 400-byte basis, `-I --block-size=700`: upstream Matched 400 / Literal 0, oc Matched 0 / Literal 400. Only an empty signature now returns `None`.

This one is also a **prerequisite**, not just an improvement: once the sender honours the wire remainder, a single-short-block basis has zero full blocks, and the old `None` would turn `build_signature_index()` into a hard error. It must land with or before that change.

## Scope guards

- **Striped scan.** A stripe's EOF is not the file's EOF. Unguarded, a non-final stripe would place the short final block mid-file, which upstream's length test never permits. A `tail_match` parameter restricts the probe to a stripe that runs to the source end.
- **Consecutive-match extension** (`OC_CONSECUTIVE_MATCH`, opt-in on both peers) keeps sending the tail as a literal. That path halves the strong sum and only trusts a match a neighbouring block corroborates; the final block has no successor to corroborate it, so admitting it would mean a second, weaker acceptance rule on the one path with half-length checksums. `consecutive_match_gate_leaves_the_short_final_block_literal` pins that choice.

## Known limitation - this is DRAFT for this reason

The fix is correct but **not yet observable on the wire**, because a second, independent root cause sits in `crates/transfer`, which is fenced to another task:

```
crates/transfer/src/generator/delta.rs:467-470   build_signature_index()
    SignatureLayout::from_raw_parts(block_length_nz, 0 /* remainder unknown from wire format */, ...)
```

The remainder is **not** unknown - it is on the wire (`io.c:2061` `sum->remainder = read_int(f)`), oc's `SumHead` parses it, and upstream applies it on the sender in `receive_sums()`:

```c
sender.c:109-110   if (i == s->count-1 && s->remainder != 0) s->sums[i].len = s->remainder;
```

oc drops it, so every reconstructed block claims `block_length` and the probe never fires. The local-copy path is unaffected because it builds its index from a real `FileSignature` carrying the true length - which is precisely why the tail matched locally and not on the wire. Tracked separately; the numbers above will not move until it lands.

## Tests

New `crates/matching/tests/short_final_block_match.rs`:

- `basis_signature_ends_in_a_short_block` - pins the fixture premise, so the other assertions cannot go vacuous
- `short_final_block_is_matched_not_resent_as_literal` - the regression
- `short_final_block_copy_is_standalone_and_carries_its_true_length` - the coalescing hazard, on token identity
- `whole_file_shorter_than_one_block_matches_as_a_single_copy`
- `a_tail_shorter_than_the_final_block_stays_literal` - the tail position must be reachable
- `striped_scan_places_the_short_block_only_at_the_true_tail` - plants the tail block's content mid-source
- `consecutive_match_gate_leaves_the_short_final_block_literal`

The fixture uses a 64-bit LCG rather than a short-cycle pattern: with a repeating period, an unaligned window near EOF matches an earlier full block by content and swallows the trailing bytes, so "the tail was not matched" would be a statement about the data instead of the matcher. That is how the first draft of these tests failed.

Three existing tests encoded the pre-fix behaviour and were corrected, not deleted - their intent is preserved and now expressed against the block count the signature actually carries.

## Verification

- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` clean, no waivers
- `cargo nextest run -p matching -p signature -p engine --all-features --no-fail-fast` - 5632 passed
- `cargo nextest run -p protocol -p transfer --all-features --no-fail-fast` - 8713 passed
- No golden wire byte changed. Note honestly *why*: the wire-reconstructed index cannot yet carry a short final block, so the goldens are currently insensitive to this path. They must be re-run and reasoned about when the remainder fix lands.
- Interop against real upstream rsync 3.4.4 is the measurement table above; the full harness run is still outstanding.

### DIFF
--- a/crates/matching/src/generator.rs
+++ b/crates/matching/src/generator.rs
@@ -741,6 +741,12 @@ impl DeltaGenerator {
         // position - the source's trailing `tail_len` bytes - and let everything
         // ahead of it drain as literals, which is what upstream's scan does over
         // those offsets anyway.
+        //
+        // That derivation is also what keeps the cost honest. `find_tail_match`
+        // is O(block_count): a short block cannot live in the BitHash or the
+        // compact lookup, which are keyed on full-block rolling sums, so it has
+        // to be found by scan. This block sits OUTSIDE the scan loop and so runs
+        // at most once per call; probing per window would be quadratic.
         if tail_match {
             if let Some(tail_len) = short_final_block_len(index, block_len) {
                 if window.len() >= tail_len {
--- a/crates/matching/src/generator.rs
+++ b/crates/matching/src/generator.rs
@@ -11,7 +11,7 @@
 
 use std::io::{self, Cursor, Read};
 
-use checksums::RollingChecksum;
+use checksums::{RollingChecksum, RollingDigest};
 use logging::debug_log;
 use rayon::prelude::*;
 
@@ -66,6 +66,22 @@ fn flush_seq_match_run(
         }
     }
     *run_len = 0;
+}
+
+/// Length of the basis's final block when it is shorter than a full block.
+///
+/// The signature layout only ever leaves the *last* block short (every earlier
+/// block spans a full `block_len`), which is why upstream derives its scan
+/// bound from that one block: `end = len + 1 - s->sums[s->count-1].len`
+/// (`match.c:174`). Returns `None` when the basis divides evenly into full
+/// blocks, in which case there is no short tail to match.
+fn short_final_block_len(index: &DeltaSignatureIndex, block_len: usize) -> Option<usize> {
+    let count = index.block_count();
+    if count == 0 {
+        return None;
+    }
+    let len = index.block(count - 1).len();
+    (len > 0 && len < block_len).then_some(len)
 }
 
 /// The single source of truth for the literal-flush cadence.
@@ -372,7 +388,9 @@ impl DeltaGenerator {
         let prune_matched = self.prune_matched;
         #[cfg(not(any(test, feature = "bench-internal")))]
         let prune_matched = true;
-        self.generate_with_prune(reader, index, prune_matched)
+        // `reader` is the whole source, so its EOF is the file's EOF and the
+        // short-final-block probe is in scope.
+        self.generate_with_prune(reader, index, prune_matched, true)
     }
 
     /// Reports whether emitting a `Copy` for basis block `idx` is safe at the
@@ -407,11 +425,17 @@ impl DeltaGenerator {
     /// only read-only lookups on `index` (no [`DeltaSignatureIndex::mark_consumed`]
     /// or [`DeltaSignatureIndex::reset_consumed`]), so a shared `&index` is safe
     /// to scan from multiple rayon workers concurrently.
+    ///
+    /// `tail_match` states whether `reader`'s EOF is the source file's EOF. Only
+    /// then may the short-final-block probe run: upstream reaches that block
+    /// solely at offset `len - s->sums[s->count-1].len` (`match.c:174`), so a
+    /// stripe that merely stops early must not claim it mid-file.
     fn generate_with_prune<R: Read>(
         &self,
         mut reader: R,
         index: &DeltaSignatureIndex,
         prune_matched: bool,
+        tail_match: bool,
     ) -> io::Result<DeltaScript> {
         let block_len = index.block_length();
         let mut window = RingBuffer::with_capacity(block_len);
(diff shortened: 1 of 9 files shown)


oferchen:
## Per-cell measurement, branch vs master vs upstream 3.4.4

All on aarch64 Linux, `--no-whole-file --block-size=700`, upstream 3.4.4 as the daemon/remote in every network cell so the **sender** is the only variable. Every destination verified byte-identical to the source (`cmp`).

**Fixture A** - 204800-byte basis = 292 x 700 + 400, source = basis with the first byte of every odd full block flipped, tail intact. Values are Literal / Matched:

| cell | upstream 3.4.4 | oc master | oc this branch |
|---|---|---|---|
| local | 102,200 / 102,600 | 102,600 / 102,200 | 102,600 / 102,200 |
| daemon push | 102,200 / 102,600 | 102,600 / 102,200 | 102,600 / 102,200 |
| ssh push | 102,200 / 102,600 | 102,600 / 102,200 | 102,600 / 102,200 |
| ssh pull | 102,200 / 102,600 | 102,200 / 102,600 | 102,200 / 102,600 |

Unchanged on the wire, as stated in the description - the sum_head remainder is still dropped, so the reconstructed index has no short block for the probe to find.

Two things this table settles that assumption would have got wrong:

- **The ssh-pull column is not evidence of anything.** It already reads at parity on **master**. On a pull the client is the receiver and reports against its own locally generated signature, which carries the true 400-byte length. That cell cannot distinguish sender behaviour and must not be used as an acceptance gate for this bug.
- **The local path has the same defect**, contrary to the assumption that it was already correct. It matches the tail only when a match happens to end exactly `tail_len` bytes before EOF; the residual window is then short and `find_tail_match_window` accepts it. When the scan reaches EOF with a *full* window that helper rejects it for being full-length - the identical shape of the bug fixed here, living in `crates/engine`'s own scan loop. Same fixture but with the source identical to the basis, so the last match ends exactly 400 bytes early: upstream 0 / 204,800, master 0 / 204,800, branch 0 / 204,800. All match, which is why the defect stayed hidden.

**Fixture B** - 400-byte basis (one short block, no full block), `-I`, source identical to basis. This is the half of the fix that **is** observable today, and it reaches upstream parity:

| | Literal | Matched |
|---|---|---|
| upstream 3.4.4 | 0 | 400 |
| oc master | 400 | 0 |
| oc this branch | **0** | **400** |

Master could not do better: `from_signature()` returned `None` for a basis with no full-length block, so the file was sent whole.

## Test runs

- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` clean, no waivers
- `cargo nextest run -p matching -p signature -p engine --all-features --no-fail-fast` - 5632 passed, 29 skipped
- `cargo nextest run -p protocol -p transfer --all-features --no-fail-fast` - 8713 passed, 5 skipped

The full `tools/ci/run_interop.sh` harness was **not** run: it builds four rsync versions and the shared host was at 92% disk with other work in flight, so running it risked breaking unrelated jobs. The interop evidence above is against a real upstream 3.4.4 binary in all four network directions plus local, which is more targeted than the harness for this change; the harness still needs to run in CI.

oferchen:
Heads-up on an overlap: #7239 fixes the same defect and reaches the same design here in `crates/matching` (EOF tail probe, stripe-seam gating, index constructor accepting a sub-block basis) - arrived at independently, which is a good sign for the approach.

One measured caveat before this lands: built from this branch's head (f979ec9fb), oc as sender against an upstream 3.4.4 daemon receiver still reports **literal 103,300 / matched 101,500**, byte-identical to master. Upstream reports 102,900 / 101,900.

The reason is in `crates/transfer`, which this PR does not touch: `build_signature_index` sets the layout remainder to 0 ("remainder unknown from wire format") and gives every block `block_length`. It is not unknown - it is the fourth sum-head field (`io.c:write_sum_head()`), and upstream's sender applies it to the last block (`sender.c:109-112`). With every block claiming `block_length`, `find_tail_match` - which requires `block.len() == tail_len` - can never fire on a real transfer. The unit tests here pass because they build the index from a real `FileSignature`, which does carry true per-block lengths; the wire path never does.

I hit exactly this trap first - my matching-only build measured no change at all.

Either land #7239, or take its `crates/transfer` half into this PR and close that one; no preference. Worth avoiding is the matching-only version, which reads as a fix and measures as a no-op.

oferchen:
## Both halves landed - the wire numbers now move

Second commit adds the `crates/transfer` half (task #360). The two are one fix: neither is verifiable alone.

### Measured, upstream as the peer, sender is the only variable

204800-byte basis = 292 x 700 + 400, source = basis with the first byte of every odd full block flipped, tail intact. Values are Literal / Matched. Every destination `cmp`-verified.

| cell | oc before | oc after | upstream 3.4.4 |
|---|---|---|---|
| daemon push | 102,600 / 102,200 | **102,200 / 102,600** | 102,200 / 102,600 |
| ssh push | 102,600 / 102,200 | **102,200 / 102,600** | 102,200 / 102,600 |

Same result against **upstream 3.4.1** as the peer (daemon push and ssh push, both 102,200 / 102,600, DEST-OK), so this is not tied to one upstream build.

### The `remainder == 0` guard the review asked for

Basis of 293 x 700 = 205100, an exact multiple, so the wire `remainder` is 0 and upstream's `&& s->remainder != 0` must keep `blength` on every block including the last:

| | Literal | Matched |
|---|---|---|
| upstream 3.4.4 | 102,200 | 102,900 |
| oc, both halves | 102,200 | 102,900 |

Unaffected, as required. A unit test pins the block lengths directly (`zero_remainder_keeps_blength_on_every_block`) so this cannot regress silently.

### Sub-block-size basis (400 bytes, one short block, no full block), local

upstream 0 / 400, oc before 400 / 0, **oc after 0 / 400**.

### Consecutive-match path, genuinely exercised

oc-to-oc daemon push with `OC_CONSECUTIVE_MATCH=1` exported to **both** the daemon and the client - the numbers change (204,800 / 0 vs 102,200 / 102,600 with it unset), which is how I know the gated path actually engaged rather than the env being ignored. Destination byte-correct either way. Everything demotes to literal on this fixture because alternating blocks are corrupted, so no match has a matching neighbour - the documented behaviour of that path, unchanged by this PR (the diff there is comment-only).

### Still out of scope, unchanged

Local copy stays at 102,600 / 102,200. That is `crates/engine`'s own scan loop, tracked separately as #365 - same upstream rule, different code.

## The four zsync interactions

- `find_tail_match` deliberately bypasses the BitHash, tag table and compact lookup. The rationale is now **in the code**: all three are keyed on rolling sums over `block_length` bytes, a short block's sum spans `tail_len` bytes, and `populate_index` never inserts a partial block - so probing them would be a category error, not a slower path. It does honour both identity prunes (`MatchedBlocks`, `consumed`).
- **Call count: once per file.** The probe sits outside the scan loop, and the derivation (upstream's shrinking window admits exactly one short-match offset) is what licenses probing one position instead of scanning. Both stated in the code.
- **O(1) short-circuit rejected, with the reason recorded.** `find_tail_match` is public API over a `FileSignature`, and `SignatureBlock::from_raw_parts` accepts any length in any order, so "only the last block can be short" is an invariant of the two *builders*, not of the type. A last-block-only check would be an unsound assumption. Left as the linear scan; the body is one length comparison per block.
- **No conflict with #272.** My diff to `generate_gated` is comment-only (verified). Separately worth flagging: the current `generate_gated` still has **no basis-index adjacency check** - `run.push(basis_idx)` is unconditional, so runs are consecutive in source offset only. If #272 was meant to add `id+1` adjacency, it is not in the tree.

## Verification

- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` clean, no waivers
- `cargo nextest run --all-features --no-fail-fast`, per crate: matching 366, signature 234, engine 5032, protocol 6029, transfer 2687 - **14,348 passed, 0 failed**
- Goldens: no wire byte changed. The protocol suite (6029) includes them. Worth stating plainly - unlike the first commit, this run **is** meaningful, because the reconstructed index can now carry a short final block, so the goldens were exercised against the changed path rather than being insensitive to it.

**Interop harness is BLOCKED on this host, and not by this change.** `tools/ci/run_interop.sh` aborts building rsync **2.6.9** (2006 vintage) on modern glibc:

```
syscall.c:296:17: error: conflicting types for 'lseek64'; have 'off64_t(void)'
/usr/include/unistd.h:350:18: note: previous declaration ... '__off64_t(int, __off64_t, int)'
make: *** [Makefile:59: syscall.o] Error 1
```

2.6.9 is in `extra_build_versions` - built and cached only, no scenarios wired to it - but under `set -euo pipefail` its failure aborts the run before **any** upstream version is installed, so 3.0.9 / 3.1.3 / 3.4.4 never get built either. That is a pre-existing harness defect on this platform, outside this PR's fence (`tools/ci/`). The interop evidence above is against real upstream 3.4.4 and 3.4.1 binaries directly; proto 30 / 31 anchors still need the CI interop job.

oferchen:
Superseded by #7239, which carries the same matching-side fix **plus** the `crates/transfer` remainder half that this PR was missing.

This PR was correct but **wire-inert**: `build_signature_index` discarded the `sum_head` remainder, so every reconstructed basis block claimed `blength`, `short_final_block_len` returned `None`, and the tail probe never fired. Measured from the branch heads:

| | Literal / Matched |
|---|---|
| master | 103,300 / 101,500 |
| this PR (#7238) | 103,300 / 101,500 |
| #7239 | 102,900 / 101,900 |
| upstream 3.4.4 | 102,900 / 101,900 |

Everything of value here has moved to #7239:

- the boundary and OOB tests are on that branch as `crates/matching/tests/short_final_block_boundaries.rs` (commit `ac34b8e`), extended with the zsync 0.6 / 0.6.1 cases
- the `find_tail_match` documentation (BitHash bypass rationale, once-per-file cost contract, why no last-block short-circuit) went with them
- the two implementations were reconciled first and found to have no substantive difference; #7239's version was kept, and so were its versions of `index/tests.rs` and `integration_tests.rs`, which document intent better than mine and add a case I lacked

Branch `fix/sender-tail-block-match` retained, not deleted.

PR #7239, opened 2026-08-06
fix(matching): match the basis's trailing short block on the wire sender

oferchen:
## Overlaps #7238 - please pick one

#7238 targets the same defect and reaches the same design in `crates/matching`
(EOF tail probe, stripe-seam gating, index constructor accepting a sub-block
basis), independently. It is missing the half that lives in `crates/transfer`,
and **without that half the change does not move a single wire byte.**

Measured, both binaries built from their own heads, oc as sender against an
upstream 3.4.4 daemon receiver, same fixture:

| | Literal | Matched |
|---|---|---|
| master (39fe17d6c) | 103,300 | 101,500 |
| **#7238 (f979ec9fb)** | **103,300** | **101,500** |
| this branch | **102,900** | **101,900** |
| upstream 3.4.4 | 102,900 | 101,900 |

#7238's unit tests pass because they build the index from a real
`FileSignature`, which carries true per-block lengths. The wire path never does:
`build_signature_index` gives every block `block_length`, so `find_tail_match` -
which requires `block.len() == tail_len` - can never fire on a real transfer. I
hit exactly this trap first: my matching-only build measured no change at all.

Happy either way - land this one, or move the `crates/transfer` half into #7238
and close this. What must not ship is the matching-only version, which reads as
a fix and measures as a no-op.

## The defect

oc's sender never matched the basis's trailing short block, so every delta paid
for it as literal. This is **not** a redo defect - it reproduces with no
`--append`, no `--append-verify` and no phase 2 anywhere.

## Two halves

**Data model (`crates/transfer`).** `build_signature_index` set the layout
remainder to 0, commented *"remainder unknown from wire format"*. It is not
unknown: it is the fourth field of the sum head the receiver's generator wrote
(`io.c:write_sum_head()`), and upstream's sender applies it to the last block -
`if (i == s->count-1 && s->remainder != 0) s->sums[i].len = s->remainder;`
(`sender.c:109-112`). Every block claimed a full `block_length`, and a short
block matches only when the lengths agree (`match.c:222-224`). The remainder is
now carried on `DeltaGeneratorConfig` and applied to the trailing block; the
basis length is summed from real block lengths (`sender.c:126`).

**Scan (`crates/matching`).** The EOF drain flushed the trailing window as
literals. Upstream runs its scan to `end = len + 1 - s->sums[s->count-1].len`
(`match.c:174`) and shrinks the rolling window at EOF (`match.c:321-331`), so
exactly one window - the file's final `remainder` bytes - can match. The drain
now stops at that length and probes once.

## The coalescing hazard

The tail token is emitted standalone and never folded into a seq-match run. A
run is carried as `len = run_len * block_len`, and `script_to_wire_delta` expands
any `Copy` whose `len` is a whole multiple of the block length into that many
FULL-length ops - so a short block swept into a run would go out claiming
`block_len` bytes, over-reading the basis and desyncing the compressed-token
dictionary both peers feed from the matched length
(`token.c:685 see_deflate_token()`). Emitting it after the match loop has
flushed its run makes that structural, and
`tail_copy_is_never_coalesced_into_a_seq_match_run` pins it against the
longest-possible run.

Only a scan reaching the true end of the source may probe: `generate_chunked`
splits into stripes, and a stripe seam is not upstream's `end` boundary, so
`tail_probe` is set for the final stripe alone, with a test that a mid-file
stripe leaves the short block alone.

## The regression the test suite caught

Applying real block lengths made `daemon_pull_delay_updates_stages_through_tmp_dir`
fail with `failed to create signature index`: its destination basis is 5 bytes,
so the only block is short, and the index constructor rejected any signature
without a full-length block. That limitation was invisible while every wire block
was forced to `block_length`. Upstream builds a hash table from such a signature
like any other (`match.c:75-87`); only an empty one is unusable (`match.c:393`).
Indexing only full blocks in the tables is kept - it is upstream's length check
expressed in the data structure, with the short block found by
`find_tail_match`'s direct scan.

## The stale justification

The comment declining the tail match cited `token.c:665` as proof that the
upstream compressed-delta receiver rejects such a token stream. That line is a
generic `inflate returned %d` failure, unrelated - upstream sends short-block
tokens routinely. Its other claim, that the index constructor would fail for a
single-partial-block basis, was true, and is fixed here rather than worked
around.

## Tests that changed because they encoded the gap

- two asserted the index constructor returns `None` for a sub-block basis;
- one floored the legal block index at `len / block_len`, silently assuming the
  trailing block is never referenced, and required every copy length to be a
  whole multiple of the block length;
- the gated-fallback test now sees the tail copied too, which is what upstream's
  ungated scan does. Its stated intent (the lone full block must not be demoted
  to a literal) is asserted directly instead of via a total.

## Verification

- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` clean
- `cargo nextest run --workspace --all-features -E 'package(transfer) or package(engine) or package(protocol) or package(matching)'`:
  14,008/14,010 passed. **Golden byte tests (package(protocol)) all pass** - this
  touches the wire, so that was the gate. The 2 failures are
  `engine local_copy::tests::{execute_preserves_read_only_permissions,
  archive_preserves_mixed_permissions_across_files}`, both
  `remove extended attribute ... Permission denied` on a local macOS temp dir,
  reproduced on an untouched base commit and unrelated to this change.
- Cross-implementation, against the real upstream 3.4.4 binary: oc sender ->
  upstream daemon receiver is now byte-exact on the delta split (above), and oc
  receiver <- upstream daemon sender still reconstructs identical bytes. The
  `--append-verify` phase-2 redo push cell (upstream receiving) reports 2
  transfers, literal 205,300 / matched 101,900 - byte-identical to upstream.
- `tools/ci/run_interop.sh` was **not** run locally: the harness is
  Ubuntu/Debian-oriented (fetches `.deb`s) and does not run on this macOS host.
  CI's required interop job is the gate for it.

There is a residual 20-byte difference in `Total bytes sent` (103,650 vs
upstream's 103,628) in that probe. It is unrelated protocol overhead, present
identically before this change (104,048 vs 103,628 = the 400 literal bytes plus
the same 20), and is not touched here.

### DIFF
--- a/crates/matching/src/generator.rs
+++ b/crates/matching/src/generator.rs
@@ -372,7 +372,7 @@ impl DeltaGenerator {
         let prune_matched = self.prune_matched;
         #[cfg(not(any(test, feature = "bench-internal")))]
         let prune_matched = true;
-        self.generate_with_prune(reader, index, prune_matched)
+        self.generate_with_prune(reader, index, prune_matched, true)
     }
 
     /// Reports whether emitting a `Copy` for basis block `idx` is safe at the
@@ -407,11 +407,19 @@ impl DeltaGenerator {
     /// only read-only lookups on `index` (no [`DeltaSignatureIndex::mark_consumed`]
     /// or [`DeltaSignatureIndex::reset_consumed`]), so a shared `&index` is safe
     /// to scan from multiple rayon workers concurrently.
+    ///
+    /// `tail_probe` says whether this scan reaches the true end of the source.
+    /// Upstream's scan runs to `end = len + 1 - s->sums[s->count-1].len`
+    /// (`match.c:174`), so the basis's short trailing block is reachable only at
+    /// the file's own end; a stripe of [`Self::generate_chunked`] that stops
+    /// mid-file must not probe it, or a short block would match bytes upstream
+    /// never offers it.
     fn generate_with_prune<R: Read>(
         &self,
         mut reader: R,
         index: &DeltaSignatureIndex,
         prune_matched: bool,
+        tail_probe: bool,
     ) -> io::Result<DeltaScript> {
         let block_len = index.block_length();
         let mut window = RingBuffer::with_capacity(block_len);
@@ -700,18 +708,87 @@ impl DeltaGenerator {
             }
         }
 
-        // Drain any bytes still in the window as literals (window held fewer
-        // than block_len bytes at EOF, so no further match is possible).
+        // EOF tail match: when the basis ends with a short block, upstream can
+        // still match it. Its scan runs to `end = len + 1 -
+        // s->sums[s->count-1].len` (`match.c:174`) and shrinks the rolling
+        // window at EOF (`match.c:321-331`: `more` is false, so `--k`), while
+        // every candidate must satisfy `l = MIN(blength, len-offset) ==
+        // s->sums[i].len` (`match.c:222-224`). Only the trailing short block has
+        // a length below `blength`, so exactly one window can match it: the
+        // file's final `short_len` bytes. Drain down to that length and probe
+        // once, rather than re-probing at every shrink step.
         //
-        // Upstream rsync matches the basis's short final block here via
-        // `l = MIN(blength, len-offset)` (`match.c:222-224`), but on the wire
-        // sender path that would (a) fail signature-index construction for
-        // small single-partial-block files and (b) produce a token stream the
-        // upstream compressed-delta receiver rejects (`token.c:665`). The
-        // "trailing partial block emitted as literal" is a pre-existing wire
-        // efficiency gap, not a correctness issue - reconstruction is
-        // byte-exact either way - so the tail match stays confined to the
-        // local-copy delta path (`engine::local_copy`).
+        // The token MUST be emitted standalone, never folded into a seq-match
+        // run: a run is carried as `len = run_len * block_len`, and the wire
+        // layer expands any `Copy` whose `len` is a whole multiple of the block
+        // length into that many full-length ops. A short block inside a run
+        // would therefore go out claiming `block_len` bytes, over-reading the
+        // basis and desyncing the compressed-token dictionary both peers feed
+        // from the matched length (`token.c:685 see_deflate_token()`). Emitting
+        // it here, after the match loop has flushed its run, makes that
+        // structural: no run is open at this point.
+        let short_tail_len = index
+            .block_count()
+            .checked_sub(1)
+            .map(|last| index.block(last).len())
+            .filter(|&len| len > 0 && len < block_len);
+        if tail_probe && let Some(tail_len) = short_tail_len {
+            while window.len() > tail_len {
+                if let Some(byte) = window.pop_front() {
+                    pending_literals.push(byte);
+                    offset += 1;
+                }
+            }
+            if window.len() == tail_len {
+                let (first, second) = window.as_slices();
+                rolling.reset();
+                rolling.update(first);
+                if !second.is_empty() {
+                    rolling.update(second);
+                }
+                let tail_match = index
+                    .find_tail_match(
+                        rolling.digest(),
+                        first,
+                        second,
+                        prune_matched.then_some(&matched_blocks),
+                    )
+                    .filter(|&idx| self.basis_offset_ok(index, block_len, idx, offset));
+                if let Some(match_idx) = tail_match {
+                    matches += 1;
+                    debug_log!(
+                        Deltasum,
+                        3,
+                        "potential tail match at {} i={} len={}",
+                        offset,
+                        match_idx,
+                        tail_len
+                    );
+                    if !pending_literals.is_empty() {
+                        literal_bytes += pending_literals.len() as u64;
+                        total_bytes += pending_literals.len() as u64;
+                        let filled =
+                            std::mem::replace(&mut pending_literals, Vec::with_capacity(block_len));
+                        tokens.push(DeltaToken::Literal(filled));
+                    }
+                    let block = index.block(match_idx);
+                    tokens.push(DeltaToken::Copy {
+                        index: block.index(),
+                        len: block.len(),
+                    });
+                    total_bytes += block.len() as u64;
+                    matched_blocks.mark_matched(match_idx);
+                    if prune_matched {
+                        index.mark_consumed(match_idx as u32);
+                    }
+                    window.clear();
+                }
+            }
+        }
+
+        // Drain any bytes still in the window as literals (the window holds
+        // fewer than block_len bytes at EOF, so no full-block match is possible,
+        // and the short-block probe above has already run).
         while let Some(byte) = window.pop_front() {
             pending_literals.push(byte);
         }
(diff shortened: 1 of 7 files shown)


oferchen:
Superseded by #7238, which now carries both halves plus strictly more.

When these two were compared earlier, #7239 was the only one with the
`crates/transfer` remainder half and #7238 measured identical to master.
That is no longer true: #7238 now has both halves and additionally carries

- `crates/matching/tests/short_final_block_match.rs`, a dedicated test file
  with no equivalent here;
- `crates/matching/src/index/mod.rs`, documenting why `find_tail_match`
  bypasses the BitHash / tag table / compact lookup (they key on rolling sums
  over `block_length` bytes; a short block's sum spans `tail_len`, and
  `populate_index` never inserts a partial block, so probing them is a
  category error rather than a slower path);
- the two-commit split with the matching half first, since `from_signature()`
  accepting a signature with no full block is a prerequisite for the
  remainder change rather than a cleanup alongside it.

Its verification is also broader: measured against upstream 3.4.4 *and* 3.4.1
as the peer, an exact-multiple (`remainder == 0`) guard pinned as a unit test
on block lengths rather than on totals, and the golden byte tests re-run at a
point where the reconstructed index can actually carry a short final block -
so they exercised the changed path instead of being insensitive to it.

Closing rather than merging both. The branch is left intact.

oferchen:
Tests ported from #7238 (now superseded), plus the verification carried over. Commit `ac34b8e`.

## Reconciliation of the two matching implementations

I diffed them line by line before porting anything. **No substantive difference.** Two independent implementations of the same upstream rule converged on the same design: a `tail_probe`/`tail_match` scan-extent parameter, drain-to-`tail_len` then probe once, fresh rolling digest over the tail, prune filter passed through, `basis_offset_ok` applied, standalone `Copy` carrying `block.len()`, `window.clear()` on hit. Both `from_signature` changes reduce to "return `None` only when `blocks.is_empty()`".

Kept **this branch's** version throughout, per instruction to minimise churn. Where both had touched a test I also kept this branch's - `index/tests.rs` and `integration_tests.rs` here are better: they document intent more fully, add `from_signature_returns_none_for_an_empty_signature` (which I did not have), and assert the fixture premise `tail_len < block_len` in `copy_tokens_reference_valid_blocks`.

I also did not port three of my tests, because the in-module tests here already cover them **in a stronger form**:

| my test | covered by | why theirs is better |
|---|---|---|
| standalone-Copy pin | `tail_copy_is_never_coalesced_into_a_seq_match_run` | also asserts every non-tail Copy is a whole multiple and that the lens sum to the basis |
| tail matched not literal | `tail_short_block_is_matched_as_a_standalone_copy` | equivalent |
| striped-scan guard | `a_mid_file_stripe_does_not_probe_the_short_tail_block` | calls `generate_with_prune` with `tail_probe` false *and* true, which isolates the guard; mine planted content mid-source, which is indirect |

## What the new file adds

`crates/matching/tests/short_final_block_boundaries.rs` - 7 tests at the edges, which nothing here covered. The existing cases sit at `tail_len` 5 and 7, comfortably inside the range.

- `tail_len == 1` and `tail_len == block_length - 1` - both extremes of the legal range
- exact-multiple basis emits no short `Copy` (`remainder == 0`, upstream's `&& s->remainder != 0`)
- a basis of one short block and nothing else, end to end through the scan
- **every slice split** `0..=tail_len` through `find_tail_match`, including both degenerate ends
- a wrapped ring buffer driven through the real scan, so the two-non-empty-slices arm is executed rather than merely reachable
- the consecutive-match gate leaves the tail literal
- `seq_matches == 1` demotes nothing

The last two exist because of documented zsync history (NEWS, github.com/cph6/zsync): **0.6** fixed an out-of-bounds access processing the last block, and **0.6.1** fixed `sequential_matches == 1` giving *false negatives*. The first is a slicing bug in exactly the two-slice path; the second is silent - a real match quietly demoted, which reads as "works, just less efficient". The `seq==1` test uses a source identical to the basis so `literal_bytes` must be exactly zero, which no false negative can satisfy; a totals-only assertion on a partially-matching fixture could not distinguish it.

Also documents on `find_tail_match` why it bypasses the BitHash, tag table and compact lookup: all three are keyed on rolling sums over `block_length` bytes, a short block's sum spans `tail_len` bytes, and `populate_index` never inserts a partial block - so probing them is a category error, not a slower path. Plus the once-per-file call contract, and why narrowing the scan to the last block would be unsound for a public API over a caller-supplied `FileSignature`.

## The byte-identity gate

Captured the actual client-to-server wire through a tee proxy, upstream 3.4.4 as the daemon in both runs, same fixture, same mtimes, `--checksum-seed=42 --checksum-choice=md5 -B700` so the streams are deterministic and the arg spellings match.

**The delta token stream is byte-identical to upstream: 103,442 of 103,453 bytes after the flist entry.**

Every residual byte accounted for:

1. **Server arg vector** - upstream emits `--server -e.LsfxCIvu -B700 --checksum-choice=md5 --checksum-seed=42`; oc emits the same options in a different **order** (`--checksum-choice` before `-e.`) and spells `-B700` as `--block-size=700`. 15 bytes.
2. **Goodbye framing** - the final 3 payload bytes `00 00 00` are framed by upstream as one 2-byte multiplex frame plus one 1-byte frame; oc emits three 1-byte frames. Decoded both: **payload identical**, frame boundaries differ.

Neither is the delta, and neither is introduced here. **Control:** the same capture on an exact-multiple basis (`293 x 700`, `remainder == 0`, so the tail probe is inert) produces the **identical** residual - same 11-vs-15 goodbye bytes, same arg delta. That is direct evidence both are independent of the tail path rather than an inference.

Also, without `--checksum-choice` oc advertises `xxh128 xxh3 xxh64 md5 md4 sha1` where upstream advertises `md5 md4 sha1`. Pre-existing capability advertisement, orthogonal to this change.

## Cells re-measured on this branch

Binary rebuilt with an explicit exit-code check and its revision confirmed (`#3c4038597`) - an earlier run of mine copied a stale binary after a silently OOM-killed build, so this is verified rather than assumed. Every destination `cmp`-verified. Literal / Matched:

| cell | upstream 3.4.4 | this branch |
|---|---|---|
| daemon push | 102,200 / 102,600 | 102,200 / 102,600 |
| ssh push | 102,200 / 102,600 | 102,200 / 102,600 |
| exact multiple (remainder 0) | 102,200 / 102,900 | 102,200 / 102,900 |
| sub-block basis (400 B, local) | 0 / 400 | 0 / 400 |
| local (`crates/engine`) | 102,200 / 102,600 | 102,600 / 102,200 |

The last row is #365 - `crates/engine` has the same defect in its own scan loop and is out of scope here.

## Test runs

- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` clean, no waivers (it caught a real `doc_overindented_list_items` in my new file; fixed)
- per crate `--all-features --no-fail-fast`: matching 370, signature 234, engine 5032, protocol 6029, transfer 2684 - **14,349 passed, 0 failed**
- goldens: no wire byte moved. Stating the caveat explicitly, because it was mine: on #7238 unchanged goldens were **not** proof of neutrality, since the wire-reconstructed index could not carry a short final block and the goldens were insensitive to the path. With the remainder half in, that insensitivity is gone and this run **is** load-bearing.

## Not done

`tools/ci/run_interop.sh` still **cannot run on this host**, and it is not disk - I retried with 12G free and it fails identically, building rsync **2.6.9** on modern glibc:

```
syscall.c:296:17: error: conflicting types for 'lseek64'; have 'off64_t(void)'
/usr/include/unistd.h:350:18: note: previous declaration ... '__off64_t(int, __off64_t, int)'
make: *** [Makefile:59: syscall.o] Error 1
```

2.6.9 is in `extra_build_versions` - built and cached only, no scenarios wired to it - but under `set -euo pipefail` its failure aborts before **any** version installs, so 3.0.9 / 3.1.3 / 3.4.4 never build and `target/interop/upstream-install` is empty. Tracked as #369. The interop evidence above is against real upstream 3.4.4 and 3.4.1 binaries directly; proto 30 / 31 anchors still need the CI job, which is unavailable during the GitHub Actions outage.

PR #7246, opened 2026-08-06
test(matching): pin short-block boundaries and the seq==1 demotion

oferchen:
Follow-up to #7238 (merged). Test-only - no production code changes.

#7238 pinned the main behaviours of the trailing-short-block match. These cover the **edges** of that handling, which have a documented history of going wrong in the code this mirrors.

zsync - whose `librcksum` these techniques derive from - shipped two fixes in exactly this area (NEWS, github.com/cph6/zsync):

- **0.6** - "fix out-of-bounds memory access when processing **last block** of non-compressed download"
- **0.6.1** - "fix librcksum handling of zsync streams with **sequential_matches == 1**; it was giving **false negatives** when applying the rsync algorithm"

The first is a length/slicing bug on the final short block. Rust turns that class into a panic rather than silent corruption, but a panic on the last block of a transfer is still an availability defect, and it would only reproduce on files whose length is not a block multiple - which is most files. The second is *silent*: a real match demoted to a literal, which reads as "works, just less efficient" and so survives review.

## Added

| test | why |
|---|---|
| `tail_lengths_at_both_extremes_are_matched_and_reconstruct` | `tail_len == 1` and `block_length - 1`. Off-by-one at either bound is the shape of the 0.6 defect; existing cases sit inside the range. |
| `an_exact_multiple_basis_emits_no_short_copy` | upstream guards with `&& s->remainder != 0` (`sender.c:109-110`). Getting this wrong breaks the common evenly-divided case, not an edge case. |
| `find_tail_match_accepts_every_slice_split_of_the_window` | every split `0..=tail_len`, including both degenerate ends. `find_tail_match` takes the window as **two** slices because the ring buffer may have wrapped, and splitting at the wrong place is precisely how the 0.6 out-of-bounds access happened. |
| `a_wrapped_window_still_matches_the_tail_through_the_scan` | drives the same two-slice path through the real scan, so the two-non-empty-slices arm is *executed* rather than merely reachable. |
| `seq_matches_one_demotes_nothing` | the 0.6.1 false-negative class. Source is the basis, so `literal_bytes` must be exactly zero - which no false negative can satisfy. |

That last one is why these are not totals-only assertions: on a partially-matching fixture, a demoted match and correct behaviour produce the same shape of output, and only an exact-zero literal count distinguishes them.

The fixture uses a 64-bit LCG rather than a short-cycle pattern. With a repeating period, an unaligned window near EOF matches an earlier full block by content and swallows the trailing bytes - which would make "the tail was not matched" a statement about the data instead of about the matcher. That is not hypothetical; it is how the first draft of these tests failed.

Deliberately **not** duplicated from #7238's `short_final_block_match.rs`: the single-short-block basis and the consecutive-match-gate pin are already there.

## Verification

- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` clean, no waivers
- `cargo nextest run -p matching --all-features --no-fail-fast` - 375 passed, 3 skipped

All five are new failures against a hypothetical regression and pass on current master, so they are guards rather than a fix.

Note on CI: GitHub Actions is in a major outage (webhooks throttled, many push/PR events not triggering runs), so checks may not appear on this PR. The verification above is local.

### DIFF
--- a/crates/matching/tests/short_final_block_boundaries.rs
+++ b/crates/matching/tests/short_final_block_boundaries.rs
@@ -1,0 +1,230 @@
+//! Boundary cases for the basis's trailing short block.
+//!
+//! `short_final_block_match.rs` pins the main behaviours: the tail is matched,
+//! it is emitted as a standalone `Copy`, and a mid-file stripe does not probe
+//! it. This file covers the edges of that handling, which have a documented
+//! history of going wrong.
+//!
+//! zsync - whose `librcksum` these techniques derive from - shipped two fixes in
+//! exactly this area (NEWS, github.com/cph6/zsync):
+//!
+//! - 0.6 - "fix out-of-bounds memory access when processing last block of
+//!   non-compressed download"
+//! - 0.6.1 - "fix librcksum handling of zsync streams with sequential_matches
+//!   == 1; it was giving false negatives when applying the rsync algorithm"
+//!
+//! The first is a length/slicing bug on the final short block. Rust turns that
+//! class of defect into a panic rather than silent corruption, but a panic on
+//! the last block of a transfer is still an availability defect, and it would
+//! only reproduce on files whose length is not a block multiple - which is most
+//! files. The second is a *silent* defect: a real match quietly demoted to a
+//! literal, which looks like "works, slightly less efficient" and so survives
+//! review. Neither is caught by a test that only checks byte totals.
+
+use matching::{DeltaGenerator, DeltaScript, DeltaSignatureIndex, DeltaToken, apply_delta};
+use protocol::ProtocolVersion;
+use signature::{
+    SignatureAlgorithm, SignatureLayoutParams, calculate_signature_layout, generate_file_signature,
+};
+use std::io::Cursor;
+use std::num::{NonZeroU8, NonZeroU32};
+
+const BLOCK_LEN: u32 = 16;
+
+/// Deterministic, non-repeating filler.
+///
+/// A short-cycle pattern would make these tests lie: with a period that divides
+/// into the scan, an unaligned window near EOF can match an earlier full block
+/// by content and swallow the trailing bytes, so "the tail was not matched"
+/// would be a statement about the data rather than about the matcher. A 64-bit
+/// LCG has a period far beyond any fixture here.
+fn pattern(len: usize) -> Vec<u8> {
+    let mut state: u64 = 0x2545_F491_4F6C_DD1D;
+    (0..len)
+        .map(|_| {
+            state = state
+                .wrapping_mul(6_364_136_223_846_793_005)
+                .wrapping_add(1_442_695_040_888_963_407);
+            (state >> 33) as u8
+        })
+        .collect()
+}
+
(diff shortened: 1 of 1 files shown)


oferchen:
## Confirmed by revert experiment, not by reading

Both questions answered empirically.

### 1. The wire-side probe recomputes; it does not truncate

`crates/matching/src/generator.rs`, the EOF probe on master:

```rust
while window.len() > tail_len {
    if let Some(byte) = window.pop_front() {
        pending_literals.push(byte);
        offset += 1;
    }
}
let tail_hit = {
    let tail = window.as_slice();
    let digest = RollingDigest::from_bytes(tail);   // fresh, over the drained window
    index.find_tail_match(digest, tail, &[], ...)
```

`RollingDigest::from_bytes(tail)` computes sum1/sum2 from scratch over exactly the `tail_len` bytes that remain. The scan's own `rolling` accumulator is never consulted at this point, so the stale-digest mechanism cannot reach the probe: there is no digest to be stale.

### 2. Yes, and here is the experiment

I patched the probe to the drain-only variant - kept the drain, replaced the recomputation with the stale accumulator, which is exactly what a naive fix would have:

```rust
-  let digest = RollingDigest::from_bytes(tail);
+  let digest = rolling.digest();
```

`cargo nextest run -p matching --all-features --no-fail-fast`: **369 passed, 6 failed.**

| test | failure |
|---|---|
| `tail_lengths_at_both_extremes_are_matched_and_reconstruct` (#7246) | `tail_len=1 must be Copied with its own length` - `left: Some(48)` vs `right: Some(1)` |
| `a_wrapped_window_still_matches_the_tail_through_the_scan` (#7246) | `left: Some(608)` vs `right: Some(9)` |
| `seq_matches_one_demotes_nothing` (#7246) | `no block may be demoted at seq_matches == 1` - `literal_bytes left: 11` vs `right: 0` |
| `short_final_block_is_matched_not_resent_as_literal` | already on master |
| `short_final_block_copy_is_standalone_and_carries_its_true_length` | already on master |
| `gated_one_full_block_plus_short_tail_still_matches_the_full_block` | already on master |

Restoring the recomputation returns the suite to **375 passed, 0 failed**.

So the discriminating coverage exists, and it is not thin: three of #7246's five tests catch it, plus three that were already on master. The `tail_len == 1` arm of `tail_lengths_at_both_extremes` fires first; the `block_length - 1` arm is in the same test and covers the case you named.

One nuance worth recording, because it cuts against the model rather than for it: the drain-only variant does not merely fail to match - it matches the **wrong** block. `Some(48)` where `Some(1)` is expected, `Some(608)` where `Some(9)` is expected. A stale digest describing `block_len` bytes still lands in the length-gated scan and can collide with some other basis block's rolling sum, which the strong sum then has to reject or not. That makes a drain-only fix worse than "sometimes misses the tail": on the local twin it would be a wrong-length `Copy` in the token stream. It is caught here only because the assertions check the Copy's **length and identity** rather than byte totals - the same reason a totals-only test would have passed the coalescing bug.

PR #7262, opened 2026-08-12
fix(daemon): close three refuse-options bypasses and match rules case-sensitively

oferchen:
Two ways a client could use an option the module had refused.

## 1. Bundle scan stopped at the first non-alphabetic byte

`-4z` ended the scan at `4`, so `z` was never tested against the refuse list -
while the decoder still enabled compression. The scan now continues past bytes
it does not recognise and stops only at `.`, exactly where the decoder stops.

### Why this deliberately does NOT model popt arity

Upstream can, because one popt pass both parses and marks refusals, so
enforcement and application cannot disagree. oc has **two** readers of the same
compact bundle, and the one that APPLIES the options -
`transfer::flags::ParsedServerFlags::parse` - walks every byte up to `.` with no
arity logic:

```rust
for &byte in &bytes[1..] {
    if byte == b'.' { break; }
    flags.parse_transfer_flag(byte);
}
```

An arity-aware scan therefore skips letters the decoder still acts on:
`-B4096z` enables compression the scan never sees. **While two readers exist,
the refuse scan must examine a superset of what the decoder acts on.**
Over-refusal fails closed; under-refusal is a hole. This was tried the other way
first and reverted.

## 2. The `[...]` glob was a second, weaker matcher

It compared a character class by literal byte membership, silently dropping two
constructs upstream's `dowild` supports:

- ranges - `lib/wildmatch.c:156-166`
- negation `[!...]` - `lib/wildmatch.c:139-143`

Its doc justified the omission as *"upstream's `[ardlptgoD]` expansion never uses
one"* - which describes oc's own expansion, not what an operator may write in
`rsyncd.conf`.

This is the fail-OPEN direction. Every other refuse-options divergence
over-refuses and so fails closed; here `refuse options = [A-Z]*` matched only
the literal bytes `A`, `-` and `Z`, so the options it was written to block were
**accepted**.

Upstream matches refuse rules with `wildmatch()` and nothing else -
`options.c:921-924` calls it for both `longName` and `shortName`, wild or not.
oc already ships that matcher as `filters::wildmatch` (a `dowild` port), and
`crates/daemon` already depended on the crate, so the hand-written copy is
deleted and the call delegates.

## Table

Replaces two drifted hand-written maps (46 vs 41 arms) with one 51-row table,
verified letter-for-letter against `options.c:600-856`: oc lacks none of
upstream's short letters and invents none.

## Known residual, deliberately not changed here

`canonical_option` lowercases, so `refuse options = O` also refuses `-o`.
Upstream's `wildmatch` is case-SENSITIVE. That **over**-refuses (fails closed)
and is an operator-visible policy change, so it is not bundled into a security
fix. Likewise `D`/`F`/`P` keep oc-specific long names where upstream has NULL -
all three over-refuse.

## Verification

Rebased onto `0ed2894a` and verified there:

| gate | result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` | pass |
| `cargo nextest run --workspace --all-features` | 32390 run / 32390 passed / 152 skipped |

The two glob tests were confirmed to execute by name, not merely to be absent
from the failure list.

### DIFF
--- a/crates/daemon/src/daemon/sections/module_parsing/refuse_options.rs
+++ b/crates/daemon/src/daemon/sections/module_parsing/refuse_options.rs
@@ -97,116 +97,111 @@ fn refused_option<'a>(module: &ModuleDefinition, options: &'a [String]) -> Optio
 ///
 /// upstream: options.c long_options[] - the canonical short/long pairing the
 /// daemon's popt-based refuse check uses to compare against `refuse options`.
-fn short_option_long_name(letter: char) -> &'static str {
-    match letter {
-        'v' => "verbose",
-        'q' => "quiet",
-        'h' => "human-readable",
-        'n' => "dry-run",
-        'a' => "archive",
-        'r' => "recursive",
-        'd' => "dirs",
-        'p' => "perms",
-        'E' => "executability",
-        'A' => "acls",
-        'X' => "xattrs",
-        't' => "times",
-        'U' => "atimes",
-        'N' => "crtimes",
-        'O' => "omit-dir-times",
-        'J' => "omit-link-times",
-        'o' => "owner",
-        'g' => "group",
-        'D' => "devices",
-        'l' => "links",
-        'L' => "copy-links",
-        'k' => "copy-dirlinks",
-        'K' => "keep-dirlinks",
-        'H' => "hard-links",
-        'R' => "relative",
-        'I' => "ignore-times",
-        'x' => "one-file-system",
-        'u' => "update",
-        'S' => "sparse",
-        'F' => "filter",
-        'C' => "cvs-exclude",
-        'W' => "whole-file",
-        'c' => "checksum",
-        'y' => "fuzzy",
-        'z' => "compress",
-        'P' => "partial",
-        'm' => "prune-empty-dirs",
-        'i' => "itemize-changes",
-        'b' => "backup",
-        's' => "secluded-args",
-        'V' => "version",
-        'B' => "block-size",
-        'T' => "temp-dir",
-        'M' => "remote-option",
-        'f' => "filter",
-        'e' => "e",
-        _ => "",
-    }
+/// One row of upstream's `long_options[]` as the refuse matcher needs it.
+///
+/// `long_options[]` is a single table that popt reads in both directions; oc
+/// previously transcribed it into two independent `match` blocks that had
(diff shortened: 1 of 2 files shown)


oferchen:
Follow-up commit `ce9edbad` removes the case-folding divergence that the original description listed as a deliberate residual.

**Upstream:** `options.c:923-924` matches refuse rules with `wildmatch`. The case-folding variant `iwildmatch` (`lib/wildmatch.c:307-318`, `force_lower_case = 1`) exists and is deliberately not called, and no `tolower`/`strcasecmp` appears in the refuse path. Matching is case-sensitive.

oc folded case in **three** places, and the third is a bypass rather than a cosmetic difference:

| site | effect | direction |
|---|---|---|
| `canonical_option` | lowercased rule and client option | over-refuse |
| short-letter compare | lowercased the candidate letter | over-refuse |
| **`is_option_vital`** | retried the vital lookup lowercased | **under-refuse** |

Vitality belongs to one `long_options[]` row, and case-paired letters are different rows: `-n` (dry-run) is vital and immune to `refuse options = *`; `-N` (crtimes) is not. Folding let `-N` inherit `-n`'s immunity, so a wildcard rule silently spared it.

The over-refusing sites were not harmless either: `refuse options = O` (--omit-dir-times) also refused `-o` (--owner), breaking every `-a` transfer against that module.

**Behaviour change for release notes:** `refuse options = DELETE` no longer refuses `--delete`. That matches upstream, but anyone relying on oc's case-insensitivity loses coverage on upgrade.

**No concurrency impact:** the refuse path is pure functions over `&ModuleDefinition` and `&[String]`, evaluated once per connection at module-open time; `refuse_options.rs` contains no `Mutex`, `RwLock`, atomic, `static mut` or thread.

Two existing tests asserted the folded behaviour and pinned the divergence; both rewritten to assert upstream's rule, each recording why the old assertion was wrong.

**Verification** on `0ed2894a`: fmt pass, clippy `-D warnings` pass, `cargo nextest run --workspace --all-features` 32391 passed. One unrelated failure, `fast_io::io_uring_mmap_pressure`, reproduces **identically on unmodified master** under the same full-workspace load and passes in isolation - tracked separately, not from this branch.

PR #7266, opened 2026-08-12
ci: install cargo-nextest from one pinned composite action

oferchen:
## Problem

PRs #7234 and #7265 both failed the `Windows (nightly)` cell at step 5,
`Install cargo-nextest (Windows direct download)`, roughly 76 s in. Build and
Test were **skipped**, so neither PR's code was compiled, let alone tested.
Two unrelated PRs, same step, same cell - the failure was the install step, not
the changes.

That step was a single `Invoke-WebRequest` against `https://get.nexte.st/latest/windows`
under `$ErrorActionPreference = 'Stop'`. One transient fetch failure takes the
entire Windows test cell with it.

It was also copy-pasted **15 times across 10 workflow files**, so any change to
it had to be made fifteen times, and the copies had already drifted (three
different comment bodies for the same eight lines).

## Change

One composite action, `.github/actions/install-cargo-nextest-windows`, replaces
all 15 copies. Two properties, in order:

1. **Reproducible.** The version is pinned (`0.9.114`) instead of tracking
   `latest`, so re-running an old commit installs what that commit was tested
   against. `latest` also meant the fetched artifact was unpinned
   supply-chain-wise. The pin lives in the action's input default - one source
   of truth, no `env` threaded through ten files.

2. **No network on the common path.** `Swatinem/rust-cache` already runs before
   this step in every one of these jobs and restores `~/.cargo/bin`, but the
   old step downloaded unconditionally anyway. The action now skips the fetch
   when the pinned binary is already present, so a cache hit performs no
   network I/O and cannot be taken down by a transient failure.

The installed version is verified after extraction: a silently wrong version
would make the matrix test something other than what the pin claims.

**This deliberately contains no retry and no backoff.** The download is still
attempted exactly once when genuinely needed, and a failure there fails the job
immediately and loudly. The fix removes the reason to re-attempt rather than
re-attempting.

## Upstream

rsync 3.4.4 has no GitHub Actions Windows CI, so there is no upstream behaviour
to mirror here. This is oc-only infrastructure and the oracle is
reproducibility, not fidelity.

## Verification

- All 15 direct-download sites removed: `grep -rc get.nexte.st .github/workflows/` returns nothing
- 15 composite call sites across the same 10 files
- Every workflow and the action parse as valid YAML
- Every call site has an `actions/checkout` before it (required for a local `uses: ./...`)
- The one conditional site retains its `if: runner.os == 'Windows'`

### DIFF
--- a/.github/actions/install-cargo-nextest-windows/action.yml
+++ b/.github/actions/install-cargo-nextest-windows/action.yml
@@ -43,8 +43,24 @@ runs:
         $exe = Join-Path $cargoBin 'cargo-nextest.exe'
 
         function Get-NextestVersion($path) {
-          # `cargo-nextest --version` prints e.g. "cargo-nextest 0.9.114 (ff4ee9a ...)".
-          (& $path --version) -replace '^cargo-nextest\s+([0-9][0-9.]*).*$', '$1'
+          # `cargo-nextest --version` prints a MULTI-LINE block, not one line:
+          #
+          #   cargo-nextest 0.9.114
+          #   release: 0.9.114
+          #   commit-hash: 9daab1c50cd0389236508435f71434f01aa2eca1
+          #   commit-date: 2025-11-19
+          #   host: x86_64-pc-windows-msvc
+          #
+          # So `& $path --version` yields a string ARRAY. Taking the first line
+          # is load-bearing: `-replace` over an array returns an array (only the
+          # first element matches, the rest pass through untouched), and `-ne`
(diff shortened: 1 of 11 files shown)


oferchen:
Marking draft: every Windows cell fails at the new install step (exit code 1, no further detail in the annotation yet, run still queued so logs are not retrievable). Not pushing a speculative second patch - the log will name the failing line. Master is unaffected; this PR is unmerged.

PR #7267, opened 2026-08-12
fix(metadata): audit the NTFS ACEs the rsync wire format cannot carry

oferchen:
Reading an NTFS DACL into an `RsyncAcl` drops every ACE the POSIX-style rsync wire format cannot express - deny, audit, alarm and object ACEs - plus any ACE whose SID resolves to no account and any ACE `GetAce` refuses to return. All of those drops were silent, so an operator syncing a tree containing a deny ACE had no way to learn the denial did not travel.

The lossiness is a deliberate scope choice and is unchanged. The silence is the defect.

The machinery to fix it already existed on the opposite side of the same translation: the apply direction records each unmappable principal in `DroppedAces` and emits a per-file audit line through `warn_dropped_aces`. This reuses it rather than adding a second mechanism.

- `dacl_to_rsync_acl` returns its `DroppedAces` alongside the `RsyncAcl`.
- Both callers report through the one funnel: `get_rsync_acl`, and the lossy named-ACE fallback in `sync_acls` (reached only when the volume refuses to serve a descriptor for the verbatim SDDL round-trip).
- `warn_dropped_aces` takes the direction-specific clause as a parameter, so one formatter serves both directions.

An ACE whose mask carries no `FR`/`FW`/`FX` bit is still skipped without a record: it conveys no POSIX-expressible permission, so nothing is lost. `resolve_acl_aces` already applies that same rule in the opposite direction.

Reading a SID out of a non-allow ACE is only sound for the four non-object types, which share `ACCESS_ALLOWED_ACE`'s `header + Mask + SidStart` layout. Object ACEs (types 5-8) interpose an object GUID before `SidStart` and are named by type alone.

## Test

`dacl_to_rsync_acl_audits_deny_aces_instead_of_dropping_them_silently` builds a two-ACE DACL in memory (one allow, one deny, both for the locale-independent `WinWorldSid`) and asserts the allow ACE still converts while the deny ACE leaves exactly one audit record naming its type. It is the mirror of the existing `resolve_acl_aces_records_unmappable_sids`, which pins the same contract for the apply direction.

## Verification

`cargo fmt --all -- --check` clean. The changed code is `cfg(windows)`; the Windows cells are the compile and behaviour oracle for it, so this stays a draft until they are green.


oferchen:
Windows type-check completed locally while GitHub Actions runners are unavailable.

`cargo check --locked -p metadata --tests --features acl,xattr --target x86_64-pc-windows-gnu` (C toolchain supplied by `cargo-zigbuild`s zig cc wrapper, which is what made this possible - a plain cross-check dies in `zstd-sys` for want of a cross C compiler) finishes clean, and `--lib` builds fully in 3m38s.

`--tests` is the load-bearing part: it type-checks `crates/metadata/src/acl_windows/tests/dacl.rs`, including the Win32 calls this PR introduces (`CreateWellKnownSid`, `InitializeAcl`, `AddAccessAllowedAce`, `AddAccessDeniedAce`, `PSID`, `WinWorldSid`). Those signatures were the stated risk in the PR description.

Scope of what this does and does not establish:
- Establishes: the `cfg(windows)` code and its test type-check against windows-rs, on `x86_64-pc-windows-gnu`.
- Does NOT establish: that it links or runs under `x86_64-pc-windows-msvc` (the target CI uses), or that the test passes at runtime - that needs a Windows runner.

Staying draft until the Windows cells run.

PR #7281, opened 2026-08-13
docs(ci): repoint upstream-testsuite citations at runtests.py

oferchen:
## What

`tools/ci/run_upstream_testsuite.sh` claimed to mirror the contract of upstream's `runtests.sh` and cited `runtests.sh:205-215` and `runtests.sh:254`.

**There is no `runtests.sh` in rsync-3.4.4 or rsync-3.5.0.** Both ship `runtests.py`, and have done since before the currently pinned version. The citations named a file a reader cannot open — and worse, cannot use to tell whether the behaviour was dropped upstream or the citation was simply wrong.

## The behaviour is real; only the pointer was wrong

Both cited behaviours exist in `runtests.py` under the same names:

| oc site | upstream |
|---|---|
| `find_setfacl_nodef()` | `runtests.py find_setfacl_nodef()` — 3.5.0:146-161, 3.4.4:67-82 |
| scratch-dir prep (clear default ACLs + setgid) | `runtests.py prep_scratch()` — 3.5.0:230-241 |

## Anchored on the function, not the line

The offsets differ between the two pinned trees (67 vs 146 for the same function), which is precisely the part that rots. Each citation now names the **function first**, with line numbers as a hint. That is the hybrid form, and it is the right shape here regardless of how the wider line-numbers-vs-anchors question lands.

## Verification

- Both functions confirmed present in **both** pinned trees (3.4.4 and 3.5.0).
- Each quoted line number confirmed to land exactly on the function's `def`.
- Zero remaining `runtests.sh` references.
- `bash -n` clean, `shellcheck -S error` clean.
- No behaviour change — comments only.

## Why it existed

Found by re-deriving upstream claims from source instead of carrying them forward. It is a **gap in the gate**, not a one-off: `tools/ci/citation_drift_audit.py` globs `crates/{crate}/src/**/*.rs` only, so every upstream citation under `tools/`, `.github/scripts/` and `xtask/` is unchecked and free to name a nonexistent file. A follow-up task covers extending that gate with a cheap file-existence check, which would have caught this immediately and has no false-positive problem.


oferchen:
Superseded: this content is already on master.

#7286 was stacked on this branch and carried the change with it, so `tools/ci/run_upstream_testsuite.sh` on master is now **byte-identical** to this PR's version:

```
git diff origin/master:tools/ci/run_upstream_testsuite.sh <this-branch>:tools/ci/run_upstream_testsuite.sh
  -> no output (identical)
```

`git grep runtests.sh` across `tools/ci/` on master returns nothing, so the phantom citations this PR fixed are gone. Merging would be a no-op; closing frees the CI matrix it was holding.

The ordering concern (that #7286's gate would be red on master until this landed) does not apply for the same reason - the gate and the citations it needs landed together in one commit.

PR #7301, opened 2026-08-13
fix(receiver): honour -B / --block-size on every wire transport

oferchen:
## What

`-B` / `--block-size` was honoured only on the local-copy path. Every network transfer silently fell back to the 700-byte `BLOCK_SIZE` default (rsync.h:25), so an operator-tuned block size produced a different block signature, a different delta token stream and a different matched/literal split than upstream — in both directions, on both transports.

## Measured

Against rsync 3.5.0 (protocol 32). 4096-byte file, basis differing only in bytes 0..49:

| | | `-B 2048` | no `-B` (control) |
|---|---|---|---|
| upstream 3.5.0 | local / ssh / daemon | 2048 / 2048 | 700 / 3396 |
| oc-rsync | local | 2048 / 2048 | 700 / 3396 |
| oc-rsync | ssh push, ssh pull | **700 / 3396** | 700 / 3396 |
| oc-rsync | daemon push | **700 / 3396** | 700 / 3396 |

oc's wire output *with* `-B 2048` was byte-for-byte its own *no*-`-B` output. The control column is what makes that unambiguous: identical numbers with and without the option mean the option was never applied, not that it happened to agree.

After the fix every cell matches upstream, including oc-client-to-upstream-daemon and upstream-client-to-oc-daemon.

## Why it happened: three decoders, three drops

The generator is the side that sizes the blocks it checksums (`generator.c:720-721`, `if (block_size) blength = block_size;`), so `-B` is only ever acted on by the receiving side. oc has three places that decode a peer's options into a receiving-side config, and the value was lost in all three:

1. **`crates/cli/src/frontend/server/run.rs`** — the `--server` argv parser captured `-B<digits>` into `ServerLongFlags::block_size` (so it would not leak into the positional path list) and then never read the field.
2. **`crates/core/src/client/remote/flags.rs`** — `block_size_override` never crossed from `ClientConfig` into the in-process `ServerConfig`, so a PULL (client is the receiver) had no value to apply either.
3. **`crates/daemon/.../client_args/long_form_args.rs`** — the daemon decodes the client argv through its own parser, which had no `-B` arm at all.

Plus one emission divergence found alongside them: the daemon arg vector emitted `--block-size=N`, an oc-only long spelling. Upstream emits the short `-B%u` (`options.c:2953-2954`), which the SSH builder already did. No upstream daemon parses the long form, so even a correct decoder would not have seen it.

## What the fix does *not* do

It does not add a second block-length rule. There is exactly one heuristic — `signature::block_size::calculate_block_length`, which already accepts the override — and the receiving side was reaching it through `SignatureLayoutParams::new(basis_size, None, ..)`. Carrying `block_size` on `ServerConfig` and threading it through `BasisFileConfig` replaces that hardcoded `None`. The bound check mirroring upstream's `parse_size_arg(.., 0, max_blength, False)` lives once, in `transfer::config::parse_block_size_arg`, and is called by both the `--server` parser and the daemon parser.

Adding the field to `BasisFileConfig` also made the compiler enumerate every construction site (four in `crates/transfer`, seven more in its integration tests) rather than leaving it to grep.

## A test pinned the defect

`block_size_forwarded_both_directions` asserted the daemon vector contained `--block-size=4096`. That assertion was the bug's monument: it locked in the spelling no peer parses. It now asserts `-B4096` **and** that the long form is absent, and is renamed accordingly.

## Tests

- `tests/block_size_wire.rs` runs the same option across local, SSH push and SSH pull **in one loop**. A per-path test would have passed on whichever path its author happened to pick — that is precisely how this survived. Reverting either wiring site independently turns it red (verified both ways).
- The no-`-B` case is the non-vacuity control (`assert_ne!` against the `-B 2048` numbers): 5 x 700 plus a 596-byte tail.
  It is **not** a short-final-block pin, though it looks like one — an earlier revision of this body claimed it was. The basis differs only in bytes 0..49, so the tail is preceded by a four-block match run and is reachable by adjacency continuation alone; a build that could only reach a short final block by adjacency would pass it identically. Pinning standalone-tail reachability needs a fixture sharing *only* the tail, which belongs with the tail-block work.
- A third case pins the upper edge: `-B` at or above the file length is one dirty block, so the delta degenerates to a whole-file send.
- Unit tests for the shared parser (protocol-dependent ceiling, `0` means heuristic, junk rejected) and for the generator actually *using* the value rather than merely storing it.
- Fixture bytes come from a seeded xorshift, not an arithmetic sequence: a short-period filler creates duplicate blocks, and with duplicates present the probe order decides which basis block a window resolves to, so two correct runs can legitimately disagree.

## Verification

```
cargo fmt --all -- --check                                                        clean
cargo clippy --locked --workspace --all-targets --all-features --no-deps -D warnings   clean
cargo nextest run --all-features -p transfer -p core -p cli -E '...'              243 passed
```

Plus the manual oracle matrix above against a real rsync 3.5.0 binary, on local, SSH (push and pull) and rsync:// daemon, at `-B` 512 / 700 / 2048 / 4096 and with no `-B`.

## Not covered here

The daemon end-to-end cell is verified by hand (script in the reproducer) but not yet in nextest — the suite has no reusable daemon harness for a delta-stats assertion. Follow-up task filed.

### DIFF
--- a/crates/cli/src/frontend/server/run.rs
+++ b/crates/cli/src/frontend/server/run.rs
@@ -672,6 +672,22 @@ fn apply_value_flags<Err: Write>(
         }
     }
 
+    // upstream: options.c:2953-2954 - the client re-emits `-B%u` to the server,
+    // and options.c:1795-1805 parses it back into the same `block_size` global
+    // the client used. Without this the server receiver's generator silently
+    // fell back to the square-root heuristic, so every `--block-size` transfer
+    // over the wire produced a different block signature and a different
+    // matched/literal split than upstream.
+    if let Some(size_str) = &long_flags.block_size {
+        match core::server::config::parse_block_size_arg(size_str, config.protocol) {
+            Ok(size) => config.block_size = size,
+            Err(msg) => {
+                write_server_error(stderr, brand, msg);
+                return Err(1);
+            }
+        }
+    }
+
     if let Some(size_str) = &long_flags.min_size {
         match parse_server_size_limit(size_str, "--min-size") {
             Ok(size) => config.file_selection.min_file_size = Some(size),
--- a/crates/core/src/client/remote/daemon_transfer/orchestration/arguments.rs
+++ b/crates/core/src/client/remote/daemon_transfer/orchestration/arguments.rs
@@ -408,12 +408,14 @@ pub(super) fn build_full_daemon_args(
         ));
     }
 
-    // upstream: options.c:2953-2957 - `-B%u` (block_size). oc mirrors the SSH
-    // builder's `--block-size=` spelling; both are accepted by the daemon's
-    // option parser and carry the identical value, so the remote receiver's
-    // generator sizes delta blocks exactly like the client requested.
+    // upstream: options.c:2953-2957 - `asprintf(&arg, "-B%u", (int)block_size)`
+    // inside `if (block_size) {`. The SHORT spelling is what upstream puts on
+    // the wire, so the daemon arg vector must carry it too: a `--block-size=`
+    // token is an oc-only spelling that no upstream daemon parses, and it left
+    // the remote generator sizing blocks by the square-root heuristic instead
+    // of the requested size.
     if let Some(bs) = config.block_size_override() {
-        args.push(format!("--block-size={}", bs.get()));
+        args.push(format!("-B{}", bs.get()));
     }
 
     // upstream: options.c:2793-2797 - --timeout=N so both peers enforce the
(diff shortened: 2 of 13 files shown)


Katomoto (commented):
Logic looks correct. One suggestion: the early return on line prevents the cleanup function from running — worth adding a finally block.

Deniwn22 (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

oferchen:
Rebased onto master (`545473f70`) to clear the `DIRTY` merge state.

**One conflict, in `daemon_transfer/orchestration/arguments.rs`** — both sides had added an `options.c` citation for the `-B` construct. Resolved against the pinned 3.5.0 source rather than by picking a side:

```
2953    if (block_size) {
2954        if (asprintf(&arg, "-B%u", (int)block_size) < 0)
2955            goto oom;
2956        args[ac++] = arg;
2957    }
```

So the construct is `2953-2957` and the `asprintf` is at `2954`. Kept master's range, this branch's explanation (it states *why* the short spelling is required, which is the actual defect), and **corrected the anchor text**: it read `asprintf(&arg, "-B%u", block_size)`, omitting the `(int)` cast, so it did not match the source verbatim and the drift audit could never have resolved it.

**⚠ One commit was dropped: `3f5c13bed` "docs(citations): retarget the -B citations at their real 3.4.4 lines".**

It was authored before the pin moved to 3.5.0, and its entire effect was `options.c:2953-2954 -> 2787-2790` — i.e. retargeting *to* 3.4.4. Under the current pin that is backwards and would have re-introduced the drift #7315 just removed. It was **comment-only** (28 changed lines, 0 non-comment), so nothing functional was lost.

It did contain one improvement worth preserving separately: it replaced the non-unique `if (am_sender)` anchor with prose. That anchor class is being handled properly in #7321, which re-anchors on `args[ac++] = "--super"` — a token unique in `options.c` — so the improvement is not lost, just relocated to where it is done thoroughly.

Verified on the rebased head: `cargo fmt --all -- --check` exit 0, `cargo check -p core --all-features` exit 0, zero remaining `options.c:2787` citations.

PR #7331, opened 2026-08-14
fix(citations): retarget three drifted upstream references to 3.5.0

oferchen:
## The gate is red on master, so every open PR inherits it

`Upstream citations track 3.5.0` fails on master HEAD and the two commits before it. Three per-crate breaches sit behind the one check name:

```
FAIL: core      suspected-drift rose 3 -> 4
FAIL: daemon    suspected-drift rose 2 -> 3
FAIL: protocol  suspected-drift rose 0 -> 1
```

**Attribution is settled without bisecting.** PR #7323 touches `crates/fast_io` and nothing else, the report shows `fast_io: suspected-drift=0 (0%)`, and it fails the gate anyway — a PR cannot inherit a red for a crate it does not touch unless the red is master's. Confirmed 4-for-4 across #7323/#7324/#7326/#7327.

## Resolutions

Each resolved against the pinned 3.5.0 source by **enclosing construct**, never by taking the first printed candidate.

| citation | → | evidence |
|---|---|---|
| `generator.c:298` | **304** | 298 is `if (DEBUG_GTE(DEL, 2))`. The cited `io_error & IOERR_GENERAL && !ignore_errors` is at 304. |
| `clientserver.c:1559` | **1738** | `return start_daemon(STDIN_FILENO, STDIN_FILENO);` is at 1738. Single candidate. |
| `flist.c:758` | **841** | 758 is an `always_checksum` block. Candidates were [463, 841, 1316]; 463 is `change_dir` failure, 1316 is the `cannot convert symlink data for` arm, 841 is the `cannot convert filename` arm at :843 — which is what this call site does. |

`flist.c:758 → 841` was reached independently by two earlier agents; this is a third route, by enclosing construct.

## Swept the file, not the flagged line

`generator.c:298` appears in **five** places, all citing the same construct — but the audit flags only the two in `flags.rs`, because the other three describe it in prose without quoting the anchor and are therefore invisible to the gate (task 672's 35%). Fixing only the flagged pair would have left three known-wrong neighbours.

`engine/` and `transfer/` already cited `:304`, so the tree was in the correct-but-inconsistent state task 388 describes. This converges it.

## No baseline lowering, deliberately

`core` returns to 3 and `daemon` to 2 — the accepted non-localisable counts recorded in the baseline's `_5_350_flip` note, not an improvement past them. `protocol` returns to its recorded 0. Lowering would be wrong here: the accepted entries are unchanged.

## Residual, stated rather than hidden

The `flist.c` anchor stays non-unique by construction — a comment quoting `io_error |= IOERR_GENERAL` cannot localise itself in a file containing three of them. Task 652 owns that class. The added note names the sibling arm so the next reader does not re-derive which one this is.

## Verification

- `citation_drift_audit.py --ratchet` → exit 0, `ratchet OK`
- `cargo fmt --all -- --check` → clean
- `cargo clippy --workspace --all-targets --all-features --no-deps -- -D warnings` → exit 0, zero errors

Comment-only change: 5 files, +12/−9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,30 @@ jobs:
         run: python3 tools/audit/unsafe_safety_comment_audit.py
 
   # ===========================================================================
+  # tools/ Unit Tests
+  # ---------------------------------------------------------------------------
+  # `tools/tests/` held unittest modules that no workflow ran, so a regression
+  # in the helpers they cover was invisible - including the coverage gate and
+  # the path-site sweep whose output scopes the confinement work. The runner
+  # discovers by pattern rather than an enumerated list, because an enumerated
+  # list is how those modules became orphans in the first place, and it asserts
+  # a non-zero test count, because `unittest discover` exits 0 when it matches
+  # nothing. Blocking, not informational: an inert gate is the defect it is
+  # meant to catch.
+  # ===========================================================================
+  tools-tests:
+    name: tools unit tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Run tools/tests
+        run: bash tools/ci/run_tools_tests.sh
+
+  # ===========================================================================
   # Test Job - Full test suite on Linux with toolchain matrix
   # ===========================================================================
   test:
--- a/crates/cli/src/frontend/arguments/parser/entry.rs
+++ b/crates/cli/src/frontend/arguments/parser/entry.rs
@@ -302,19 +302,14 @@ where
 
     let min_size = matches.remove_one::<OsString>("min-size");
     let max_size = matches.remove_one::<OsString>("max-size");
-    let block_size = match matches.remove_one::<OsString>("block-size") {
-        Some(value) => {
-            let s = value.to_string_lossy();
-            if s.parse::<u64>().is_err() {
-                return Err(clap::Error::raw(
-                    clap::error::ErrorKind::ValueValidation,
-                    format!("invalid --block-size value '{s}': must be a positive integer\n"),
-                ));
-            }
-            Some(value)
-        }
-        None => None,
-    };
+    // `--block-size` is validated where `--min-size` / `--max-size` are, by the
+    // shared suffix-aware parser (`parse_block_size_argument`). A bare-integer
+    // gate here would pre-empt it and make upstream's whole size grammar
+    // unreachable for this one option.
+    //
+    // upstream: options.c:1802 `parse_size_arg(arg, 'b', "block-size", 0,
+    // max_blength, False)` - the same parser the three siblings use.
+    let block_size = matches.remove_one::<OsString>("block-size");
 
     let rayon_threads = parse_thread_count(&mut matches, "rayon-threads")?;
     let tokio_threads = parse_thread_count(&mut matches, "tokio-threads")?;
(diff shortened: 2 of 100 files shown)


oferchen:
Reviewed against task 681's triage table. The citation choices are right:

- `generator.c:298 -> 304` and `clientserver.c:1559 -> 1738` are the two entries triage classified **uniquely resolvable**.
- `flist.c:758 -> 841` is one of the six multi-candidate entries, but it is the best-evidenced of them: reached independently by two agents by different methods, and `841` is present in the printed candidate set. Taking this one and leaving the other five is the correct cut.
- Sweeping every site of each citation rather than only the flagged line is right — a one-line patch leaves known-wrong neighbours behind.

**One thing to check before this merges: whether the baseline needs lowering.**

`tools/ci/citation_drift_baseline.json` is untouched here. Current values are `core: 3`, `daemon: 2`, `protocol: 0`, and master's gate reports `core 3 -> 4`, `daemon 2 -> 3`, `protocol 0 -> 1`.

If each fix removes exactly one drifted line, the counts return to baseline and the gate goes green with no baseline edit — correct as-is.

But the audit counts **drifted lines, not distinct citations**, and this PR changes `generator.c:298` at four sites. If those four sites were all of `core`'s drift, `core` drops to 0 against a baseline of 3, which prints:

```
note: core improved 3 -> 0; lower the baseline
```

and — per the `stale` branch — does **not** fail. That is exactly the unclaimed headroom `_2_lowering_is_mandatory` exists to prevent: the ratchet keeps accepting 3 and silently absorbs the next three regressions.

So please run the audit on the branch and paste the output. If it reports `improved` for any crate, lower that number in this same commit per `_2`. If every crate lands exactly on its baseline, say so and this is fine unchanged.

⚠ Do not lower `core` below the count that `_5_350_flip` describes as accepted false positives — those are non-localisable anchors, deliberately kept, and lowering past them would make the gate fail on the next unrelated edit.

oferchen:
**Overlap note for whoever merges second — the two PRs agree, so a conflict here is mechanical, not semantic.**

#7318 and #7331 both touch `crates/protocol/src/flist/read/name.rs`, and on the one shared line they make the **identical** change:

```
-  // upstream: flist.c:758 sets `io_error |= IOERR_GENERAL` with
+  // upstream: flist.c:841 sets `io_error |= IOERR_GENERAL` with
```

#7318 is a **superset** for this file — it sweeps every citation in it, not just the flagged one:

```
flist.c:768-772 -> 851-855      flist.c:731-734 -> 814-817
flist.c:757-764 -> 840-847      flist.c:759-762 -> 842-845
flist.c:758     -> 841          (the line #7331 also changes)
```

Two consequences:

1. **If a conflict appears in this file, resolve toward #7318's version.** It is strictly broader and the overlapping hunk is byte-identical, so nothing of #7331's is lost. #7331's other two retargets — `generator.c:298 -> 304` in `core` and `clientserver.c:1559 -> 1738` in `daemon` — are unique to it and unaffected.

2. **`flist.c:758 -> 841` now has four independent derivations**: two earlier agents, #7331's enclosing-construct reading (`841` is `cannot convert FILENAME`, against `463` = `change_dir` failure and `1316` = `cannot convert SYMLINK DATA`), and #7318's independent file sweep. For an anchor the audit cannot localise — `io_error |= IOERR_GENERAL` occurs three times in that file — that is about as settled as it gets.

⚠ Neither PR clears the `Upstream citations track 3.5.0` red on its own. The gate ratchets **per crate**, and master is currently breaching in three: `core 3->4`, `daemon 2->3`, `protocol 0->1`. #7318 addresses only `protocol`; #7331 addresses one line in each of the three. Do not read either PR's red as its own defect — see #7323, which touches only `crates/fast_io`, reports `fast_io: drift=0`, and fails the gate anyway.

PR #7333, opened 2026-08-14
fix(receiver): sanitize received symlink targets when munging is off

oferchen:
## The defect

A daemon module that sets `munge symlinks = false` gave up the `/rsyncd-munged/` prefix and got **nothing** in its place. A client holding only write access could push `escape -> ../outside`, have it stored verbatim, and reach outside the module root on the next transfer through that name — two transfers, no out-of-band step.

Upstream does not leave the target untouched. `flist.c:1182`:

```c
if (sanitize_paths && !munge_symlinks && *bp)
        sanitize_path(bp, bp, "", lastdir_depth, SP_DEFAULT);
```

and `clientserver.c:994-995` sets `sanitize_paths` for **every** daemon module. The two transforms are mutually exclusive and between them leave no daemon configuration in which a peer-supplied target reaches disk unchanged.

## Measured

Real `oc-rsync --daemon` vs real `rsync` daemons, macOS 15 aarch64, `use chroot = false`, `read only = false`, `munge symlinks = false`, client pushing `escape -> ../outside`:

| daemon | stored target | escaped |
|---|---|---|
| rsync 3.4.4 | `outside` (sanitized) | no |
| rsync 3.5.0 | `outside` (sanitized) | no |
| **oc-rsync** | `../outside` (verbatim) | **yes**, exit 0, stderr empty |

⚠ **Not the default config.** oc implements the `munge symlinks` `!use_chroot` auto default (`clientserver.c:997-998`) correctly — all three daemons produce `/rsyncd-munged/../outside` and no escape. The trigger is the operator turning munging off, which upstream treats as supported precisely *because* it still sanitizes.

## The change

`sanitize_received_symlink_target` in `crates/transfer/src/receiver/directory/links.rs`, gated exactly as upstream gates it: `!munge_symlinks && is_daemon_connection && non-empty`. `is_daemon_connection` is oc's expression of upstream's `sanitize_paths`, which is daemon-module-scoped.

Two decisions worth review:

**Placement — before the `--safe-links` check, not at creation.** Upstream sanitizes during the file-list decode, so by the time `generator.c:1559` evaluates `--safe-links` it reads the already-sanitized `F_SYMLINK(file)`. Sanitizing after the check would make oc skip with `skipping unsafe symlink` where upstream creates the link — trading this defect for a different divergence. Pinned by `sanitization_precedes_the_safe_links_check`.

**Byte-exact, not `&str`.** A symlink target is a raw byte string on the wire. The existing `sanitize_path(&str)` would replace non-UTF-8 bytes and rewrite the value the sanitizer exists to make safe, so this adds `sanitize_path_bytes_default` beside the existing `..._keep_dot_dirs_bytes`. Both delegate to the same byte core, so the traversal guard cannot diverge between them.

## Verification

TDD: both behavioural tests written and watched fail first.

```
RED  daemon_receiver_sanitizes_...    left: Some("../outside")  right: Some("outside")
RED  sanitization_precedes_...        left: None                right: Some("outside")
GREEN  5 tests run: 5 passed
```

The second RED is the ordering evidence — `None` because `--safe-links` skipped the unsanitized target entirely.

Mutation sweep, each anchor asserted to match exactly once before applying (a silently-unapplied mutation reads as a survivor):

| mutation | target test | result |
|---|---|---|
| drop the `is_daemon_connection` gate | `a_non_daemon_receiver_leaves_the_target_alone` | KILLED |
| drop the `!munge_symlinks` gate | `munging_still_wins_when_it_is_enabled` | KILLED |
| lossy `String` instead of bytes | `a_non_utf8_target_is_sanitized_byte_exactly` | KILLED |

- `cargo nextest run -p transfer --all-features`: **2731 run, 2729 passed, 2 failed**
- `cargo clippy --workspace --all-targets --all-features --no-deps -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean

⚠ The 2 failures are `sandbox::symlink_race_tests::anchored_mode_*`, **inherited from master**, not from this change — verified by running them on pristine `3f5a093fb` with none of this branch applied (2 passed, 2 failed, identical). They are the pair PR #7328 fixes.

## Scope

- **Unix only.** `create_symlinks` has three `cfg` implementations; this touches the unix one, where the defect is measured and where upstream's daemon semantic applies. The Windows sibling is untouched and unexamined — a Windows daemon with munging off is a separate question, deliberately not claimed here.
- **Linux not separately measured**, but this is a file-list string transform with no platform-conditional code and the tests are plain `#[cfg(unix)]` unit tests, so they execute on every unix CI cell.

Different layer from the receiver path-walk escape — file-list receive, not path resolution. Filed and fixed separately on purpose.

### DIFF
--- a/crates/transfer/src/receiver/directory/links.rs
+++ b/crates/transfer/src/receiver/directory/links.rs
@@ -74,6 +74,30 @@ impl ReceiverContext {
 
             let relative_path = entry.path();
 
+            // upstream: flist.c:1329 - `if (sanitize_paths && !munge_symlinks
+            // && *bp) sanitize_path(bp, bp, "", lastdir_depth, SP_DEFAULT)`.
+            // `sanitize_paths` is set for a daemon module serving a path
+            // (clientserver.c:1068), so turning munging off does not leave the
+            // received target untouched: it still cannot resolve above the
+            // module root. The two transforms are mutually exclusive upstream,
+            // which is why this runs only when munging is off.
+            //
+            // Applied *before* the safe-links evaluation because upstream
+            // sanitizes during the file-list decode, so the `--safe-links`
+            // check at generator.c:1951 reads the already-sanitized
+            // `F_SYMLINK(file)`. Sanitizing after the check instead would skip
+            // entries upstream creates.
+            let sanitized: PathBuf;
+            let wire_target = if !self.config.munge_symlinks
+                && self.config.connection.is_daemon_connection
+                && !wire_target.as_os_str().is_empty()
+            {
+                sanitized = sanitize_received_symlink_target(wire_target);
+                &sanitized
+            } else {
+                wire_target
+            };
+
             // upstream: generator.c:1951 - `if (safe_symlinks && unsafe_symlink(sl, fname))`
             // skips unsafe symlinks when --safe-links is set. The check stays
             // here (not in sanitize_file_list) to preserve protocol index
@@ -1018,6 +1042,28 @@ impl ReceiverContext {
 /// receives the ASCII prefix without any lossy decode. The matching strip
 /// happens on the sender via `strip_symlink_munge_prefix` in
 /// `crate::generator::file_list::entry`.
+/// Sanitizes a received symlink target so it cannot resolve above the module
+/// root, for daemon modules that disabled `munge symlinks`.
+///
+/// Mirrors upstream `flist.c:1182`, which applies `sanitize_path(...,
+/// SP_DEFAULT)` to the target whenever `sanitize_paths && !munge_symlinks`.
+/// `sanitize_paths` is set for every daemon module (`clientserver.c:994-995`),
+/// so the two transforms between them leave no daemon configuration in which a
+/// peer-supplied target reaches the disk verbatim.
+///
+/// Byte-level like the munge sibling above: a symlink target is a raw byte
+/// string on the wire, and decoding it through `String` would rewrite the very
+/// value being made safe.
+#[cfg(unix)]
+fn sanitize_received_symlink_target(target: &Path) -> std::path::PathBuf {
+    use std::ffi::OsString;
+    use std::os::unix::ffi::{OsStrExt, OsStringExt};
+
+    let sanitized =
+        crate::sanitize_path::sanitize_path_bytes_default(target.as_os_str().as_bytes());
+    std::path::PathBuf::from(OsString::from_vec(sanitized))
+}
+
 #[cfg(unix)]
 pub(in crate::receiver) fn apply_symlink_munge_prefix(target: &Path) -> std::path::PathBuf {
     use std::ffi::OsString;
--- a/crates/transfer/src/receiver/directory/links.rs
+++ b/crates/transfer/src/receiver/directory/links.rs
@@ -92,7 +92,7 @@ impl ReceiverContext {
                 && self.config.connection.is_daemon_connection
                 && !wire_target.as_os_str().is_empty()
             {
-                sanitized = sanitize_received_symlink_target(wire_target);
+                sanitized = sanitize_received_symlink_target(wire_target, relative_path);
                 &sanitized
             } else {
                 wire_target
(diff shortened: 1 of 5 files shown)


oferchen:
## Root cause of the `interop` red: the sanitize depth is 0, upstream passes `lastdir_depth`

The `interop / interop with upstream rsync` failure is caused by this branch, not inherited. Measured by diffing the failure list against a passing PR on the same base (#7332): the delta is exactly six lines, one cause, every protocol version.

```
UNEXPECTED FAIL: up:safe-links (fp=28)
UNEXPECTED FAIL: up:safe-links (fp=29)
UNEXPECTED FAIL: up:safe-links (fp=30)
UNEXPECTED FAIL: up:safe-links (fp=31)
UNEXPECTED FAIL: up:safe-links (fp=32)
UNEXPECTED FAIL: up:safe-links (fp=native)
```

The 53 `fp=28` entries in the log are present in the passing run too and are pre-existing, non-gating.

Harness detail:

```
[upstream 3.4.4→oc] safe-links
  --safe-links: unsafe symlink was transferred
```

### The defect

`sanitize_received_symlink_target` calls `sanitize_path_bytes_default`, i.e. depth **0**. The upstream line this change cites passes **`lastdir_depth`**:

```c
3.4.4  flist.c:1182-1183   if (sanitize_paths && !munge_symlinks && *bp)
                                   sanitize_path(bp, bp, "", lastdir_depth, SP_DEFAULT);
3.5.0  flist.c:1328-1329   (identical)
```

This is the *only* `sanitize_path` call in `flist.c` that does not pass 0 — the other four (`:857`, `:1425`, `:2571`, and the 3.4.4 equivalents) all pass 0. The non-zero depth there is deliberate.

`depth` is the budget of leading `..` allowed to survive. At `lastdir_depth`, a symlink living in `sub/dir/` keeps the `../..` that legitimately walks back to the transfer root, so a target that escapes *above* the root still reads as escaping when `--safe-links` evaluates it. At depth 0 every leading `..` is dropped, so `../../etc/passwd` collapses to `etc/passwd` — now in-tree and safe-looking — and the safe-links check at `generator.c:1547` creates the link instead of skipping it.

So the ordering argument in the new comment is right (upstream does sanitize before the safe-links check), but it only holds together with upstream's depth. With depth 0 the sanitize *launders* the unsafe target rather than confining it.

### Fix direction

Thread the entry's containing-directory depth through instead of using the depth-0 wrapper. `sanitize_path_with_depth` already takes the parameter; only the new `_default` entry point hard-codes 0.

### Test gap

The suite in `sanitize_symlink_targets.rs` passes as-is, so it does not currently cover the case that matters: a symlink **in a subdirectory** whose target escapes above the transfer root, asserted to be *skipped* under `--safe-links`. Worth adding both arms — the in-subdirectory relative target that must survive, and the escaping one that must not — since depth 0 and `lastdir_depth` agree at the root and only diverge below it. That is why every existing test passed.

### Related

This interacts with the already-recorded divergence that under `munge symlinks` + `--safe-links` oc creates links both upstreams skip. This change widens that class rather than narrowing it, so the two are worth settling together.

oferchen:
## Correction: my earlier root-cause for the `interop` red was wrong

I attributed the `up:safe-links` failure to the sanitize depth. That is wrong, and the way it is wrong matters, so: the failing fixture is

```sh
ln -sf /etc/hostname "$sdir/unsafe_link.txt"
```

an **absolute** target at the **transfer root**. `sanitize_path` strips the leading `/` unconditionally; there is no `..` in it for a depth budget to apply to. Depth 0 and `lastdir_depth` produce byte-identical output here, so the depth cannot explain the failure. I had a confirmed mechanism and assumed it applied to the path I was looking at.

Pushed in f8d7981 are two separate things, each measured.

### 1. The depth defect is real, just not this failure

Upstream passes `lastdir_depth` (`flist.c:1329`, from `flist.c:867`), the only `sanitize_path` call in `flist.c` that does not pass 0. Measured against a real rsync 3.5.0 daemon receiver under `use chroot = false` + `munge symlinks = false`, pushing two links from `sub/dir/`:

```
../../hello.txt     -> ../../hello.txt    (in budget, kept intact)
../../../escape.txt -> ../../escape.txt   (one level over, collapsed)
```

At depth 0 those become `hello.txt` and `escape.txt` - the first silently retargeted at a sibling that does not exist. Every existing test in the module placed the link at the transfer root, where the budget is 0 and both readings agree, which is why all of them passed either way. The two new arms sit in a subdirectory and cover both sides of the budget. Reverting the depth to 0 gives `7 tests run: 5 passed, 2 failed` - exactly the new pair.

### 2. The interop cell's assertion is not upstream-faithful

The cell required `unsafe_link.txt` to be **absent**. Under the config the harness itself writes (`write_rust_daemon_conf` / `write_upstream_conf`: `use chroot = false`, `munge symlinks = false`), upstream **keeps** it. Measured, both releases, as daemon receiver:

```
rsync 3.4.4  ->  unsafe_link.txt -> etc/hostname   CREATED
rsync 3.5.0  ->  unsafe_link.txt -> etc/hostname   CREATED
```

`sanitize_paths` is on (`clientserver.c:1068`), munging is off, so `flist.c:1329` rewrites `/etc/hostname` to the relative in-tree `etc/hostname`, and `--safe-links` (`generator.c:1951`) then correctly does not skip it. So this branch made the receiver match upstream and the cell's expectation is what diverges. With upstream's own default (`munge symlinks` unset, which resolves to on for a module with a path) the target instead gains the `/rsyncd-munged/` prefix, stays absolute, and is skipped - hence the absence the cell assumed.

The check now asserts what `--safe-links` plus the sanitize actually guarantee, over every symlink in the destination rather than one name: none keeps an absolute target, and none resolves above the destination root. That holds under both module configurations.

### Noted separately, not changed here

Direction 2 (`oc:safe-links`, oc client -> upstream daemon) passes, but not for the reason the cell implies: oc's **sender** drops the unsafe link, so the upstream receiver never sees it. Upstream's `safe_symlinks` is receiver-only - the only uses are `generator.c:1951` and `backup.c` - so the sender-side filtering is an oc divergence. It is pre-existing and untouched by this branch; filing it on its own rather than widening this one.

PR #7386, opened 2026-08-19
fix(output): print `skipping directory` at default verbosity

oferchen:
Upstream gates the notice on `S_ISDIR(st.st_mode) && !xfer_dirs` and
nothing else - there is no `INFO_GTE` and no verbosity test at either
call site (flist.c:1338 in send_file_name, flist.c:2452 in
send_file_list). It therefore prints at DEFAULT verbosity, and the only
thing that suppresses it is `quiet` inside `rwrite()` (log.c:345).

oc emitted it from the verbose renderer instead. That renderer only ever
runs over a collected event log, and collection is gated on
`force_event_collection || verbosity > 0 || progress || list_only`
(core: client/config/client/output.rs `collect_events`). At the default
every disjunct is false, so `LocalCopyContext::events` is `None`,
`record()` never stores the entry, and the renderer iterates an empty
slice. The notice was reachable only at `-v` and above - the exact
inverse of upstream, which prints at the default and hides it under -q.

Move the emission to the decision site in `record_skipped_directory`,
and make the renderer arm a `continue`. This is the shape already used
for `skipping non-regular file` in the same file, whose doc comment
records the identical failure mode. `--quiet` still suppresses the line,
via `logging::message_stream`, which owns upstream's
`case FINFO: if (quiet) return;` rule.

`--quiet` reaches that funnel for the first time here: the parser folds
`-q` into `verbosity = 0`, which cannot express "default verbosity but
quiet", and the renderer's `StreamContext` had `quiet` hardcoded to
`false`. Give quiet its own thread-local, mirroring upstream's global,
and feed it to the funnel.

`record_skipped_directory` is the single owner: all three producers
(sources/handlers.rs twice, sources/orchestration.rs) call it, and
`ClientEventKind::SkippedDirectory` is only ever built from
`LocalCopyAction::SkippedDirectory`, so the renderer arm served local
copy alone and removing it cannot silence a network path.

Tests count occurrences rather than asserting containment, since
containment holds for one copy and for five. `--info=name0` is the
discriminating cell: it disables NAME output while upstream still
prints. The `-q` and `-v` cells are the non-vacuity companions - without
`-q` the pin would also pass on a renderer that printed
unconditionally, and `-v` is the only cell that catches a re-introduced
duplicate.

### DIFF
--- a/crates/cli/src/frontend/execution/drive/workflow/run.rs
+++ b/crates/cli/src/frontend/execution/drive/workflow/run.rs
@@ -282,6 +282,13 @@ where
 
     let verbosity_config = VerbosityConfig::from_verbose_level(verbosity);
     logging::init(verbosity_config);
+    // The parser already folded `--quiet` into `verbosity = 0`, which is what
+    // silences every verbosity-gated line. Upstream additionally keeps `quiet`
+    // as its own global and consults it in the `FINFO` arm of `rwrite()`
+    // (log.c:344-345), so a notice it prints at DEFAULT verbosity - `skipping
+    // directory %s` (flist.c:1484) - still disappears under `-q`. Carrying the
+    // flag past the parser is what makes those two states distinguishable.
+    logging::set_quiet(quiet);
 
     let rayon_thread_count = rayon_threads.and_then(|n| NonZeroUsize::new(n as usize));
     let tokio_thread_count = tokio_threads.and_then(|n| NonZeroUsize::new(n as usize));
--- a/crates/cli/src/frontend/progress/diagnostic.rs
+++ b/crates/cli/src/frontend/progress/diagnostic.rs
@@ -79,11 +79,15 @@ pub fn render_diagnostic_events<O: Write, E: Write>(
         // rule; a second copy here is what made #352 and #357 two separate
         // bugs instead of one.
         let ctx = logging::StreamContext {
-            // This renderer runs after the CLI has already folded `--quiet`
-            // into the verbosity level, so the events that reach it are
-            // pre-filtered; passing `quiet: false` keeps that behaviour
-            // unchanged. Wiring the real flag through is task #346.
-            quiet: false,
+            // upstream: log.c:345 `if (quiet)` - the FINFO arm of rwrite()
+            // returns without writing when quiet is set. Folding
+            // `--quiet` into `verbose = 0` at parse time silences only the
+            // verbosity-gated events; a notice upstream prints at DEFAULT
+            // verbosity - `skipping non-regular file "%s"` under
+            // `INFO_GTE(NONREG, 1)`, which `info_verbosity[0]` enables - still
+            // reaches this renderer and must be suppressed here, where
+            // upstream suppresses it.
+            quiet: logging::finfo_suppressed(),
             // `Never` and `Default` select the same stream (log.c:253 keys on
             // `== 1`), so a caller that only knows the boolean loses nothing
             // here. The tri-state is available for callers that have it.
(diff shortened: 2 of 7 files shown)


oferchen:
Marking draft: CI found a real defect in this change, root-caused below.

## What failed

`cli::relative_dot_root_skipped_directory::leading_dot_operand_without_xfer_dirs_reports_skipping_directory_dot`, deterministically on nextest/macOS/musl across all three toolchains (1 failed / 6943 passed; the other 8 red cells are the same single test).

```
assertion `left == right` failed
  left: Some("sub/")
 right: Some("skipping directory .")
```

with actual stdout:

```
sub/
sub/f.txt

sent 3 bytes  received 0 bytes  6.00 bytes/sec
total size is 3  speedup is 1.00
skipping directory .        <- must be the FIRST line
```

## Root cause

The change is right about **when** (default verbosity) and wrong about **where in the stream**.

oc has two independent output paths:

1. the engine's `LocalCopyRecord` log, rendered in scan order — correctly interleaved, but only *collected* when `verbosity > 0 || progress || list_only`;
2. `logging::emit_info_coded`, which pushes onto a thread-local `EVENTS` buffer drained at the very end of the run (`progress/diagnostic.rs`, `execution/drive/summary.rs`).

Moving the notice from (1) to (2) bought default-verbosity reachability and lost stream position. Upstream has one funnel — `rprintf(FINFO, ...)` writes at the decision point — so it lands before any name row.

## Why this is not a quick fix

Getting both properties needs the info channel to emit in stream order, which is the open architectural item (one `rwrite()` logcode-to-stream funnel). Reordering the global drain to satisfy this one notice would move every other info message too, unmeasured. That is a bigger, separately-verified change than this PR.

Parking here rather than shipping an ordering hack. The upstream analysis in the commit message stands and is worth keeping: the gate is `S_ISDIR && !xfer_dirs` with no `INFO_GTE`, and the `collect_events` finding explains why the renderer arm was unreachable at the default.

oferchen:
## Root cause of the 13 red checks: a flush-order defect, not a stream mix-up

Measured on this branch (macOS, `-Rv ./sub/f.txt` from inside `src/`, streams
separated):

```
--- STDOUT ---
sub/
sub/f.txt

sent 3 bytes  received 0 bytes  6.00 bytes/sec
total size is 3  speedup is 1.00
skipping directory .
--- STDERR ---
(empty)
```

The notice **is** on stdout and **is** produced. It lands *last*, after the
summary, so `lines.first()` in
`crates/cli/tests/relative_dot_root_skipped_directory.rs:84` sees `sub/`. That
test passes on master, so this is a regression, not a pre-existing red.

Mechanism: `info_log!` at the new emission site enqueues into the thread-local
diagnostic queue, and that queue is drained by `flush_diagnostics` at
`crates/cli/src/frontend/mod.rs:368` / `:384` - *after* `execute()` returns, and
therefore after the local-copy renderer has already written its rows and the
summary. Upstream has no queue: `rprintf(FINFO, ...)` inside `make_file()`
writes synchronously during the flist scan, which precedes every name row.

## Why the obvious fix is wrong

Draining the queue once before the renderer would put this notice first, but it
would also hoist **every** other queued info event to the front - including the
`--info=backup` notice, which upstream prints interleaved per file (the comment
at `mod.rs:360-366` records that this channel already carries it). That trades
one ordering divergence for another.

## What it actually needs

Interleaving by production order, i.e. a sequence stamp on each event plus a
merge at render time. That is exactly the `feat/logcode-funnel` work
(`Stamped`/`Sequence` in `crates/logging`, `PendingDiagnostics` in
`cli/frontend/progress/interleave.rs`). It was parked as having no consumer.
This PR is the consumer.

Recommendation: hold this PR until the funnel lands, then re-measure. The
gating analysis in the description is correct and worth keeping - it is only
the placement of the emission that is unresolved.

oferchen:
Superseded by #7433, which carries this commit (authorship preserved) on top of the deferred-diagnostic funnel that supplies the missing ordering.

Measured there against the real rsync 3.5.0 binary: `-Rv` / `-R` / `-Rq` all three cells now byte-match upstream, and `relative_dot_root_skipped_directory.rs` passes again.

Leaving this open rather than closing it — that call is the author's.

oferchen:
Closing as superseded by #7433, which reaches the same goal by the mechanism that works.

**Measured, not assumed:**

- All 13 red checks here are the *same* test:
  `leading_dot_operand_without_xfer_dirs_reports_skipping_directory_dot`
  (`crates/cli/tests/relative_dot_root_skipped_directory.rs:84`)
  ```
  assertion `left == right` failed: rsync 3.4.4 leads with "skipping directory ."
  for `-Rv ./sub/f.txt`, got: sub/
  ```
- That test **exists on master and passes there**, and this branch does not touch
  it — so this branch regresses master rather than merely being blocked.
- The whole test contribution of this PR,
  `crates/cli/src/frontend/tests/transfer_request_skips_directories_without_recursion.rs`,
  is present on #7433 **byte-identical** (8540 bytes on both branches).
- #7433 is `SUCCESS` on nextest stable, beta and nightly — it delivers the
  behaviour without the regression. Its only two red checks are the upstream
  testsuite legs, and those are a separate, already-diagnosed master defect fixed
  by #7435.

The two branches attack the same divergence differently: this one edits
`diagnostic.rs`/`render.rs` directly, #7433 routes the notices through the
log-code sequence funnel. Keeping both open would duplicate the work and leave
the regressing variant available to merge.

Reopen if the funnel approach is later abandoned; nothing here is lost.

PR #7389, opened 2026-08-19
fix(security): confine the commit rename and the backup path against symlink races

oferchen:
The local-copy commit rename was a bare `fs::rename` on Unix. A destination
parent directory flipped to a symlink between temp-create and commit therefore
redirected the transferred file outside the destination tree.

Upstream 3.5.0 ships a test for exactly this, and it reproduces **deterministically**
against oc-rsync (Linux aarch64, release build, `runtests.py --rsync-bin=<oc-rsync>`):

```
rename-fullpath symlink race: files were written OUTSIDE the destination tree
(['f10', 'f12', 'f13', 'f2', 'f20', 'f21', 'f22', 'f24', 'f29', 'f34', 'f4', 'f6'])
FAIL    rename-fullpath-symlink-race
```

12 payload files landed in the attacker's `outside/` tree.

## Root cause

`engine/src/local_copy/executor/file/guard.rs::hardened_rename` hardened the
Windows arm (`rename_no_follow`, handle-anchored) and left the Unix arm as
`fs::rename` / `try_rename_via_io_uring` - no dirfd anchoring at all.

The sibling `transfer` disk-commit path *does* anchor, but all-or-nothing: its
helper drops to a path-based rename unless **both** endpoints anchor. That is
the rule upstream 3.5.0 deliberately replaced, and its comment says why:

> Confine each side independently. [...] Doing each side independently means an
> absolute source never disables confinement of a relative destination.
> - `rsync-3.5.0/syscall.c:1866` `do_rename_at()`

An absolute `--temp-dir` is exactly what puts the source outside the tree, which
is how the escape becomes reachable.

## Change

- `fast_io`: new `confined_rename(root, old, new, replace)`. Each endpoint that
  lies beneath `root` has its parent resolved by
  `DirSandbox::open_dest_anchor_confined` - the per-component walk that follows
  relative in-tree symlinks and refuses absolute targets and climbs above the
  anchor. An endpoint outside `root` keeps ambient `AT_FDCWD` resolution.
  Refusals surface as `ELOOP`, deliberately **not** `EXDEV`: callers treat
  `EXDEV` as cross-device and fall back to copy+remove, which would defeat the
  refusal.
- `engine`: `hardened_rename` takes the destination root and routes through
  `confined_rename` when it has one. `DestinationWriteGuard::new_confined` is the
  constructor the production copy paths use; the 4-arg `new` remains for tests
  and documents that its commit is unconfined.

## Tests

`crates/fast_io/tests/confined_rename.rs`, three cells:

- `confined_rename_commits_a_nested_destination` - the non-vacuity companion.
  Without it the two refusal tests would also pass if `confined_rename` simply
  failed for every input.
- `confined_rename_refuses_a_destination_parent_symlinked_outside`.
- `an_out_of_tree_source_does_not_disable_destination_confinement` - the
  absolute-`--temp-dir` shape, i.e. the per-side-independence rule itself.

## Verification (Linux aarch64)

- Upstream 3.5.0 `rename-fullpath-symlink-race`: **FAIL before, PASS after**.
  The RED side is measured on a release binary built from the parent commit, not
  assumed.
- Full 3.5.0 suite, 345 tests, diffed per test against the outcome manifest on
  `chore/upstream-testsuite-350`. Four cells differ; only one is this change:

  | test | manifest | this branch | attribution |
  |---|---|---|---|
  | `rename-fullpath-symlink-race` | fail | pass | **this fix** (control fails 3/3) |
  | `rrsync-no-overwrite-delay-updates` | fail | pass | not this fix - passes 2/2 on the control too |
  | `simd-checksum` | pass | skip | host (aarch64) |
  | `symlink-race-source` | pass | fail | not this fix - fails 3/3 on the control too; a pre-existing **sender**-side escape the manifest recorded as a lucky pass |

- `cargo nextest run -p engine -p fast_io --all-features`: 6128 passed, 0 failed.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets
  --all-features --no-deps -- -D warnings` clean.

## Deliberately out of scope

- The anonymous `O_TMPFILE` commit path materialises with `linkat(2)` against the
  final path rather than a rename, so it does not pass through `hardened_rename`.
  Confining that sink belongs with the rest of the leaf-sink sweep.
- The ambient arm for an endpoint outside the root keeps plain path resolution.
  Upstream additionally runs an ownership walk over an operator-supplied absolute
  side; that is a separate hardening, not part of the escape fixed here.

---

# Second fix in this branch: the backup path (`backup-dir-symlink-race`)

The same suite run surfaces a second escape, in `make_backup` rather than the
commit rename, and it needs the **other** upstream resolver.

```
backup-dir parent symlink race: a 0777 dest file was backed up into .../outside
through the flipped backup/sub symlink -- make_backup followed a foreign-owned
parent component (escape).
FAIL    backup-dir-symlink-race
```

## Why the confined walk is the wrong tool here

A `--backup-dir` is an **operator** path: the person running rsync named it, and
it may legitimately sit outside the transfer tree. Confining it beneath a root
would break the ordinary out-of-tree backup directory. Upstream's answer is a
second resolver whose trust signal is **authority**, not location:

> An operator path may legitimately point outside the tree, so the trust signal
> is authority (ownership), not location.
> - `rsync-3.5.0/syscall.c:544-551`

Per component: `statat(AT_SYMLINK_NOFOLLOW)`; a symlink whose `st_uid` is neither
0 nor our euid is refused with `ELOOP` (`syscall.c:406`); a trusted one is
followed by splicing its target into the remaining path, an absolute target
restarting at `/`, bounded at 40 hops (`syscall.c:361`).

## Change

- `fast_io::owner_walk` - new module: `owner_trusted_parent`, plus
  `operator_rename` / `operator_link` built on it, mirroring `do_rename_at`
  (`syscall.c:1894`) and `do_link_at` (`syscall.c:676`) under
  `operator_path_resolve`.
- `engine`: `backup_rename` routes through `operator_rename`, and a new
  `create_backup_hard_link` routes the backup link tier through `operator_link`.

## The link tier is why the first attempt did not work

Routing only `backup_rename` left the escape fully reproducible. Upstream's
`link_or_rename` (`backup.c:200-207`) tries `do_link_at` **first**; oc has the
same two-tier shape at `context_impl/backup.rs:129`, so the hard link - not the
rename - was the escaping sink. Both tiers are now resolved.

`create_backup_hard_link` is deliberately separate from `create_hard_link`: the
latter also serves `-H`, a transfer path that takes the *confined* resolver, not
this one. Folding them would apply the wrong policy to one of the two callers.

## Verification (Linux aarch64, as root)

Full 3.5.0 suite, 345 tests, **against a control binary built without either fix,
same host, same run conditions** - so attribution is measured, not inferred:

| test | control | this branch |
|---|---|---|
| `rename-fullpath-symlink-race` | fail | **pass** |
| `backup-dir-symlink-race` | fail | **pass** |
| `nondaemon-symlink-race` | fail | **pass** |
| `symlink-dest-backupdir` | fail | **pass** |

Four cells moved, all in the same direction, and nothing else in the suite
changed: 94 failures -> 90, 200 passes, 55 skips both runs. The two extra passes
are the same backup-symlink family reached by different fixtures.

- `cargo nextest run -p engine -p fast_io --all-features`: 6128 passed, 0 failed.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets
  --all-features --no-deps -- -D warnings` clean.

### DIFF
--- a/crates/engine/src/local_copy/context.rs
+++ b/crates/engine/src/local_copy/context.rs
@@ -60,7 +60,7 @@ use filters::FilterRule;
 use logging::info_log;
 use protocol::flist::FileListWriter;
 
-use super::overrides::{backup_rename, create_backup_hard_link, create_hard_link};
+use super::overrides::{backup_rename, create_backup_hard_link};
 #[cfg(target_os = "linux")]
 use super::overrides::{cached_parent_device, same_filesystem};
 
--- a/crates/engine/src/local_copy/context.rs
+++ b/crates/engine/src/local_copy/context.rs
@@ -60,7 +60,7 @@ use filters::FilterRule;
 use logging::info_log;
 use protocol::flist::FileListWriter;
 
-use super::overrides::{backup_rename, create_hard_link};
+use super::overrides::{backup_rename, create_backup_hard_link, create_hard_link};
 #[cfg(target_os = "linux")]
 use super::overrides::{cached_parent_device, same_filesystem};
 
--- a/crates/engine/src/local_copy/context_impl/backup.rs
+++ b/crates/engine/src/local_copy/context_impl/backup.rs
@@ -126,14 +126,14 @@ impl<'a> CopyContext<'a> {
         let hard_linked = if prefer_rename {
             None
         } else {
-            match create_hard_link(destination, &backup_path) {
+            match create_backup_hard_link(destination, &backup_path) {
                 Ok(()) => Some(BackupStrategy::HardLink),
                 Err(error) if error.kind() == io::ErrorKind::NotFound => return Ok(()),
                 // upstream: backup.c:247-256 - a stale backup entry is
                 // removed and the hard link retried once.
                 Err(error) if is_stale_backup_conflict(&error) => {
                     remove_stale_backup_entry(&backup_path)?;
-                    match create_hard_link(destination, &backup_path) {
+                    match create_backup_hard_link(destination, &backup_path) {
                         Ok(()) => Some(BackupStrategy::HardLink),
                         Err(_) => None,
                     }
--- a/crates/engine/src/local_copy/context_impl/state.rs
+++ b/crates/engine/src/local_copy/context_impl/state.rs
@@ -197,7 +197,10 @@ impl<'a> CopyContext<'a> {
     }
 
     /// Returns the root destination directory for the transfer.
-    pub(super) fn destination_root(&self) -> &Path {
+    ///
+    /// Also the confinement anchor for the temp -> final commit rename; see
+    /// `DestinationWriteGuard::new_confined`.
+    pub(in crate::local_copy) fn destination_root(&self) -> &Path {
         &self.destination_root
     }
 
--- a/crates/engine/src/local_copy/executor/file/copy/transfer/execute/clonefile.rs
+++ b/crates/engine/src/local_copy/executor/file/copy/transfer/execute/clonefile.rs
@@ -45,7 +45,15 @@ pub(super) fn eligible(
         ..
     } = flags;
 
-    let transfer_ok = existing_metadata.is_none()
+    // The injected `PlatformCopy` still decides whether this tier runs at all:
+    // `NoReflinkPlatformCopy` (and any test fake) reports no reflink support and
+    // vetoes the fast path, leaving the standard copy to do the work. The
+    // *destination write* no longer goes through the strategy - a path-based
+    // `copy_file(src, dst)` cannot express an anchored destination, which is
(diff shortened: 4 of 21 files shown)


oferchen:
Superseded by #7393, which combines this with its counterpart. The two halves of the 3.5.0 confinement model cannot pass the required `upstream testsuite` contexts separately - each branch was red on exactly the test the other one fixed. Every commit here is merged into `fix/confined-symlink-race-cluster`; the branch is retained.

PR #7390, opened 2026-08-19
fix(security): confine the local-copy source read beneath the transfer root

oferchen:
The local-copy source read was a bare `fs::File::open` - no `O_NOFOLLOW` on the
leaf and no confinement of the parent components. A source directory flipped to
a symlink between scan and read therefore redirected the read outside the
transfer tree.

Upstream 3.5.0 ships a test for exactly this, and it reproduces
**deterministically** against oc-rsync (Linux aarch64, release build):

```
source-tree TOCTOU: content from outside the source tree
('OUTSIDE-SECRET-do-not-copy') was copied into the output -- the sender
followed a parent-component symlink it should have confined.
FAIL    symlink-race-source
```

## Root cause

`engine/.../copy/transfer/open.rs::open_source_file` opened the source with
`fs::File::open` (or the `O_NOATIME` variant). Two gaps against upstream, not
one: no `O_NOFOLLOW` leaf (upstream's baseline since <=3.4.4), and no confined
ancestor walk (upstream 3.5.0's new fix, which is what this test exercises - a
**parent** component flip, not a leaf flip).

The confined open already existed and was already tested
(`fast_io::open_source_confined`), with exactly one consumer whose anchor is
`Some(..)` only for a non-chroot **daemon** sender. The local-copy path never
went through it.

## Change

- `CopyContext` carries a `source_anchor`, set per operand in the source loop.
  The anchor is the directory the sender would `chdir` into: the operand itself
  for a trailing-slash (contents) operand, its parent otherwise.
- `open_source_file` takes that anchor plus the follow-symlinks decision and
  mirrors upstream's three arms.

## The gate is the substance, and my first version had it wrong

Upstream applies the confinement **only** when no symlink-following mode is
active:

```c
} else if (!copy_links && !copy_unsafe_links && !copy_dirlinks && !insecure_links) {
        fd = sender_open_confined(NULL, fname, O_RDONLY);   /* sender.c:685-705 */
} else {
        fd = do_open_checklinks(fname);                      /* sender.c:706 */
}
```

Confining the following modes breaks them outright - `--copy-unsafe-links`
exists precisely to materialise links whose targets lie OUTSIDE the tree. Two
existing engine tests caught this before it shipped:
`execute_with_copy_unsafe_links_materialises_file_target` and a `-k`
dir-link cell. All three options are now in the predicate.

Note the anchor question too: upstream passes `NULL` (relative to cwd) and
chdirs to the transfer root first, so "cwd" and "transfer root" name the same
directory - the anchor here is upstream's, not an invention.

⚠ The confined/copy-links pairing upstream also keeps
(`sender_open_copylinks_confined`, `sender.c:682`) is the **daemon** branch,
where the anchor is the module root and leaving it is a module escape. The
non-daemon sender this module implements takes the plain branch.
`--insecure-links` is upstream's fourth exclusion term; oc does not implement
that option yet, so it cannot be spelled in the predicate.

## Verification (Linux aarch64)

- Upstream 3.5.0 `symlink-race-source`: **3/3 FAIL before, 3/3 PASS after**.
  The test is a race, so both sides are measured repeatedly rather than once.
- Full 3.5.0 suite, 345 tests, diffed per test against a control built from
  this branch's parent commit. Exactly one cell differs -
  `symlink-race-source`, fail -> pass. Nothing else moved.

  ⚠ My first control was **invalid** and briefly showed a second diff
  (`rrsync-no-overwrite-delay-updates`, pass -> fail). That binary was built
  from a *different* branch carrying an unmerged operand fix. Rebuilt from the
  real parent, that cell fails 3/3 there too - it is a pre-existing failure,
  not a regression.
- Three new unit cells in `open.rs`: an in-tree nested source still opens (the
  non-vacuity companion), a parent symlinked outside the root is refused, and a
  symlink-following mode still reads through the same escape - the cell that
  discriminates the gate from a blanket confinement.
- `cargo nextest run -p engine --all-features`: 5089 passed, 0 failed.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets
  --all-features --no-deps -- -D warnings` clean.

## Deliberately out of scope

The alt-base override open (`--copy-dest` / `--link-dest` after a cross-device
degrade) is an operator-named path outside the source tree. It takes the
operator-path resolver, not the transfer-root confinement, so it is passed no
anchor here; that routing belongs with the alt-dest family.

### DIFF
--- a/crates/engine/src/local_copy/context.rs
+++ b/crates/engine/src/local_copy/context.rs
@@ -145,6 +145,16 @@ pub(crate) struct CopyContext<'a> {
     stop_at: Option<SystemTime>,
     last_progress: Instant,
     destination_root: PathBuf,
+    /// Confinement anchor for source-file content reads: the directory the
+    /// sender would `chdir` into for the operand currently being walked.
+    ///
+    /// `None` before the first operand and for any operand whose anchor cannot
+    /// be determined, in which case the open degrades to the `O_NOFOLLOW` leaf
+    /// rule alone. Set per operand because each source argument has its own
+    /// transfer root. upstream: `rsync-3.5.0/sender.c` opens file content
+    /// through the confined held-ancestor-dirfd stack anchored at the transfer
+    /// root, so a parent component flipped to a symlink is refused.
+    source_anchor: Option<PathBuf>,
     /// Number of leading path components in `relative` that represent the
     /// transfer root name (e.g. the source directory name when copying
     /// without a trailing slash).  These components inflate the depth
--- a/crates/engine/src/local_copy/context_impl/link_dest.rs
+++ b/crates/engine/src/local_copy/context_impl/link_dest.rs
@@ -40,7 +40,13 @@ impl<'a> CopyContext<'a> {
         let mut best: Option<(PathBuf, u8)> = None;
         for entry in self.options.link_dest_entries() {
             let candidate = entry.resolve(self.destination_root(), relative);
-            let candidate_metadata = match fs::metadata(&candidate) {
+            // upstream: generator.c try_dests_reg() -
+            // `if (basis_link_stat(cmpbuf, &sxp->st) < 0 || !S_ISREG(sxp->st.st_mode)) continue;`
+            // basis_link_stat resolves the leaf with `link_stat_at(dfd, leaf, stp, 0)`, i.e. an
+            // LSTAT, so a basis entry that is itself a symlink reports S_IFLNK and is skipped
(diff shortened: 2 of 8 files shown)


oferchen:
Superseded by #7393, which combines this with its counterpart. The two halves of the 3.5.0 confinement model cannot pass the required `upstream testsuite` contexts separately - each branch was red on exactly the test the other one fixed. Every commit here is merged into `fix/confined-symlink-race-cluster`; the branch is retained.

PR #7394, opened 2026-08-20
fix(daemon): refuse a client-sent --insecure-links

oferchen:
## What

A daemon must refuse a client-sent `--insecure-links`.

`--insecure-links` is a **local-only** opt-out of symlink confinement. A client
must never be able to switch a daemon's confinement off, so upstream appends the
option to the daemon's refuse list unconditionally - after the module's own
`refuse options` rules have been applied, so no `!insecure-links` negation can
re-enable it. A daemon's own opt-out is the `insecure links` module parameter,
never the client flag.

## Why this one was missing

oc already mirrors the two sibling unconditional refusals from that same
`am_daemon` block - `log-file*` and the charset-gated `iconv` - but not this
third one, which 3.5.0 added. A client forwarding `-M--insecure-links` was
therefore not refused.

upstream: `options.c:1077-1088`, with
`parse_one_refuse_match(0, "insecure-links", list_end)` at `:1087`.

## Scope

The refuse evaluator already canonicalises raw client option strings without
consulting a known-option table, so the new clause needs nothing else wired up.
No behaviour changes for any other option.

## Tests

Three, alongside the existing `log-file*` / `iconv` pins:

- `default_refuses_client_insecure_links` - the refusal itself.
- `insecure_links_refusal_survives_negation` - `!insecure-links` in a module's
  `refuse options` cannot re-enable it, matching upstream's ordering.
- `insecure_links_refusal_does_not_catch_unrelated_options` - non-vacuity
  companion: the match is exact, not a prefix sweep, so `--insecure` and
  `--insecure-links-extra` stay allowed. Without it a stray `starts_with` would
  pass the first two while silently refusing a family of options.

RED proven by removing the clause: the two pins fail, the companion stays green
(1 passed, 2 failed).

## Verification

- `cargo fmt --all -- --check` - exit 0
- `cargo clippy --locked --workspace --all-targets --all-features --no-deps -- -D warnings` - exit 0
- `cargo nextest run -p daemon --all-features` - 1800 passed, 6 skipped

Covered end-to-end by upstream's `testsuite/operator-path-insecure-links-refused`.

### DIFF
--- a/crates/daemon/src/daemon/sections/module_parsing/refuse_options.rs
+++ b/crates/daemon/src/daemon/sections/module_parsing/refuse_options.rs
@@ -415,7 +415,14 @@ fn is_option_refused(
     //     server-side logging.
     //   - `iconv` (options.c:1007-1008) is refused only when the module has no
     //     `charset` configured (`!*lp_charset(module_id)`).
+    //   - `insecure-links` (options.c:1087, new in 3.5.0) is a LOCAL-ONLY
+    //     opt-out, so a client must never be able to switch off the daemon's
+    //     symlink confinement by sending it. The daemon's own opt-out is the
+    //     `insecure links` module parameter, never this flag.
     if long_name.starts_with("log-file") {
+        return true;
+    }
+    if long_name == "insecure-links" {
         return true;
     }
     if long_name == "iconv"
--- a/crates/daemon/src/daemon/sections/tests.rs
+++ b/crates/daemon/src/daemon/sections/tests.rs
@@ -401,6 +401,47 @@ fn default_log_file_refusal_survives_negation() {
 }
 
 #[test]
+fn default_refuses_client_insecure_links() {
+    // upstream: options.c:1087 (new in 3.5.0) - a daemon always appends
+    // `insecure-links` to the refuse list. The flag is a LOCAL-ONLY opt-out, so
+    // a client must never be able to switch off the daemon's symlink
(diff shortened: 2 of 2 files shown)


oferchen:
Superseded by #7396, which carries this exact clause and these exact three
tests (adopted verbatim - the exact-match companion covering `--insecure`
and `--insecure-links-extra` is better than what #7396 originally had).

Closing rather than merging because the two changes **cannot be sequenced**.
The refusal here is unreachable on its own: oc's daemon argv builder never
appended `remote_options[]`, so a client's `-M--insecure-links` was silently
discarded before it ever reached a daemon. Measured - with only this clause
applied, `operator-path-insecure-links-refused` still fails, and the daemon
log shows the option never arriving:

    client args: --server --sender -de.LsfxCIvu . mod/

Once the forwarding is fixed the option arrives and is refused, so the test
passes. But the 3.5.0 expect manifests pin that row `fail` on both legs, and
`--expect-result` scores an unexpected pass as a failure. So:

| landed | test outcome | manifest | required context |
|---|---|---|---|
| this PR alone | fail | fail | green |
| forwarding alone | fail | fail | green |
| both | **pass** | fail | **red** |

There is no merge order that avoids the red, because the manifest row has to
flip in the same change that makes the outcome move. #7396 does all three
together: the refusal, the forwarding, and the single manifest flip.

Nothing from this branch is lost.

PR #7402, opened 2026-08-20
fix(cli): honour the `--` end-of-options marker in server-mode argv

oferchen:
## Problem

Every one of the 12 `rrsync-*` tests in the upstream 3.5.0 testsuite failed with the same symptom:

```
rsync: [sender] link_stat "--" failed: No such file or directory (2)
```

Upstream's own `support/rrsync` builds its command as `(RSYNC, *rsync_opts, '--', '.', *rsync_args)` (support/rrsync:655). oc's server-mode argv parser has no concept of the `--` end-of-options marker, so it fell through to the path list and became the source operand.

`upstream: options.c:1491` — the server runs the same popt parser as the client (`poptGetContext(RSYNC_NAME, argc, argv, long_options, 0)`), and popt consumes a bare `--` and hands back everything after it as a plain operand. **oc's client already had this** (`operands_after_end_of_options_are_preserved`); only the server path was missing it.

## The consequential half

The leaked token is the visible symptom. The real defect is that **oc walks the server argv from three places where upstream makes one popt pass**, and none of them stopped at the marker:

| scanner | what a post-`--` path could do |
|---|---|
| `parse_server_flag_string_and_args` | `--` becomes the source path |
| `parse_server_long_flags` | a file named `--trust-sender` / `--write-devices` **sets that option** |
| `detect_secluded_args_flag` | a file named `-s` switches the server onto the stdin arg stream |

That is precisely what rrsync passes `--` to prevent. Its comment at `support/rrsync:558` states the property it depends on: *"an arg that starts with a '-' is safe due to our use of '--' in the cmd tuple."*

## Fix

One `split_at_end_of_options` owns the rule; **each of the three scanners applies it for itself** rather than trusting its caller, so a scanner added later cannot silently reintroduce the hole. Upstream gets this property once, structurally, from a single popt pass — oc cannot, so the invariant is made local to each walker.

The `.` placeholder keeps being consumed on **either** side of the marker: `server_options()` emits it before (clientserver.c:303) while rrsync emits it after, and `main.c` `do_server_sender()`/`do_server_recv()` take `argv[0]` as the directory to change into rather than as a transferred path.

An argv with **no** marker parses exactly as before.

## Verification

- **Testsuite, measured:** rrsync cluster **0/12 → 9/12**.
- **RED proven:** with only the split neutralised, the new tests fail.
- Each assertion is paired with the same token placed *before* the marker as its **non-vacuity companion** — without those, a build that simply discarded every post-marker argument would pass too.
- `cargo fmt --all -- --check` clean; full-workspace `cargo clippy --all-targets --all-features -D warnings` clean.
- `cargo nextest run -p cli -p bin --all-features`: **4299 passed, 0 failed**.

## Manifest

The 9 now-passing rows flip `fail` → `pass` in both manifests (nonroot 56 → 47 failing, root 76 → 67). The change is pure argv parsing, so it is privilege- and platform-independent; both legs are flipped together on that basis.

⚠ The 9 were measured on macOS while the manifests were generated from a real Linux run. If Linux turns out to clear one of the 3 residuals too, that surfaces here as an unexpected pass and is a one-line follow-up — CI is the authoritative Linux oracle for these rows.

## Residuals (separate defects, deliberately left `fail`)

- `rrsync-specials-denied`
- `rrsync-pull-arg-shapes`
- `rrsync-merge-file-confine` — the server reads a merge file outside the restricted dir; belongs to the operator-named-file confinement work, not to argv parsing.


oferchen:
### Note on the 9-row manifest flip: a concrete Linux divergence mechanism

An independent diagnosis of this cluster reproduced the same root cause and the same 9 flipping cells via a different route — a 4-line shim that does nothing but strip a bare `--` from argv before exec'ing an *unmodified* oc binary took the cluster from 0/12 to 9/12. That is a clean corroboration of the mechanism from outside this change.

It also supplies a **specific reason the Linux count may differ**, which upgrades the caveat in the PR description from generic caution:

`support/rrsync` branches on `proc_self_fd_pins()`. On macOS that is false and rrsync hands oc plain pathnames. **On the Linux CI legs it is true**, so rrsync rewrites the arguments to `/proc/self/fd/N` and several of these cells then assert a *different* set of things — e.g. `rrsync-backup-dir-inband-pivot` requires `'/proc/self/fd/' in stderr`, and `rrsync-alt-dest-inband-pivot` requires a first-run empty-basis success.

So the 9 should be read as a **lower bound on the mechanism**, not as a predicted Linux pass count. Causes for the three residuals are PINS-independent by mechanism; the trailing-`/.` one survives pinning too, because rrsync re-appends `/.` to the pinned path (`support/rrsync:935-938`).

The two required `upstream testsuite` contexts on this PR are the authoritative Linux oracle. If any of the 9 does not flip there, the correction is confined to those manifest rows — the code change itself is independent of the branch rrsync takes, since it only concerns how oc's server parses `--`.

### Root causes for the three residuals (for whoever picks them up)

| cell | cause | site |
|---|---|---|
| `rrsync-pull-arg-shapes` | a trailing `/.` source operand is not treated as `DOTDIR_NAME` on the **server sender** (local copy is already correct, so it is flist-only) | `crates/transfer/src/generator/file_list/mod.rs:635-661` `non_relative_walk_base()` tests only `bytes.last() == b'/'`; upstream `flist.c:2603` has the `.`-after-`/` arm |
| `rrsync-specials-denied` | `--drop-D` is parsed into `ServerLongFlags::drop_devices` (`crates/cli/src/frontend/server/flags.rs:507`) and has **zero consumers** — the server half is inert | needs consuming in `crates/cli/src/frontend/server/run.rs` alongside the other long-flag applications |
| `rrsync-merge-file-confine` | `--confine-root` reaches `config.connection.confine_root` but its only consumer is the sender's file-content open (`crates/transfer/src/generator/open_source.rs:72`); merge-file opens never consult it | the operator-named-FILE confinement work |

None of the three requires oc to ship a restricted shell.

oferchen:
### Linux verdict is in — corrected to 7 rows, and my prediction was half wrong

The two required contexts adjudicated the flip. `runtests.py` exits with the mismatch count (`overall result is 2`), and the log names them:

```
rrsync-backup-dir-inband-pivot: expected pass, got fail
rrsync-no-overwrite-logfile:    expected pass, got fail
```

Identical on both legs. Both are now back to `fail` — measured Linux reality, not a waiver; they were `fail` before this PR and still are. **Net: 7 of 12 flip** (nonroot 55 → 48 failing, root 75 → 68).

**`backup-dir-inband-pivot` is the PINS case I flagged in advance.** `support/rrsync` branches on `proc_self_fd_pins()`: false on macOS, true on Linux, where it rewrites arguments to `/proc/self/fd/N` and the cell then requires `'/proc/self/fd/' in stderr` — a different assertion from the one macOS runs.

**`no-overwrite-logfile` is not that.** It has zero references to `proc_self_fd_pins` or `/proc/self/fd`, so it is a second and distinct platform-dependent cause. Meanwhile two cells that *do* branch on PINS — `alt-dest-inband-pivot` and `pull-delivers-content` — passed on Linux. So PINS-referencing turned out to be neither necessary nor sufficient, and with 2 mismatches against 3 PINS candidates the source-level prediction wasn't even unique. I waited for the log rather than guessing; filed separately for its own diagnosis.

The code change is untouched by any of this — it only concerns how oc's server parses `--`, which is independent of the branch rrsync takes.

Rebased onto current master (#7401 landed and touched the same manifest files); the rebase replayed clean and #7401's `ki58-log-format-percent pass` row is intact.

PR #7421, opened 2026-08-21
fix(receiver): stat alt-dest basis via the ownership walk

oferchen:
## What

The receiver resolved a `--compare-dest` / `--copy-dest` / `--link-dest` basis
entry with a plain `fs::symlink_metadata`, which re-walks every component of a
**peer-supplied** entry name at call time. Upstream reaches the same entry
through `owner_walk_parent()` + `link_stat_at()` (`generator.c:979-990`), so a
symlink component owned by neither root nor our euid is refused with `ELOOP`
rather than traversed.

This routes the basis stat onto `fast_io::owner_walk` - the same primitive the
sibling operator-path families already use for `--temp-dir` (#7398) and
`--backup-dir` (#7404).

## Why this shape

Only the **parent** is walked. The leaf is `fstatat(AT_SYMLINK_NOFOLLOW)`'d and
never opened:

- upstream must report a symlink at the basis name **as a symlink** so its
  `!S_ISREG` filter drops it. Opening the leaf resolves it, which measurably
  re-opens the hole #7419 just closed - the existing
  `link_dest_skips_symlink_basis_entry` cell caught exactly that while this was
  being built.
- opening would also demand read permission that the `--link-dest` hard-link
  path does not need.

The `fs::Metadata` handed downstream comes from a second path stat, accepted
only when it names the same `(dev, ino)` the confined stat saw. That keeps
`quick_check_matches` / `metadata_unchanged` on their `std` types while making
a parent flipped between the two calls refusable rather than silently followed.

## Scope, stated plainly

The ownership walk's rule is ownership, not location: uid 0 or our euid is the
operator's own layout (`--link-dest=/backup` where `backup -> /mnt/disk`) and is
followed; any other uid is refused.

**Measured against real rsync 3.5.0**: with `basis/sub -> ../outside` owned by
the invoking user, upstream consumes the basis and does not transfer
`sub/payload.bin`. oc did the same before this change, so the traversing half is
unchanged - and it is now pinned, because breaking it would break the deployments
that rely on it.

The half that changes behaviour needs a **foreign-owned** symlink and therefore
root to construct, so it is not reachable from a non-root `nextest` cell. It
stays pinned at the predicate, by `owner_walk`'s existing
`refuses_an_owner_that_is_neither_root_nor_the_euid`.

## Verification

- `symlink_basis_tests` 6/6, including the three `--copy-dest` / `--link-dest` /
  `--compare-dest` leaf cells from #7419 as the regression guard.
- full `-p transfer --all-features`: 2762 run, 2762 passed, 3 skipped.
- `cargo fmt --all -- --check` clean; `clippy -p transfer -p fast_io
  --all-targets --all-features -D warnings` clean.
- oracle script comparing oc against the built `rsync-3.5.0` binary on the
  symlinked-parent and real-parent cells.

Closes the alt-dest half of the operator-path confinement family.

### DIFF
--- a/crates/fast_io/src/dir_sandbox/at_syscalls/metadata.rs
+++ b/crates/fast_io/src/dir_sandbox/at_syscalls/metadata.rs
@@ -192,3 +192,41 @@ pub fn fstatat_nofollow(dirfd: BorrowedFd<'_>, name: &OsStr) -> io::Result<AtMet
         Err(io::Error::last_os_error())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::fstatat_nofollow;
+    use std::os::fd::AsFd;
+    use std::os::unix::fs::MetadataExt;
+
+    /// [`AtMetadata`](super::AtMetadata) and [`std::fs::Metadata`] must agree on
+    /// the identity and comparison fields for the same file.
+    ///
+    /// This is not a tautology across platforms. `fstatat_nofollow` calls
+    /// `libc::fstatat`, while `std::fs::symlink_metadata` on Linux goes through
+    /// `statx` and reconstructs `st_dev` from the returned major/minor pair. Any
+    /// caller that stats a file both ways and then compares the two results -
+    /// which is how the receiver's alt-dest basis lookup verifies that a
+    /// path-based stat landed on the inode the confined stat saw - is silently
+    /// broken if those reconstructions differ on any target.
+    ///
+    /// Kept as a standing invariant rather than a one-off probe: it costs two
+    /// syscalls and it fails loudly on the platform where it stops holding.
+    #[test]
+    fn at_metadata_agrees_with_std_metadata_on_the_same_file() {
+        let temp = tempfile::tempdir().expect("tempdir");
+        let path = temp.path().join("probe.bin");
+        std::fs::write(&path, b"probe payload").expect("write probe file");
+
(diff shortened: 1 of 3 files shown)


oferchen:
Marking draft: the required `interop` check is red, and it is this change.

```
--link-dest: hello.txt not hardlinked (count=1)                 UNEXPECTED FAIL: up:link-dest
--compare-dest: hello.txt was transferred despite matching ref  UNEXPECTED FAIL: up:compare-dest
```

Both fail at every protocol (unforced and fp=28..32), in the `up:` cells only — upstream as client with oc as the server-side receiver via `--rsync-path`. Master is green at this branch's base (80d5461ab), so this is a regression, not an inherited red. Failure mode is "basis looks absent", which is exactly what `basis_stat` returning `None` produces.

Ruled out so far, by measurement on macOS against the same shapes:
- relative vs absolute alt-dest path — both skip/hardlink correctly
- `--compare-dest` / `--link-dest` / `--copy-dest` on the local-copy path — all correct
- upstream-client push to an oc server receiver with the reference dir inside the destination (the harness's exact layout) — correct

So the trigger is Linux-specific and not reproducible here. Not merging until it is root-caused; the earlier `operator_open_read` misstep in this same PR is a reminder that the peer-facing receiver is a different path from the one the unit tests exercise.

PR #7535, opened 2026-08-26
fix(backup): confine the in-place backup copy to the module root

oferchen:
The `--inplace --backup` pre-image copy opened the operator-named backup path
through the ownership walk alone. The walk FOLLOWS a symlink owned by uid 0 or
our euid by design - that is the operator's own layout - and a non-chrooted
daemon writes its `--backup-dir` entries as its own uid, so a backup entry is
TRUSTED-owned by construction. Ownership therefore cannot be the whole defence
here, exactly as it was not for the merge-file open in #7512.

## The escape

Measured end-to-end against a real daemon (module root `.../module`, `path =`
that directory, `read only = false`, no chroot):

```
module/payload            "PRE-IMAGE-IN-MODULE"      the destination
module/bak/payload   ->   ../../outside/secret       trusted-owned plant
outside/secret            "OUTSIDE-THE-MODULE-UNTOUCHED"
```

`oc-rsync --inplace --backup --backup-dir=bak --ignore-times src/payload
rsync://127.0.0.1:PORT/data/` leaves `outside/secret` holding
`PRE-IMAGE-IN-MODULE` and the client exits **0**. An in-module file's contents
are written to a path outside the module root, silently, with no diagnostic.

The plant is deliberate and pre-transfer, not a race: the escape needs only a
trusted-owned symlink to EXIST at the backup leaf.

## Why upstream does not have it

Upstream raises `operator_path_resolve` around the whole backup
(`backup.c:443-449` `make_backup()`) and again around the in-place backup that
bypasses `make_backup()` (`generator.c:2281-2301` and `:2327-2349`). That flag
is precisely what arms `abspath_outside_confinement()`
(`syscall.c:186-240`), which refuses a resolved path that has left
`confinement_root()` - `module_dir` for a daemon. oc carried the ownership half
of that rule and not the root half.

## Change

`fast_io` grows the `PathKind::Confined` twin of the create-and-truncate
operator open, and `copy_pre_image_to_backup` takes it. The walk still follows
a trusted symlink; the module root then judges where the walk actually landed.
Refusal is `ELOOP`, matching `syscall.c:464-466` - `EXDEV` would be laundered by
callers that treat it as cross-device.

Scope is the one measured sink. The surrounding backup tiers already anchor
through the receiver's destination sandbox (rename, hard link, `mkdir`); the
in-place copy was the unanchored one, which is the same asymmetry upstream calls
out in its own comment. The `Ancillary` twin stays with `--write-batch`, which
may legitimately live outside the tree.

A session with no confinement root - every plain local client, which installs a
local session with `root: None` - is unaffected.

## Verification

`crates/transfer/tests/daemon_inplace_backup_dir_confinement.rs`, three cells
through a real daemon:

| cell | asserts |
|---|---|
| `a_backup_symlink_leaving_the_module_must_not_be_written_through` | the out-of-module file is untouched and the plant survives |
| `a_backup_symlink_staying_inside_the_module_is_still_followed` | POSITIVE CONTROL - an in-module symlink target still receives the pre-transfer bytes, exit 0 |
| `an_unplanted_backup_leaf_gets_the_pre_transfer_bytes` | NON-VACUITY - the same push with nothing planted still makes a backup and rewrites the destination |

Plus two `fast_io` cells for the new entry point, in the file #7512 established.
They run unprivileged: upstream refuses only `st_uid != 0 && st_uid !=
trusted_uid` (`syscall.c:406`), so uid 0 and the euid take one identical follow
path into one identical confinement check, and the daemon here runs as the
test's own uid.

### Mutation-proven three ways

1. `refuse_if_outside` neutered to `Ok(())` - the confinement half - fails
   `a_backup_symlink_leaving_the_module_must_not_be_written_through` (`left:
   "PRE-IMAGE-IN-MODULE"`) and `a_confined_create_leaving_the_module_is_refused`,
   with both positive controls still green.
2. The call site reverted to the unconfined open, everything else intact: 3
   tests run, 2 passed, 1 failed - exactly the escape cell. So the change is
   load-bearing and not over-refusing.
3. The walk made to refuse EVERY symlink: both in-module follow cells and the
   unplanted cell go red, so the positive controls really do bite.

`fast_io` 784/784, `engine -E 'test(backup)'` 133/133, `transfer -E 'test(backup)
+ binary(daemon_inplace_backup_dir_confinement)'` 51/51, and the root
`integration_delete_backup` / `server_backup_temp_dir` / `append_implies_inplace`
targets 29/29. fmt and clippy clean.

## Noted, not fixed here

A daemon push whose destination operand names a FILE rather than a directory
fails with `Not a directory (os error 20)` as soon as `--backup-dir` is present,
independently of this change. It reproduces on master and is a separate defect.

### DIFF
--- a/crates/engine/src/local_copy/executor/file/backup.rs
+++ b/crates/engine/src/local_copy/executor/file/backup.rs
@@ -17,7 +17,8 @@ use crate::local_copy::create_symlink;
 use crate::local_copy::map_metadata_error;
 
 /// Duplicates a destination's pre-transfer bytes into the operator-named
-/// backup path, resolving that path with the ownership walk.
+/// backup path, resolving that path with the ownership walk and confining the
+/// result to the session's confinement root.
 ///
 /// Used by the `--inplace --backup` paths, where the destination inode is
 /// rewritten in place rather than replaced, so the pre-image must be COPIED
@@ -31,6 +32,15 @@ use crate::local_copy::map_metadata_error;
 /// at the `--backup-dir` leaf. `std::fs::copy` has exactly that behaviour,
 /// which is what upstream's `operator-path-inplace-backup-dir` cell observes as
 /// an escape out of the transfer tree.
+///
+/// `operator_path_resolve` is BOTH halves of upstream's rule, so the open is
+/// the confined one: the ownership walk follows a trusted-owned symlink by
+/// design, and a non-chrooted daemon's own `--backup-dir` entries are
+/// trusted-owned by construction, so ownership alone would let a backup entry
+/// pointing outside the module carry an in-module file's contents out of it
+/// (syscall.c:186-240 `abspath_outside_confinement()`). A session with no
+/// confinement root - every plain local client - is unaffected: there is
+/// nothing to be outside of.
 ///
 /// Permissions are taken from the source and applied through the open
 /// descriptor, matching `fs::copy` (and upstream's
@@ -49,7 +59,7 @@ pub fn copy_pre_image_to_backup(source: &Path, backup_path: &Path) -> io::Result
 
         // Mask to the permission bits: `mode()` carries the file-type bits too,
         // and only the low 12 are meaningful as an `O_CREAT` mode.
-        fast_io::operator_open_write_create(backup_path, permissions.mode() & 0o7777)?
+        fast_io::operator_open_write_create_confined(backup_path, permissions.mode() & 0o7777)?
     };
     // Non-Unix has no ownership walk to run; degrade to a plain create, as the
     // other operator-path helpers do.
--- a/crates/fast_io/src/lib.rs
+++ b/crates/fast_io/src/lib.rs
@@ -409,8 +409,8 @@ pub use nofollow_open::open_basis_nofollow;
 pub use owner_walk::{
     operator_link, operator_mkdir, operator_open_append, operator_open_create_new,
     operator_open_read, operator_open_read_confined, operator_open_recv, operator_open_rw_create,
-    operator_open_write_create, operator_read_to_string, operator_rename,
-    operator_symlink_metadata, owner_trusted_parent, symlink_owner_is_trusted,
+    operator_open_write_create, operator_open_write_create_confined, operator_read_to_string,
+    operator_rename, operator_symlink_metadata, owner_trusted_parent, symlink_owner_is_trusted,
 };
 pub use refs_detect::{clear_refs_cache, is_refs_filesystem};
 #[cfg(unix)]
--- a/crates/fast_io/src/owner_walk.rs
+++ b/crates/fast_io/src/owner_walk.rs
@@ -747,6 +747,44 @@ pub fn operator_open_write_create(path: &Path, mode: u32) -> io::Result<std::fs:
     )
 }
 
+/// Open an operator-supplied path for writing, creating or truncating it,
+/// additionally bound to the session's confinement root.
+///
+/// The `--backup-dir` shape under `--inplace`. The in-place backup has to
+/// duplicate the destination's pre-transfer bytes into the operator-named
+/// backup path, and the ownership walk follows a trusted-owned symlink at that
+/// path by design. A non-chrooted daemon writes its `--backup-dir` entries as
+/// its own uid, so a backup entry is TRUSTED-owned by construction: point one
+/// outside the module and the follow carries an in-module file's contents out
+/// of the tree. Ownership decides whether to follow; the root decides whether
+/// the landing site is acceptable.
+///
+/// The `Ancillary` twin [`operator_open_write_create`] stays with the opens
+/// that may legitimately live outside the tree - the `--write-batch` files.
+///
+/// # Upstream Reference
+///
+/// - `rsync-3.5.0/backup.c:443-449` `make_backup()` - `operator_path_resolve =
+///   1` around the whole backup, naming `--backup-dir` as an operator path.
+/// - `rsync-3.5.0/generator.c:2281-2301` and `:2327-2349` - the in-place backup
+///   bypasses `make_backup()`, so the generator raises the same flag around its
+///   own `copy_file()` / `do_open_at()`.
+/// - `rsync-3.5.0/syscall.c:186-240` `abspath_outside_confinement()`.
+///
+/// # Errors
+///
+/// - `ELOOP` when a component is an untrusted-owner symlink, or when the
+///   resolved path lands outside the confinement root.
+/// - Otherwise see `operator_open_with`.
+pub fn operator_open_write_create_confined(path: &Path, mode: u32) -> io::Result<std::fs::File> {
+    operator_open_kind(
+        path,
+        OFlags::WRONLY | OFlags::CREATE | OFlags::TRUNC,
+        Mode::from_bits_truncate(mode as rustix::fs::RawMode),
+        crate::confinement::PathKind::Confined,
+    )
+}
+
 /// Exclusively create an operator-supplied path through the ownership walk.
 ///
 /// The `O_EXCL` twin of [`operator_open_write_create`], for the receiver's
--- a/crates/fast_io/tests/operator_open_module_confinement.rs
+++ b/crates/fast_io/tests/operator_open_module_confinement.rs
@@ -181,6 +181,62 @@ fn a_plain_path_inside_the_module_opens() {
     assert_eq!(read_all(file), "PLAIN");
 }
 
+/// THE WRITE-SIDE PIN. The same rule on the create-and-truncate entry point
+/// the `--inplace --backup-dir` pre-image copy uses. A read-side leak returns
+/// out-of-module TEXT to the peer; this one WRITES an in-module file's contents
+/// to a path outside the module, so both entry points have to carry the check.
+///
+/// upstream: `rsync-3.5.0/backup.c:443-449` `make_backup()` and
+/// `generator.c:2281-2301` - `operator_path_resolve = 1` around the backup and
+/// around the in-place copy that bypasses it.
+#[test]
+fn a_confined_create_leaving_the_module_is_refused() {
+    let fixture = fixture();
+    let planted = fixture.module.join("backup/sub/evil");
+    symlink(&fixture.secret, &planted).expect("plant the symlink");
+
+    let _session = serve_module(&fixture.module);
+    let error = fast_io::operator_open_write_create_confined(&planted, 0o600)
+        .expect_err("a confined create must not resolve out of the module");
+
+    assert_eq!(
+        error.raw_os_error(),
+        Some(libc::ELOOP),
+        "the refusal must be ELOOP, as on the read side"
+    );
+    assert_eq!(
+        fs::read_to_string(&fixture.secret).expect("the outside file survives"),
+        "SECRET-MARKER",
+        "a refused create must not have truncated the outside file: O_TRUNC \
+         does its damage at open time, so a refusal that happened after the \
+         open would leave nothing to read back"
+    );
+}
+
+/// NEGATIVE CONTROL for the write side. A trusted symlink whose target stays
+/// inside the module is still followed and written through - the ordinary
+/// operator layout the walk exists to keep working. Without this, "refuse every
+/// symlink at the backup leaf" would satisfy the pin above.
+#[test]
+fn a_confined_create_staying_inside_the_module_is_followed() {
+    let fixture = fixture();
+    let inside = fixture.module.join("data");
+    fs::write(&inside, "IN-MODULE").expect("write in-module target");
+    let link = fixture.module.join("backup/sub/benign");
+    symlink(&inside, &link).expect("plant the in-module symlink");
+
+    let _session = serve_module(&fixture.module);
+    let mut file = fast_io::operator_open_write_create_confined(&link, 0o600)
+        .expect("an in-module target must still be written through");
+    std::io::Write::write_all(&mut file, b"REWRITTEN").expect("write");
+    drop(file);
+
+    assert_eq!(
+        fs::read_to_string(&inside).expect("read the in-module target"),
+        "REWRITTEN"
+    );
+}
+
 /// The other half of upstream's rule: the confinement applies to CONFINED opens
 /// only. `--log-file`, the `--*-from` family and the daemon's lock and motd
 /// files may legitimately live outside the tree, so the ANCILLARY entry point
--- a/crates/transfer/tests/daemon_inplace_backup_dir_confinement.rs
+++ b/crates/transfer/tests/daemon_inplace_backup_dir_confinement.rs
@@ -1,0 +1,328 @@
+//! A daemon's `--inplace --backup-dir` must not write outside the module root
+//! through a *trusted-owned* symlink planted at the backup leaf.
+//!
+//! # Why the ownership walk is not enough here
+//!
+//! `--inplace --backup` bypasses the rename-to-backup mechanism: the
+//! destination inode is rewritten in place, so its pre-transfer bytes must be
+//! COPIED aside first. That copy opens the operator-named backup path with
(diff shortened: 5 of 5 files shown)


oferchen:
### CI reds root-caused: blocked on #7541, no code change needed here

`nextest (beta)`, `nextest (nightly)` and `Linux musl (nightly)` are red on the
**same two tests**, with the **same errno**, in this PR's own new test file:

```
a_backup_symlink_staying_inside_the_module_is_still_followed      (:275)
an_unplanted_backup_leaf_gets_the_pre_transfer_bytes             (:304)

```

Three red cells, one cause — not three.

`Permission denied (13)` is **EACCES**, which is the signature of the ownership
walk being unable to *traverse* under a Landlock sandbox. That is the defect
#7541 fixes (its A/B against this branch's head moves the file from
1-passed-2-failed to 3-passed).

Both failing tests are this PR's **positive controls** — the in-module cases that
must *succeed*. The escape pin passes either way, so the discriminator is the
cause (EACCES vs ELOOP), not the outcome. Cross-PR control: #7542 has
`nextest (beta)` and `Linux musl (nightly)` both SUCCESS, which rules out
toolchain, runner and master-wide causes.

**Ordering:** land #7541, then rebase this branch. `git merge-tree` reports the
two merge clean against each other, so the rebase is mechanical. #7538 touches
`fast_io/src/lib.rs` and `owner_walk.rs` too and should land after both.

oferchen:
**This failure is shared with the sibling PR — it is one defect, not two.**

Both open confinement PRs are red in `nextest (stable)` with the same error, in their own separately-written test files:

| PR | failing tests | error |
|---|---|---|
| #7538 | `daemon_backup_dir_symlink_confinement`: `an_ordinary_backup_dir_gets_the_pre_transfer_bytes`, `an_ordinary_backup_dir_works_under_delay_updates`, `a_backup_dir_symlink_staying_inside_the_module_is_still_followed` | `mkstemp "payload" (in data) failed: Permission denied (13)` |
| #7535 | `daemon_inplace_backup_dir_confinement`: `an_unplanted_backup_leaf_gets_the_pre_transfer_bytes`, `a_backup_symlink_staying_inside_the_module_is_still_followed` | same |

All five are **positive controls** — in-module operations the confinement is documented to allow — and all exit 23 where 0 is expected. Two authors, two fixtures, one errno, one option combination. That cross-PR agreement is the evidence that the cause is shared rather than per-PR.

**What is established:**

- Reproduces **only** with the `landlock` feature. `Feature flag combinations / compression (linux)` runs these same test files under `--features zstd,lz4` — no landlock — and is green. `nextest (stable)` runs `--all-features` and is red. The green cell cannot exercise the sandbox, so it is not coverage.
- **Not** a blanket sandbox denial: ~32,469 tests pass alongside in the same run.
- macOS is fully green (7809/7809 for the affected crates). Landlock is Linux-only, so that is expected rather than reassuring.
- Both test files are **absent on master**, so neither set has ever been green. These are the first tests of *daemon push + `--backup-dir` under Landlock*.

⚠ Both runs were fail-fast truncated (817 and 815 tests never ran), so five is a lower bound.

**Open, and deliberately not guessed at:** whether the behaviour is already broken on master — these tests being correct and exposing a pre-existing ruleset gap — or whether the new confined operations cause it. Prior art points at the former: the daemon walk starting at `/` and being EACCESed (task 608), Landlock setup failing under `use chroot` with chroot silently carrying the module (task 1050), and module re-resolution EACCESing after the privilege drop (task 640).

Settling it needs a Linux run with `--features landlock`; I have not made that measurement. Tracked as task 1056. **Neither PR should be patched in isolation until it is answered** — a per-PR fix would mask a shared defect.

oferchen:
## Correction: the `landlock` attribution in my earlier comment is wrong

I previously wrote that this failure "reproduces only with the `landlock` feature". **That is refuted.** Retracting it here so the wrong attribution does not get acted on.

**Why it is wrong.** `landlock` is not a root-package feature. `crates/cli/Cargo.toml:103` and `crates/daemon/Cargo.toml:114` both pin it directly:

```toml
[target.'cfg(target_os = "linux")'.dependencies]
fast_io = { path = "../fast_io", default-features = false, features = ["landlock"] }
```

That is unconditional on Linux — not behind any cargo feature. So Landlock is compiled and engaged in **every** Linux workspace build, including the ones that pass.

**The actual measured discriminator**, from this PR's own check rollup:

| Linux cell | command | these tests |
|---|---|---|
| `Feature flag combinations / default-features (linux)` | `cargo nextest run --locked --profile ci --workspace` | **PASS** |
| `nextest (stable)` | `cargo nextest run --locked --profile ci --workspace --all-features` | **FAIL** |

Both are Linux, both build the `oc-rsync` binary, both run the full workspace with no test filter (`_test-features.yml:269` is a bare `cargo nextest run --locked --profile ci ${{ matrix.args }}`), and both have Landlock live. The only difference is the non-default features.

So the cause is somewhere in what `--all-features` adds. Root package: `zlib-ng`, `iouring-send-zc`, `mmap-free-basis`, `openssl`, `openssl-vendored`, `quic`, `sd-notify`. Plus every workspace member's non-default features, which under one `--workspace --all-features` build unify into the binary — notably **`daemon/daemon-seccomp`**, which is off in the passing cell.

**Two candidates checked and excluded so far:**

- **seccomp directly** — its default action is `Errno(EPERM)` (errno 1); the observed failure is `Permission denied (13)` = EACCES. `seccomp.rs:4` states the distinction explicitly: "Landlock denies path-based syscalls with EACCES; seccomp … EPERM".
- **seccomp → umask → mode 000** (the known indirect route to EACCES: `umask(2)` is absent from the allowlist, a lazy read returns -1, `dflt_perms` collapses to 0). Already closed on master by `sanitize_umask` + eager `init_orig_umask` (`crates/metadata/src/apply/permissions.rs:60-135`), which rejects any value outside `0o777` and falls back to `0o022`.

**What settles it** is a two-run discriminator on a Linux host, running just these test files:

1. `--workspace --all-features` → expect FAIL (reproduces)
2. same, minus `daemon/daemon-seccomp` → PASS or FAIL splits the remaining space in half

I do not have a usable Linux host at the moment (the shared one is at load 11.5 with 13 GB free and another testsuite run in flight), so this is measured-so-far, not concluded.

**This PR should still not be patched in isolation** — the same failure hits the sibling PR identically in its own separately-written test file, so a per-PR fix would mask a shared defect.

oferchen:
## Phase 2 (pattern analysis): the fixture shape is excluded, and two further findings

Following the correction above, I compared the failing fixture against the working examples instead of continuing to reason about feature flags.

### 1. The config shape is NOT the differentiator

All 13 master daemon integration tests (`daemon_dest_mode_parity`, `daemon_push_refused_file_reports_error`, `daemon_total_transferred_size_stats`, `delay_updates_daemon_pull`, `msg_no_send_unsendable_file`, `verify_redo_recovers_network_paths`, …) use the **same** config shape as this fixture:

```
pid file = <tmpdir>/rsyncd.pid      # outside the module root
log file = <tmpdir>/rsyncd.log      # outside the module root
use chroot = false
[data]
path = <tmpdir>/module
```

Same tempdir module, same `use chroot = false`, same pid/log files outside every Landlock-granted root. Those tests pass under `--all-features`. So "log/pid outside the module root" and "unprivileged tempdir daemon" are both excluded.

### 2. The one difference is `--backup-dir`, and it has ZERO integration coverage on master

**No master integration test passes `--backup-dir` (or `--temp-dir` / `--partial-dir` / `--compare-dest` / `--copy-dest` / `--link-dest`) over a real `rsync://` daemon connection.** The only in-tree daemon test naming one is `crates/daemon/src/tests/chunks/daemon_compare_dest_push.rs`, which is an in-crate test and does not go through the real connection path.

That matters structurally. `engage_landlock_sandbox` has exactly **one** production call site:

```
crates/daemon/src/daemon/sections/module_access/transfer/orchestration.rs:590
    if !engage_landlock_sandbox(ctx, module, &extra_allowed)? {
```

and `extra_allowed` is populated only from those staging / alt-dest options (`sandbox.rs:242`). So `roots.len() > 1` — the multi-root Landlock ruleset on a live daemon — **is reached for the first time by these two PRs**. Every other daemon test engages Landlock with exactly one root.

So the honest characterisation is a **conjunction**: `--backup-dir` over a daemon (previously untested) *and* some non-default feature. `--backup-dir` alone is not sufficient — the `default-features (linux)` cell runs these same tests and passes.

### 3. ⚠ Two of the five tests pass VACUOUSLY

Of the five failing-file tests, the three that FAIL are the positive controls. The two that PASS —
`a_backup_dir_symlink_leaving_the_module_must_not_receive_the_pre_image` and
`a_delayed_update_backup_must_not_leave_the_module` — assert a **refusal**, and a transfer that aborts at EACCES also fails to escape. They cannot distinguish "the confinement refused the escape" from "the transfer died before reaching the backup".

This is the same vacuity already recorded for #7535's escape pin. **Even once the EACCES is resolved, the escape assertions need to be re-confirmed non-vacuous** — otherwise the security fix these PRs exist for is unverified on Linux.

### Status

The two-run discriminator (`--workspace` control vs `--workspace --all-features`, same host, same two test binaries) is building now on an idle aarch64 Linux box. aarch64 is a caveat: a reproduction there is decisive, a non-reproduction is not, since some candidates (e.g. the `*at`-only seccomp allowlist) are x86_64-specific by construction.

oferchen:
## Phase 3: measured on Linux. The CI configuration does NOT reproduce on aarch64 — and that asymmetry is the lead.

I built the `oc-rsync` binary in three feature configurations on an idle aarch64 Linux box and drove the *ordinary* case directly (`--backup --backup-dir=bak --ignore-times`, real `bak` directory, `use chroot = false`, tempdir module) — the same shape as `an_ordinary_backup_dir_gets_the_pre_transfer_bytes`.

Crucially I read the **daemon's own log**, which reports its sandbox state. No test does this today, and it is what made the result interpretable.

| cell | landlock | seccomp | client exit | backup |
|---|---|---|---|---|
| `--workspace` (default features) | `fully enforced over 1 root(s)` | `unavailable in this build` | **0** | correct |
| `-p bin --all-features` | `fully enforced over 1 root(s)` | `unavailable in this build` | 0 | correct |
| `--workspace --all-features` | `fully enforced over 1 root(s)` | **`BPF filter engaged (EPERM on unlisted syscalls)`** | **0** | correct |

Three findings.

**1. ⚠ My middle cell was INERT, and the log caught it.** `cargo build -p bin --all-features` enables all features of the **`bin` package only** — not every member's. `cargo nextest run --workspace --all-features` (what CI runs) enables `daemon/daemon-seccomp`; `-p bin --all-features` does not. Both binaries logged `seccomp unavailable`, which is how I noticed. Worth stating plainly because the two spellings look interchangeable and are not.

**2. Seccomp being engaged is NOT sufficient.** The third cell has the exact CI feature configuration, seccomp genuinely installed, and it still exits 0 with a correct backup. So `daemon-seccomp` alone does not explain the failure.

**3. `--backup-dir` does NOT widen the Landlock ruleset.** All three cells log **`over 1 root(s)`**. My earlier "these PRs are the first to reach a multi-root ruleset" claim is wrong — `extra_allowed_paths` stays empty here, and `bak` is covered because it already sits inside the module root.

### The lead: this is arch-asymmetric, and the seccomp allowlist is arch-asymmetric in exactly the same way

`crates/daemon/src/daemon/sections/seccomp.rs` allowlists **only the `*at` syscall forms** — `openat`, `openat2`, `newfstatat`, `statx`, `renameat`, `renameat2`, `unlinkat`, `mkdirat`, `linkat`, `symlinkat`, `readlinkat`, `fchmodat`, `fchownat`. The legacy `open` / `stat` / `rename` / `unlink` / `mkdir` / `link` are absent, and the only `#[cfg(target_arch = "x86_64")]` additions are `arch_prctl` plus portability shims for `rseq` / `sendfile` / `fadvise64` — **no legacy syscalls are added back on x86_64**.

On **aarch64 those legacy syscalls do not exist**, so a `*at`-only allowlist is complete by construction. On **x86_64 they do exist**, and anything issuing one gets `EPERM`. That predicts exactly what is observed: green on aarch64 with seccomp engaged, red on x86_64 CI.

**This is a hypothesis, not a conclusion.** The unresolved half is the errno: seccomp's default action is `EPERM` (1) and the failure reports `EACCES` (13), so if this is the cause there is an intermediate step — a fallback path, or a later Landlock-checked operation reached only after an earlier legacy syscall was refused. I am not asserting the mechanism.

**What settles it:** the same manual repro on an **x86_64** host with a `--workspace --all-features` binary, reading the daemon log and `strace`. Our x86_64 box is at 98 % disk with another testsuite run in flight, so I have not run it — I filled a different host's disk earlier today with a `--workspace` nextest build and will not repeat that on a box with 13 G free.

### Two things worth fixing regardless of the root cause

- **The fixture discards the daemon's own diagnosis.** `rsyncd.log` carries the landlock root count, the enforcement tier and the seccomp state; the tempdir is dropped and no test reads it. Surfacing it on failure would have made this a one-run diagnosis instead of a day's work.
- **Two of the five tests still pass vacuously** (the escape assertions — an EACCES abort also fails to escape). That must be re-confirmed once the transfer succeeds.

oferchen:
## This is the same defect as #7538, and it is not about symlinks

#7535 and #7538 fail the same three `nextest` cells with a byte-identical error,
from **two independently written test files**:

| PR | file | failing shape |
|---|---|---|
| #7538 | `daemon_backup_dir_symlink_confinement.rs` | 3 tests, all positive controls |
| #7535 | `daemon_inplace_backup_dir_confinement.rs` | incl. `an_unplanted_backup_leaf_gets_the_pre_transfer_bytes` |

```
```

Two separately authored fixtures producing the same errno at the same step is
what identifies this as a **shared cause upstream of both PRs**, not a test bug
in either. Neither should be patched in isolation — a per-PR workaround would
mask it.

**The `unplanted` cell is the sharpest statement of it.** That test plants no
symlink and asserts only *"the plain push must succeed"*. So the minimal failing
case is:

> a plain in-module daemon push with `--backup-dir` fails with EACCES when the
> receiver creates its destination temp file.

No symlink, no escape, no confinement decision — the transfer dies before any
backup logic runs. Every confinement-flavoured reading of this failure,
including several of mine, was aiming at the wrong layer.

**Scope, measured on three Linux full-workspace cells at the same commit:**
`default-features (--workspace)` PASS, `Linux musl (--no-default-features + 6)`
PASS, `nextest (--all-features)` FAIL. Ruled out so far: seccomp (its default
action is `EPERM`, not `EACCES`), the `umask` route (closed at
`sanitize_umask` → `FALLBACK_UMASK`), and `acl` (it is *in* the default set, so
the passing cell already has it).

⚠ The `openssl` / `zlib-ng` / `landlock` feature-matrix rows do **not** exonerate
those features — each is `-p`-scoped (`-p checksums`, `-p compress -p protocol`,
`-p fast_io`) and never builds the `transfer` integration tests.

#7538 now carries the daemon's `rsyncd.log` in its assertion output, which will
say directly whether landlock is enforced and over how many roots. That evidence
lands on the next run and applies to both PRs.

Related: task 1032 records a pre-existing daemon-push-with-`--backup-dir`
failure in this same fixture family.

oferchen:
Root-caused, and it is **not this PR**. The shared failure both PRs hit is
fixed on **#7545**.

`crates/fast_io/src/owner_walk.rs` opened the walk's anchor (`/`) via
`rustix::fs::open`, which on x86_64 lowers to the raw legacy `SYS_open` — a
syscall the daemon's seccomp worker allowlist deliberately omits in favour of
the `*at` variants. strace, in the daemon process:

```
```

so the anchor open is refused and every confined resolution beneath it fails —
surfacing as `mkstemp "payload" (in data) failed: Permission denied (13)`.

Proven by A/B on one binary: `--all-features` fails, `+OC_RSYNC_NO_SECCOMP=1`
passes; and with #7545 applied the same cell passes with seccomp still on.

`git log -S` puts the raw `open` in exactly one commit — #7541, merged earlier
today — so this is a same-day regression on master, not a defect in either PR.
Rebase once #7545 lands.

PR #7538, opened 2026-08-26
fix(transfer): confine the --backup-dir link and rename to the module root

oferchen:
## The escape

A daemon push with `--backup --backup-dir=bak`, where a **trusted-owned directory symlink** stands at `bak`, moves the destination's pre-transfer bytes **outside the module root**. The client exits **0** with no diagnostic. No `--inplace` is involved, and the plant is persistent - nothing here is a race.

Measured on `2f4b383d8`, unfixed:

```
=== plant=dirlink flags= exit=0 ===
outside/payload      : PRE-IMAGE-IN-MODULE      <- the in-module file's contents, outside the module
module/payload       : NEW-CONTENT
bak is symlink       : yes
client stderr        :
```

The unplanted control on the same fixture proves it reaches the sink:

```
=== plant=none flags= exit=0 ===
outside/payload      : OUTSIDE-UNTOUCHED
module/bak/payload   : PRE-IMAGE-IN-MODULE
```

`--delay-updates` escapes through a different sink and `--fuzzy` through the same one; all three are covered.

## Why the ownership walk cannot be the defence

It **follows** a symlink owned by uid 0 or our own euid, by design - that is the operator's own layout and refusing it would break the ordinary case. A non-chrooted daemon writes everything it creates as its own uid, so a directory symlink a writable module already contains is trusted-owned *by construction*.

Upstream closes this with the second half of the rule: `make_backup()` raises `operator_path_resolve` around the **whole** backup (`backup.c:443-449`), which is what arms `abspath_outside_confinement()` inside `owner_walk_parent()` (`syscall.c:576-590`). Ownership decides whether to *follow*; the module root decides whether the *landing site* is acceptable.

## The structural cause

oc already had the root check. It was never armed on this path, and could not be: the shared parent walk `owner_trusted_parent()` hard-coded `PathKind::Ancillary`, so **every** `*at` operator helper - `operator_link`, `operator_rename`, `operator_mkdir`, `operator_symlink_metadata` - was permanently unconfined no matter what the call site wanted. Two of those call sites' own doc comments already cited the upstream region that arms it.

Upstream arms confinement by **region**; oc passes the kind per call. That is a deliberate and better design - a missing answer is a compile error rather than a control-flow accident - but it only works if the parameter actually reaches the walk.

## What this changes

`fast_io`:
- `owner_walk_open` can now report where it landed, and a kind-taking `owner_trusted_parent_kind` judges `<resolved parent>/<leaf>` against the root, exactly as upstream does. Judging the parent alone is a *different* rule: an absolute walk may pass through the root's own ancestors on the way down, so a backup directory resolving to the module root's PARENT reads as "still descending" while the file it deposits plainly is not inside the module.
- New `operator_link_confined` / `operator_rename_confined`, the `Confined` twins of the existing helpers.

Call sites routed through them - all five `make_backup()` sinks:
- the disk-commit **hard-link tier**, which runs FIRST and is the tier that escapes wherever `link(2)` succeeds (macOS, and Linux without `openat2`);
- the disk-commit **rename tier**, the tier that escapes wherever the nested link is refused by `RESOLVE_BENEATH`;
- the **`--delay-updates`** sweep, a separate `make_backup()` site with its own rename;
- the **synchronous receiver's** backup rename;
- the **local-copy executor's** backup link and rename.

The SEC-1.j sandbox dirfd route is kept for the single-component-under-`dest_dir` shape it was designed for, which is the shape `--backup` without `--backup-dir` produces. Only the `--backup-dir` shape - which that anchor was never the right one for - moves to the ownership walk.

Plus one upstream-fidelity fix that the refusal needs to mean anything: a failed backup now **skips** the delayed update instead of overwriting the destination anyway (`receiver.c:694-695`). The pre-image is exactly what the backup could not save, so committing over it destroys what the refusal was protecting.

## Not over-confinement

An in-module trusted symlink at the backup directory is **still followed and written through**, pinned as a positive control at both levels. A session with no confinement root - every plain local or remote-shell client - is unaffected: there is nothing to be outside of. Both were measured:

```
=== CONTROL: daemon, --backup-dir is an IN-MODULE symlink (exit=0) ===
module/realbak/payload : PRE-IMAGE-IN-MODULE
bak still a symlink    : yes

=== CONTROL: plain local --backup-dir (exit=0) ===
dst/bak/payload : PRE-IMAGE

=== CONTROL: local --backup-dir OUTSIDE the dest tree (exit=0) ===
elsewhere/payload : PRE-IMAGE
```

## Tests, and that they are not vacuous

`crates/transfer/tests/daemon_backup_dir_symlink_confinement.rs` - 5 end-to-end daemon cells (two pins, one over-refusal control, two non-vacuity controls). `crates/fast_io/tests/operator_open_module_confinement.rs` - 6 new helper-level cells.

Four mutations, each killing a different cell:

| mutation | cells reddened |
|---|---|
| revert the call sites (i.e. the pre-fix state) | both daemon pins - `left: "PRE-IMAGE-IN-MODULE"`, client exit 0; the 3 controls stay green, so the fixture is proven to reach the sink |
| drop the leaf half of the parent check | `a_confined_rename_into_the_roots_own_ancestor_is_refused`, and only that one |
| confine the `Ancillary` walk too (blanket) | `an_ancillary_rename_may_still_leave_the_module` |
| refuse every symlink | `a_confined_rename_through_a_symlinked_dir_staying_inside_the_module_is_followed` and the other in-tree controls |

## Verification

- `cargo fmt --all -- --check` clean; `tools/ci/check_include_fmt.py` clean (498 files)
- `cargo clippy --workspace --all-targets --all-features --no-deps -- -D warnings` clean
- `cargo nextest run -p fast_io --all-features` - 787 passed
- `cargo nextest run -p engine --all-features` - 5046 passed
- `cargo nextest run -p transfer --all-features` - 2747 passed
- root-level `integration_delete_backup`, `server_backup_temp_dir`, `append_implies_inplace` - 29 passed
- `RUSTFLAGS=-D warnings cargo check --workspace --all-features --lib --tests --target x86_64-pc-windows-gnu` clean

## Adjacent, deliberately not fixed here

- `create_backup_path_parents` still creates the `--backup-dir` tree with `fs::create_dir_all`, so a `--backup-dir` symlink pointing at a *non-existent* outside path can still leave an empty directory outside the module. No data escapes - the link and rename that follow are now refused - and upstream confines that `mkdir` too, so it is worth a follow-up.
- `operator_mkdir` and `operator_symlink_metadata` remain structurally unconfined for the same reason the link and rename were: they go through the `Ancillary` parent walk. Their reachable call sites are `--partial-dir` creation and the alternate-basis stat, which upstream *does* confine.
- The `--delay-updates` refusal is logged and sets `IOERR_GENERAL`, but the client still exits 0 - the io_error does not reach the client's exit code on that path.

## Current status: one red cell, diagnosed but not yet attributed

`nextest` fails on exactly one test, identically on stable, beta and nightly, at head
`31fee35e`. The counts reconcile (32486 passed + 1 failed = 32487 run), so this is the
complete failure set rather than a fail-fast prefix.

The failing cell is this PR's own non-vacuity companion,
`an_ordinary_backup_dir_works_under_delay_updates`, which expects exit 0 and gets 23.

**The failure is not in the code this PR changes.** The daemon log names the syscall:

```
module 'data': landlock fully enforced over 1 root(s)
module 'data': seccomp BPF filter engaged (EPERM on unlisted syscalls)
```

`mkstemp` is temp-file creation. This PR routes the backup **link** and **rename** onto the
ownership walk and touches no temp-file path - confirmed by grepping the diff, whose only
matches for temp creation are inside the new test file itself. The transfer also dies early
(`receiving file list`, 123 bytes received), i.e. before the backup tier this PR modified
is reached at all.

Evidence that isolates the trigger:

| cell | flags | result |
|---|---|---|
| `an_ordinary_backup_dir_gets_the_pre_transfer_bytes` | `--backup-dir` | PASS |
| `an_ordinary_backup_dir_works_under_delay_updates` | `--backup-dir --delay-updates` | exit 23 at `mkstemp` |

So `--backup-dir` alone is fine; adding `--delay-updates` is the discriminator, and it fails
in the receiver's temp-file creation under Landlock.

**Open, and deliberately not claimed:** master carries no daemon + `--delay-updates`
executable test (the only match is a golden message TOML), so this cell is the first to
exercise that combination. That is consistent with a pre-existing daemon/Landlock defect this
PR merely exposes, but it is **not proven** - the discriminating measurement is the same
daemon push run against a master-built binary on a Landlock-enabled host, which has not been
run yet. Until that is done the failure should not be attributed either way.

### DIFF
--- a/crates/engine/src/local_copy/overrides.rs
+++ b/crates/engine/src/local_copy/overrides.rs
@@ -70,8 +70,12 @@ pub(super) fn create_hard_link(source: &Path, destination: &Path) -> io::Result<
 /// Hard-links a destination entry into the backup area.
 ///
 /// Same test-override seam as [`create_hard_link`], but on Unix production
-/// resolves both endpoints through the ownership walk
-/// ([`fast_io::operator_link`]). A backup area is an operator path, and the link
+/// resolves both endpoints through the ownership walk, bound to the session's
+/// confinement root ([`fast_io::operator_link_confined`]). Ownership alone is
+/// not enough: a non-chrooted daemon owns everything it creates, so a directory
+/// symlink standing at the `--backup-dir` is trusted-owned by construction and
+/// is followed by design; only the root can refuse where it lands.
+/// A backup area is an operator path, and the link
 /// tier runs *before* the rename tier, so confining only [`backup_rename`]
 /// would leave upstream's `backup-dir-symlink-race` escape wide open: the
 /// attacker's flipped parent redirects the link and the rename is never
@@ -95,7 +99,7 @@ pub(super) fn create_backup_hard_link(source: &Path, destination: &Path) -> io::
 
     #[cfg(unix)]
     {
-        fast_io::operator_link(source, destination)
+        fast_io::operator_link_confined(source, destination)
     }
     #[cfg(not(unix))]
     {
@@ -169,11 +173,14 @@ where
 /// Renames a destination entry to its backup location.
 ///
 /// On Unix both endpoints are resolved by the ownership walk
-/// ([`fast_io::operator_rename`]): every path component is inspected without
-/// following it, a symlink owned by uid 0 or our euid is followed as the
+/// ([`fast_io::operator_rename_confined`]): every path component is inspected
+/// without following it, a symlink owned by uid 0 or our euid is followed as the
 /// operator's own layout, and one owned by anyone else is refused. A backup
 /// directory may legitimately sit outside the transfer tree, so location cannot
-/// be the trust signal here - authority is.
+/// be the trust signal for WHETHER TO FOLLOW - authority is. It is still the
+/// signal for whether the landing site is acceptable when the session has a
+/// confinement root, which is what stops a followed trusted symlink from
+/// carrying an in-module file out of a served module.
 ///
 /// Without that, an attacker who can create entries inside the backup tree
 /// flips a parent component between a real directory and a symlink pointing
@@ -200,7 +207,7 @@ pub(super) fn backup_rename(from: &Path, to: &Path) -> io::Result<()> {
 
     #[cfg(unix)]
     {
-        fast_io::operator_rename(from, to, true)
+        fast_io::operator_rename_confined(from, to, true)
     }
     #[cfg(not(unix))]
     {
--- a/crates/fast_io/src/lib.rs
+++ b/crates/fast_io/src/lib.rs
@@ -417,10 +417,11 @@ pub use linux_capabilities::openat2_supported;
 pub use nofollow_open::open_basis_nofollow;
 #[cfg(unix)]
 pub use owner_walk::{
-    operator_create_dir_all, operator_link, operator_mkdir, operator_open_append,
-    operator_open_create_new, operator_open_read, operator_open_read_confined, operator_open_recv,
-    operator_open_rw_create, operator_open_write_create, operator_open_write_create_confined,
-    operator_read_to_string, operator_rename, operator_symlink_metadata, owner_trusted_parent,
+    operator_create_dir_all, operator_link, operator_link_confined, operator_mkdir,
+    operator_open_append, operator_open_create_new, operator_open_read,
+    operator_open_read_confined, operator_open_recv, operator_open_rw_create,
+    operator_open_write_create, operator_open_write_create_confined, operator_read_to_string,
+    operator_rename, operator_rename_confined, operator_symlink_metadata, owner_trusted_parent,
     owner_trusted_parent_kind, symlink_owner_is_trusted,
 };
 pub use refs_detect::{clear_refs_cache, is_refs_filesystem};
--- a/crates/fast_io/src/owner_walk.rs
+++ b/crates/fast_io/src/owner_walk.rs
@@ -183,6 +183,24 @@ impl AbsPathTracker {
         }
     }
 
+    /// The absolute path the walk has resolved, or `None` when the tracker is
+    /// disabled.
+    ///
+    /// The parent-walk entry point needs the value and not just the verdict:
+    /// upstream judges `<resolved parent>/<leaf>`, not the parent alone, so a
+    /// backup landing in an ANCESTOR of the root - which the parent-only test
+    /// reads as "still descending" - is still refused.
+    ///
+    /// upstream: `rsync-3.5.0/syscall.c:284-286` - `ona_open()`'s `out_abs`
+    /// out-parameter, which is what `owner_walk_parent()` builds `leafabs`
+    /// from.
+    fn resolved(&self) -> Option<&Path> {
+        match self {
+            Self::Disabled => None,
+            Self::Tracking { abspath } => Some(abspath),
+        }
+    }
+
     /// Refuse with `ELOOP` when the resolved path has left the confinement
     /// root.
     ///
@@ -344,6 +362,26 @@ fn owner_walk_open(
     mode: Mode,
     kind: crate::confinement::PathKind,
 ) -> io::Result<OwnedFd> {
+    let mut discarded = None;
+    owner_walk_open_tracked(path, flags, mode, kind, &mut discarded)
+}
+
+/// [`owner_walk_open`], additionally reporting where the walk actually landed.
+///
+/// `resolved` receives the physical absolute path of the opened component, and
+/// only while the tracker is live - a [`Confined`](crate::confinement::PathKind::Confined)
+/// walk in a session that has a confinement root. Every other walk leaves it
+/// `None`, which is upstream's `pabs[0] == '\0'` and means "nothing to judge".
+///
+/// upstream: `rsync-3.5.0/syscall.c:286` `ona_open()` with its `out_abs` /
+/// `out_cap` pair.
+fn owner_walk_open_tracked(
+    path: &Path,
+    flags: OFlags,
+    mode: Mode,
+    kind: crate::confinement::PathKind,
+    resolved: &mut Option<PathBuf>,
+) -> io::Result<OwnedFd> {
     // upstream: syscall.c:300-302 - "Opted out (local --insecure-links, or a
     // daemon module with `insecure links = yes`): restore the legacy
     // symlink-following open." The test sits at the top of ona_open(), so it
@@ -393,6 +431,7 @@ fn owner_walk_open(
                     // `--partial-dir`/`--temp-dir` shapes unconfined.
                     tracker.step(&name);
                     tracker.refuse_if_outside()?;
+                    *resolved = tracker.resolved().map(Path::to_path_buf);
                     return open_final(dirfd.as_fd(), name.as_os_str(), flags, mode);
                 }
                 Err(errno) => return Err(io::Error::from_raw_os_error(errno.raw_os_error())),
@@ -430,6 +469,7 @@ fn owner_walk_open(
             // than diverged.
             tracker.step(&name);
             tracker.refuse_if_outside()?;
+            *resolved = tracker.resolved().map(Path::to_path_buf);
             return open_final(dirfd.as_fd(), name.as_os_str(), flags, mode);
         }
         tracker.step(&name);
@@ -451,6 +491,7 @@ fn owner_walk_open(
     // `O_PATH`, which names a location and is not a working descriptor. The
     // reopen is the same open the caller would have got before, and it is
     // subject to the sandbox exactly as that one was.
+    *resolved = tracker.resolved().map(Path::to_path_buf);
     if by_location {
         return rustix::fs::openat(dirfd.as_fd(), ".", flags | OFlags::CLOEXEC, mode)
             .map_err(|errno| io::Error::from_raw_os_error(errno.raw_os_error()));
@@ -477,7 +518,8 @@ pub fn owner_trusted_parent(path: &Path) -> io::Result<(OwnedFd, OsString)> {
     owner_trusted_parent_kind(path, crate::confinement::PathKind::Ancillary)
 }
 
-/// [`owner_trusted_parent`] with the caller's [`PathKind`](crate::confinement::PathKind).
+/// [`owner_trusted_parent`] with the caller's
+/// [`PathKind`](crate::confinement::PathKind).
 ///
 /// The parent walk and the confinement judgement are one decision upstream -
 /// `owner_walk_parent()` reads `operator_path_resolve`, which its call sites
@@ -486,13 +528,28 @@ pub fn owner_trusted_parent(path: &Path) -> io::Result<(OwnedFd, OsString)> {
 /// [`Ancillary`](crate::confinement::PathKind::Ancillary) spelling and keeps
 /// its existing callers unchanged.
 ///
+/// A [`Confined`](crate::confinement::PathKind::Confined) walk additionally
+/// carries upstream's SECOND check: once the parent is resolved, the LEAF's
+/// would-be absolute path is judged against the confinement root. Judging the
+/// parent alone is not the same rule - an absolute walk is allowed to pass
+/// through the root's own ancestors on the way down, so a parent resolving to
+/// `/srv` under a root of `/srv/mod` reads as "still descending" while
+/// `/srv/<leaf>` has plainly diverged. An ancillary walk leaves the tracker
+/// disabled, so it reports no resolved path and that second check cannot fire
+/// for it.
+///
 /// upstream: `rsync-3.5.0/syscall.c:558` `owner_walk_parent()`, and
 /// `rsync-3.5.0/syscall.c:552` `int operator_path_resolve = 0;` - the flag this
-/// parameter replaces.
+/// parameter replaces. The leaf judgement is `rsync-3.5.0/syscall.c:581-596` -
+/// the `snprintf(leafabs, "%s/%s", pabs, *bname)` and the
+/// `abspath_outside_confinement(leafabs)` that follows it, both gated on that
+/// same flag.
 ///
 /// # Errors
 ///
-/// See [`owner_trusted_parent`].
+/// - `ELOOP` when the walk refuses an untrusted-owner symlink, or when the
+///   resolved leaf lands outside the confinement root.
+/// - Otherwise see [`owner_trusted_parent`].
 pub fn owner_trusted_parent_kind(
     path: &Path,
     kind: crate::confinement::PathKind,
@@ -501,12 +558,23 @@ pub fn owner_trusted_parent_kind(
         return Err(io::Error::from_raw_os_error(libc::EINVAL));
     };
     let parent = path.parent().unwrap_or_else(|| Path::new(""));
-    let dirfd = owner_walk_open(
+    let mut parent_abs = None;
+    let dirfd = owner_walk_open_tracked(
         parent,
         OFlags::RDONLY | OFlags::DIRECTORY,
         Mode::empty(),
         kind,
+        &mut parent_abs,
     )?;
+    // upstream: `if (pabs[0])` - an untracked walk has nothing to judge.
+    if let Some(abs) = parent_abs
+        && crate::confinement::outside_session_root(
+            &abs.join(&leaf),
+            crate::confinement::PathKind::Confined,
+        )
+    {
+        return Err(io::Error::from_raw_os_error(libc::ELOOP));
+    }
     Ok((dirfd, leaf))
 }
 
@@ -588,7 +656,7 @@ pub fn operator_create_dir_all(path: &Path, mode: u32) -> io::Result<()> {
 ///
 /// # Upstream Reference
 ///
-/// - `rsync-3.5.0/syscall.c:1894` `do_rename_at()` under `operator_path_resolve`
+/// - `rsync-3.5.0/syscall.c:1891` `do_rename_at()` under `operator_path_resolve`
 ///   - `owner_walk_parent` on each side, then `renameat`.
 /// - `rsync-3.5.0/backup.c:200-219` `make_backup()` - the caller that sets the
 ///   operator-path mode around the backup rename.
@@ -608,6 +676,43 @@ pub fn operator_rename(old_path: &Path, new_path: &Path, replace: bool) -> io::R
     )
 }
 
+/// [`operator_rename`] additionally bound to the session's confinement root.
+///
+/// The rename tier of the same `--backup-dir` backup as
+/// [`operator_link_confined`], and the tier that actually runs wherever the
+/// link tier cannot: a `--backup-dir` on another mount, a non-regular
+/// pre-image, or a receiver whose nested-path link is refused by its own
+/// beneath-confinement. A trusted-owned directory symlink standing at the
+/// backup dir sends the destination's pre-transfer bytes out of the module
+/// wholesale, since the rename MOVES the inode rather than copying it.
+///
+/// # Upstream Reference
+///
+/// - `rsync-3.5.0/backup.c:443-449` `make_backup()` - `operator_path_resolve =
+///   1` around the whole backup.
+/// - `rsync-3.5.0/syscall.c:1891` `do_rename_at()` under
+///   `operator_path_resolve`.
+///
+/// # Errors
+///
+/// - `ELOOP` when a component is an untrusted-owner symlink, or when either
+///   resolved endpoint lands outside the confinement root. Deliberately not
+///   `EXDEV`: callers treat that as cross-device and fall back to copy+remove,
+///   which would launder the refusal.
+/// - Otherwise the `renameat(2)` errno.
+pub fn operator_rename_confined(old_path: &Path, new_path: &Path, replace: bool) -> io::Result<()> {
+    let kind = crate::confinement::PathKind::Confined;
+    let (old_dirfd, old_leaf) = owner_trusted_parent_kind(old_path, kind)?;
+    let (new_dirfd, new_leaf) = owner_trusted_parent_kind(new_path, kind)?;
+    crate::renameat(
+        old_dirfd.as_fd(),
+        &old_leaf,
+        new_dirfd.as_fd(),
+        &new_leaf,
+        replace,
+    )
+}
+
 /// Hard-link `old_path` to `new_path` with both endpoints resolved by the
 /// ownership walk.
 ///
@@ -617,9 +722,9 @@ pub fn operator_rename(old_path: &Path, new_path: &Path, replace: bool) -> io::R
 ///
 /// # Upstream Reference
 ///
-/// - `rsync-3.5.0/backup.c:200-207` `link_or_rename()` - `do_link_at` first,
+/// - `rsync-3.5.0/backup.c:226-247` `link_or_rename()` - `do_link_at` first,
 ///   `do_rename_at` on failure.
-/// - `rsync-3.5.0/syscall.c:676` `do_link_at()` under `operator_path_resolve` -
+/// - `rsync-3.5.0/syscall.c:961` `do_link_at()` under `operator_path_resolve` -
 ///   `owner_walk_parent` on each side, then `linkat`.
 ///
 /// # Errors
@@ -630,6 +735,38 @@ pub fn operator_rename(old_path: &Path, new_path: &Path, replace: bool) -> io::R
 pub fn operator_link(old_path: &Path, new_path: &Path) -> io::Result<()> {
     let (old_dirfd, old_leaf) = owner_trusted_parent(old_path)?;
     let (new_dirfd, new_leaf) = owner_trusted_parent(new_path)?;
+    crate::linkat(old_dirfd.as_fd(), &old_leaf, new_dirfd.as_fd(), &new_leaf)
+}
+
+/// [`operator_link`] additionally bound to the session's confinement root.
+///
+/// The `--backup-dir` link tier. Ownership alone cannot bound this: a
+/// non-chrooted daemon writes everything it creates as its own uid, so a
+/// directory symlink standing where the operator named `--backup-dir` is
+/// TRUSTED-owned by construction and the walk follows it by design. The link
+/// then deposits a second name for the destination's pre-transfer inode outside
+/// the module, and the temp-to-destination rename that follows leaves that name
+/// holding the only copy of the pre-image. Ownership decides whether to follow;
+/// the root decides whether the landing site is acceptable.
+///
+/// # Upstream Reference
+///
+/// - `rsync-3.5.0/backup.c:443-449` `make_backup()` - `operator_path_resolve =
+///   1` around the WHOLE backup, so it covers `link_or_rename()`'s link tier
+///   and not only the rename it falls back to.
+/// - `rsync-3.5.0/backup.c:226-247` `link_or_rename()` - `do_link_at` first.
+/// - `rsync-3.5.0/syscall.c:961` `do_link_at()` under `operator_path_resolve`.
+///
+/// # Errors
+///
+/// - `ELOOP` when a component is an untrusted-owner symlink, or when either
+///   resolved endpoint lands outside the confinement root.
+/// - Otherwise the `linkat(2)` errno - notably `EXDEV`, which the caller treats
+///   as its signal to fall through to the rename tier.
+pub fn operator_link_confined(old_path: &Path, new_path: &Path) -> io::Result<()> {
+    let kind = crate::confinement::PathKind::Confined;
+    let (old_dirfd, old_leaf) = owner_trusted_parent_kind(old_path, kind)?;
+    let (new_dirfd, new_leaf) = owner_trusted_parent_kind(new_path, kind)?;
     crate::linkat(old_dirfd.as_fd(), &old_leaf, new_dirfd.as_fd(), &new_leaf)
 }
 
--- a/crates/fast_io/tests/operator_open_module_confinement.rs
+++ b/crates/fast_io/tests/operator_open_module_confinement.rs
@@ -282,3 +282,167 @@ fn without_a_confinement_root_a_confined_open_still_follows() {
 
     assert_eq!(read_all(file), "SECRET-MARKER");
 }
+
+/// THE RENAME-SIDE PIN, at the shape a `--backup-dir` actually has: the escape
+/// is a trusted-owned DIRECTORY symlink standing where the operator named the
+/// backup area, not a symlink at the backup leaf. `renameat` never follows a
+/// leaf symlink - it replaces the name - so a leaf plant cannot express this
+/// escape at all, and only the parent walk can refuse it.
+///
+/// A refused rename must also leave the source in place: the rename MOVES the
+/// inode, so a refusal that fired after the syscall would leave nothing behind
+/// to read back.
+///
+/// upstream: `rsync-3.5.0/backup.c:443-449` `make_backup()` -
+/// `operator_path_resolve = 1` around the whole backup; `syscall.c:1891`
+/// `do_rename_at()` walks each side with `owner_walk_parent()` while it is set.
+#[test]
+fn a_confined_rename_through_a_symlinked_dir_leaving_the_module_is_refused() {
+    let fixture = fixture();
+    let pre_image = fixture.module.join("payload");
+    fs::write(&pre_image, "PRE-IMAGE-IN-MODULE").expect("seed the in-module file");
+    let outside_dir = fixture.secret.parent().expect("outside dir").to_path_buf();
+    let backup_dir = fixture.module.join("backup/out");
+    symlink(&outside_dir, &backup_dir).expect("plant the out-of-module dir symlink");
+
+    let _session = serve_module(&fixture.module);
+    let error = fast_io::operator_rename_confined(&pre_image, &backup_dir.join("payload"), true)
+        .expect_err("a confined rename must not land outside the module");
+
+    assert_eq!(
+        error.raw_os_error(),
+        Some(libc::ELOOP),
+        "the refusal must be ELOOP: callers treat EXDEV as cross-device and \
+         fall back to copy+remove, which would launder it"
+    );
+    assert_eq!(
+        fs::read_to_string(&pre_image).expect("the in-module file must survive"),
+        "PRE-IMAGE-IN-MODULE",
+        "a refused rename must not have moved the source"
+    );
+    assert!(
+        !outside_dir.join("payload").exists(),
+        "nothing may have been deposited outside the module"
+    );
+}
+
+/// The LINK tier of the same backup. It runs BEFORE the rename
+/// (`link_or_rename()` with `prefer_rename = 0`), so confining only the rename
+/// leaves the escape open wherever the link succeeds - which is the ordinary
+/// same-filesystem case.
+///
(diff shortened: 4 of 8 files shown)


oferchen:
### CI red is this PR's own regression, and it is Linux-only

Two cells fail — `nextest (nightly)` and
`Feature flag combinations / no-incremental-flist (linux)` — on the **same
single test**, with the same assertion:

```
FAIL bin::delay_updates_backup_failure a_placeable_backup_still_updates_the_destination
tests/delay_updates_backup_failure.rs:193
assertion `left == right` failed: the delayed rename must still land when the backup succeeds
```

Two independent cells with one cause, so this is not a nightly-toolchain
artefact.

**Why it matters more than a single red:** that test is the declared
*non-vacuity companion* for the cell directly above it. The sibling pins that an
**unplaceable** backup skips the delayed rename (upstream `receiver.c:694`
`continue`); this one exists to prove that pin is not passing merely because the
fixture can never transfer anything. So the failure says the success path now
behaves like the failure path — a writable `--backup-dir` places the backup, and
the delayed rename no longer lands.

**Attribution.** The test file is untouched by this branch, but the branch does
change the exact path it exercises: `backup_rename_syscall` went from a bare
`fs::rename` to `operator_rename_confined`, and `backup_hardlink_syscall` gained
a `new_path.parent() == Some(dest_dir)` guard that reroutes every `--backup-dir`
shape onto the confined resolver. If the confined rename reports failure where
the plain rename succeeded, `backup_rename_or_copy` returns "not placed" and the
delayed rename is skipped — which is precisely the observed outcome.

**Platform.** Measured locally on macOS at this branch head:
`cargo nextest run --test delay_updates_backup_failure --all-features` →
**2 tests run, 2 passed**. So the two arms of `operator_rename_confined`
disagree, and only the Linux arm refuses. That is the same platform-arm
divergence class as #7349 (task 603) and the Linux-only gap in task 613 — a
macOS green here is evidence the arms differ, not evidence the change is sound.

Next step is a Linux repro to see which refusal fires: with no confinement root
installed for a plain local push, the doc comment added in this PR says only the
ownership half should apply, and every component in the fixture is euid-owned.
If it refuses anyway, the comment describes an intent the code does not
implement.

oferchen:
### Escalation: a REQUIRED context is now red, and it is a second, independent symptom

`upstream-testsuite-tcp / upstream testsuite` — a required context — has now
failed, so this PR is blocked rather than merely carrying non-required reds.
It is a different instrument from the nextest cells and it reports two
**expected-result mismatches**:

```
----- expected-result mismatches:
      operator-path-backup-dir-exclude-daemon: expected pass, got fail
      operator-path-dir-daemon-leaf:           expected pass, got fail
overall result is 2
```

Both are `expected pass, got fail`, i.e. cells the committed manifest says pass
on master. These are regressions from this branch, not pre-existing divergences,
and the manifest was verified current at master in task 712 / task 842.

**One story fits all three cells: the new confinement OVER-REFUSES.**

| cell | path | shape |
|---|---|---|
| `a_placeable_backup_still_updates_the_destination` | local `--backup-dir` + `--delay-updates` | placeable backup treated as unplaceable, so the delayed rename is skipped |
| `operator-path-backup-dir-exclude-daemon` | daemon `--backup-dir` | legitimate backup destination refused |
| `operator-path-dir-daemon-leaf` | daemon operator path | legitimate leaf refused |

That is the mirror image of the risk this PR is guarding against: the change
correctly stops a trusted-owned dirlink carrying data out of a module, but as
written it also refuses backup destinations that must be allowed. Note the two
daemon cells *do* have a module root, so this is not only the
no-confinement-root case — the refusal fires where a root exists too.

**Please do not re-baseline those two manifest rows.** They are `expected pass`
because upstream passes them; flipping them would convert a regression into a
recorded divergence. Same reasoning as task 907, where an XPASS was proven
environmental and the row deliberately left alone.

Local macOS remains 2/2 green on the nextest pair, so none of this reproduces
off Linux — the next step is a Linux repro to see which arm of the confined
resolver refuses, and on what component.

oferchen:
### Decisive: this PR's OWN positive controls fail

A fourth cell (`Feature flag combinations / compression (linux)`) ran far deeper
than the others — 14200 of 14537 tests, versus 1329 in `nextest (nightly)` —
and it surfaces a **different pair** of failures, both in this branch's own new
test file:

```
FAIL transfer::daemon_backup_dir_symlink_confinement an_ordinary_backup_dir_gets_the_pre_transfer_bytes
FAIL transfer::daemon_backup_dir_symlink_confinement a_backup_dir_symlink_staying_inside_the_module_is_still_followed
```

Both are positive controls: *an ordinary backup dir*, and *an in-module symlink
that must still be followed*. The second is the exact case the PR's own doc
comment says must keep working — "a symlink owned by uid 0 or our euid is
followed as the operator's own layout".

So the confinement refuses the very shapes this PR documents as legitimate. That
is a complete diagnosis from the branch's own tests; no external oracle needed.

**Consolidated: 4 red cells, 5 distinct failing tests, one cause.**

| cell | tests | verdict |
|---|---|---|
| `nextest (nightly)` | `a_placeable_backup_still_updates_the_destination` | over-refusal (local) |
| `Feature flag combinations / no-incremental-flist` | same | same |
| `Feature flag combinations / compression` | the two `daemon_backup_dir_symlink_confinement` controls | over-refusal (daemon) |
| `upstream-testsuite-tcp` *(required)* | `operator-path-backup-dir-exclude-daemon`, `operator-path-dir-daemon-leaf` | over-refusal (daemon) |

⚠ **Method note worth keeping:** each cell truncates at a different point under
nextest fail-fast, so each showed a *different subset*. Reading any one cell's
list as "the failures" would have understated this to a single test. A failing
cell's test list is a lower bound (task 1051).

**Not attempted:** a Linux repro. `172.16.1.74` currently has another
upstream-testsuite run live (`ts-work/target/release/oc-rsync --daemon`) and is
at 98% disk with 13 GB free. Starting a competing build there would republish
the shared unversioned binary path that run depends on (task 983), so the repro
waits for the host to clear rather than corrupting someone else's measurement.

oferchen:
Rebased onto master now that the ownership-walk Landlock traversal fix (#7541, `243836639`) has landed.

**Why this is a re-test, not just a refresh.** My earlier diagnosis on this PR concluded the reds were this branch's own over-refusal. That diagnosis was made *before* #7541 was on master, and several of the failing cells are daemon backup-dir tests — exactly the shape #7541 fixes. A non-daemon `--server` receiver installs Landlock too, so even the `bin::delay_updates_backup_failure` cell is not automatically outside its reach. Rather than defend the attribution, this rebase lets CI settle it.

Two outcomes, both informative:

- **Reds clear** — the over-refusal attribution was wrong; the cause was the traversal EACCES, and this PR needs no code change.
- **Reds persist** — the attribution stands, and the remaining failures are this branch's own. In particular `an_ordinary_backup_dir_gets_the_pre_transfer_bytes` and `a_backup_symlink_staying_inside_the_module_is_still_followed` are this PR's *own positive controls*; if they still fail, the confinement is refusing in-module operations it is documented to allow.

**Conflict resolution.** `crates/fast_io/src/owner_walk.rs` conflicted in one hunk — the no-components arm, where #7541 added the `by_location` reopen and this branch added `*resolved = tracker.resolved()...`. Both are kept. The assignment is placed *before* the early return, matching the `is_last` arm, which sets `*resolved` immediately before its own `return`; putting it after would leave the out-param unset on the `by_location` path.

Verified locally before pushing: `cargo check -p fast_io` clean, and `fast_io` 794/794 (`passed + skipped == run`, so no fail-fast truncation). macOS cannot exercise the Landlock arm, so Linux CI is the discriminator here.

Refs task 1053, task 1036, task 1048.

oferchen:
**Attribution settled: my over-refusal diagnosis was wrong.**

I previously concluded this PR's reds were its own over-refusal regression. The rebase onto the ownership-walk Landlock traversal fix (`243836639`) has refuted that.

The discriminator is `Feature flag combinations / compression (linux)`. That is the cell I read the two failing positive controls from:

- `an_ordinary_backup_dir_gets_the_pre_transfer_bytes`
- `a_backup_symlink_staying_inside_the_module_is_still_followed`

Both are in-module cases this PR's own confinement is documented to allow, and both are in this PR's own new test file. **That cell is now green on the rebased head**, so both pass. Nothing in this branch changed — only its base did.

Also green now, having previously been red or unreported: `upstream-testsuite` and `upstream-testsuite (root)`, plus both `upstream-testsuite-tcp` legs. Four of the slowest required contexts, zero failures anywhere on the PR.

**Why I got it wrong.** I made the call before `243836639` was on master, then read a failing test list as if it named the cause. Several of the failing cells were daemon backup-dir tests — exactly the shape the traversal fix addresses — and I did not weigh that. The specific error, `mkstemp "payload" (in data) failed: Permission denied (13)`, is an EACCES, which is the traversal signature, not a confinement refusal; a genuine over-refusal from this PR would have surfaced as ELOOP or EXDEV. The errno was in front of me and I read the outcome instead of the cause.

That is the same trap recorded on the fix itself: the escape pin passes in every broken arm, so the discriminator is the *cause*, not the outcome.

Residual: `nextest (stable)`, `macOS (stable)`, `Windows (stable)` and `Linux musl (stable)` have not reported yet. The attribution for the positive controls is settled; full green is not yet claimed.

oferchen:
**Retracting my previous comment.** It claimed the rebase had settled the attribution and that the two positive controls now pass. That was wrong, and the evidence I used was invalid.

`nextest (stable)` — a required context — is red on the rebased head with **three** failures, all in this PR's own new test file:

```
transfer::daemon_backup_dir_symlink_confinement an_ordinary_backup_dir_gets_the_pre_transfer_bytes
transfer::daemon_backup_dir_symlink_confinement an_ordinary_backup_dir_works_under_delay_updates
transfer::daemon_backup_dir_symlink_confinement a_backup_dir_symlink_staying_inside_the_module_is_still_followed
```

All three: `rsync: [receiver] mkstemp "payload" (in data) failed: Permission denied (13)`, exit 23 where 0 is expected. That is the **same error as before the rebase**, so `243836639` did not fix it.

**Why my evidence was invalid.** I cited `Feature flag combinations / compression (linux)` going green. That cell runs

```
-p compress -p protocol -p engine -p transfer --features zstd,lz4
```

It does include `-p transfer`, so it compiles and runs this test file — but its feature set does **not** include `landlock`. `nextest (stable)` runs `--workspace --all-features`, which does. So the two cells disagree because only one of them can exercise the sandbox at all, and I read the green one as though it covered the same ground.

**What is actually established:**

- The three failures reproduce only with the `landlock` feature enabled.
- The errno is EACCES, which is a reachability failure, not a confinement refusal — consistent with my original reading of the error and inconsistent with my later claim that the traversal fix resolved it.
- A full macOS run of `-p engine -p transfer --all-features --no-fail-fast` is 7809/7809, so nothing here is platform-independent.

**What is not established:** whether this is this PR's new confined operations failing under the sandbox, or a pre-existing gap in the module's Landlock ruleset that only these new tests reach. That needs a Linux run with `landlock` enabled; I have not made that measurement and am not going to guess at it.

oferchen:
**This failure is shared with the sibling PR — it is one defect, not two.**

Both open confinement PRs are red in `nextest (stable)` with the same error, in their own separately-written test files:

| PR | failing tests | error |
|---|---|---|
| #7538 | `daemon_backup_dir_symlink_confinement`: `an_ordinary_backup_dir_gets_the_pre_transfer_bytes`, `an_ordinary_backup_dir_works_under_delay_updates`, `a_backup_dir_symlink_staying_inside_the_module_is_still_followed` | `mkstemp "payload" (in data) failed: Permission denied (13)` |
| #7535 | `daemon_inplace_backup_dir_confinement`: `an_unplanted_backup_leaf_gets_the_pre_transfer_bytes`, `a_backup_symlink_staying_inside_the_module_is_still_followed` | same |

All five are **positive controls** — in-module operations the confinement is documented to allow — and all exit 23 where 0 is expected. Two authors, two fixtures, one errno, one option combination. That cross-PR agreement is the evidence that the cause is shared rather than per-PR.

**What is established:**

- Reproduces **only** with the `landlock` feature. `Feature flag combinations / compression (linux)` runs these same test files under `--features zstd,lz4` — no landlock — and is green. `nextest (stable)` runs `--all-features` and is red. The green cell cannot exercise the sandbox, so it is not coverage.
- **Not** a blanket sandbox denial: ~32,469 tests pass alongside in the same run.
- macOS is fully green (7809/7809 for the affected crates). Landlock is Linux-only, so that is expected rather than reassuring.
- Both test files are **absent on master**, so neither set has ever been green. These are the first tests of *daemon push + `--backup-dir` under Landlock*.

⚠ Both runs were fail-fast truncated (817 and 815 tests never ran), so five is a lower bound.

**Open, and deliberately not guessed at:** whether the behaviour is already broken on master — these tests being correct and exposing a pre-existing ruleset gap — or whether the new confined operations cause it. Prior art points at the former: the daemon walk starting at `/` and being EACCESed (task 608), Landlock setup failing under `use chroot` with chroot silently carrying the module (task 1050), and module re-resolution EACCESing after the privilege drop (task 640).

Settling it needs a Linux run with `--features landlock`; I have not made that measurement. Tracked as task 1056. **Neither PR should be patched in isolation until it is answered** — a per-PR fix would mask a shared defect.

oferchen:
## Correction: the `landlock` attribution in my earlier comment is wrong

I previously wrote that this failure "reproduces only with the `landlock` feature". **That is refuted.** Retracting it here so the wrong attribution does not get acted on.

**Why it is wrong.** `landlock` is not a root-package feature. `crates/cli/Cargo.toml:103` and `crates/daemon/Cargo.toml:114` both pin it directly:

```toml
[target.'cfg(target_os = "linux")'.dependencies]
fast_io = { path = "../fast_io", default-features = false, features = ["landlock"] }
```

That is unconditional on Linux — not behind any cargo feature. So Landlock is compiled and engaged in **every** Linux workspace build, including the ones that pass.

**The actual measured discriminator**, from this PR's own check rollup:

| Linux cell | command | these tests |
|---|---|---|
| `Feature flag combinations / default-features (linux)` | `cargo nextest run --locked --profile ci --workspace` | **PASS** |
| `nextest (stable)` | `cargo nextest run --locked --profile ci --workspace --all-features` | **FAIL** |

Both are Linux, both build the `oc-rsync` binary, both run the full workspace with no test filter (`_test-features.yml:269` is a bare `cargo nextest run --locked --profile ci ${{ matrix.args }}`), and both have Landlock live. The only difference is the non-default features.

So the cause is somewhere in what `--all-features` adds. Root package: `zlib-ng`, `iouring-send-zc`, `mmap-free-basis`, `openssl`, `openssl-vendored`, `quic`, `sd-notify`. Plus every workspace member's non-default features, which under one `--workspace --all-features` build unify into the binary — notably **`daemon/daemon-seccomp`**, which is off in the passing cell.

**Two candidates checked and excluded so far:**

- **seccomp directly** — its default action is `Errno(EPERM)` (errno 1); the observed failure is `Permission denied (13)` = EACCES. `seccomp.rs:4` states the distinction explicitly: "Landlock denies path-based syscalls with EACCES; seccomp … EPERM".
- **seccomp → umask → mode 000** (the known indirect route to EACCES: `umask(2)` is absent from the allowlist, a lazy read returns -1, `dflt_perms` collapses to 0). Already closed on master by `sanitize_umask` + eager `init_orig_umask` (`crates/metadata/src/apply/permissions.rs:60-135`), which rejects any value outside `0o777` and falls back to `0o022`.

**What settles it** is a two-run discriminator on a Linux host, running just these test files:

1. `--workspace --all-features` → expect FAIL (reproduces)
2. same, minus `daemon/daemon-seccomp` → PASS or FAIL splits the remaining space in half

I do not have a usable Linux host at the moment (the shared one is at load 11.5 with 13 GB free and another testsuite run in flight), so this is measured-so-far, not concluded.

**This PR should still not be patched in isolation** — the same failure hits the sibling PR identically in its own separately-written test file, so a per-PR fix would mask a shared defect.

oferchen:
## Phase 2 (pattern analysis): the fixture shape is excluded, and two further findings

Following the correction above, I compared the failing fixture against the working examples instead of continuing to reason about feature flags.

### 1. The config shape is NOT the differentiator

All 13 master daemon integration tests (`daemon_dest_mode_parity`, `daemon_push_refused_file_reports_error`, `daemon_total_transferred_size_stats`, `delay_updates_daemon_pull`, `msg_no_send_unsendable_file`, `verify_redo_recovers_network_paths`, …) use the **same** config shape as this fixture:

```
pid file = <tmpdir>/rsyncd.pid      # outside the module root
log file = <tmpdir>/rsyncd.log      # outside the module root
use chroot = false
[data]
path = <tmpdir>/module
```

Same tempdir module, same `use chroot = false`, same pid/log files outside every Landlock-granted root. Those tests pass under `--all-features`. So "log/pid outside the module root" and "unprivileged tempdir daemon" are both excluded.

### 2. The one difference is `--backup-dir`, and it has ZERO integration coverage on master

**No master integration test passes `--backup-dir` (or `--temp-dir` / `--partial-dir` / `--compare-dest` / `--copy-dest` / `--link-dest`) over a real `rsync://` daemon connection.** The only in-tree daemon test naming one is `crates/daemon/src/tests/chunks/daemon_compare_dest_push.rs`, which is an in-crate test and does not go through the real connection path.

That matters structurally. `engage_landlock_sandbox` has exactly **one** production call site:

```
crates/daemon/src/daemon/sections/module_access/transfer/orchestration.rs:590
    if !engage_landlock_sandbox(ctx, module, &extra_allowed)? {
```

and `extra_allowed` is populated only from those staging / alt-dest options (`sandbox.rs:242`). So `roots.len() > 1` — the multi-root Landlock ruleset on a live daemon — **is reached for the first time by these two PRs**. Every other daemon test engages Landlock with exactly one root.

So the honest characterisation is a **conjunction**: `--backup-dir` over a daemon (previously untested) *and* some non-default feature. `--backup-dir` alone is not sufficient — the `default-features (linux)` cell runs these same tests and passes.

### 3. ⚠ Two of the five tests pass VACUOUSLY

Of the five failing-file tests, the three that FAIL are the positive controls. The two that PASS —
`a_backup_dir_symlink_leaving_the_module_must_not_receive_the_pre_image` and
`a_delayed_update_backup_must_not_leave_the_module` — assert a **refusal**, and a transfer that aborts at EACCES also fails to escape. They cannot distinguish "the confinement refused the escape" from "the transfer died before reaching the backup".

This is the same vacuity already recorded for #7535's escape pin. **Even once the EACCES is resolved, the escape assertions need to be re-confirmed non-vacuous** — otherwise the security fix these PRs exist for is unverified on Linux.

### Status

The two-run discriminator (`--workspace` control vs `--workspace --all-features`, same host, same two test binaries) is building now on an idle aarch64 Linux box. aarch64 is a caveat: a reproduction there is decisive, a non-reproduction is not, since some candidates (e.g. the `*at`-only seccomp allowlist) are x86_64-specific by construction.

oferchen:
## Phase 3: measured on Linux. The CI configuration does NOT reproduce on aarch64 — and that asymmetry is the lead.

I built the `oc-rsync` binary in three feature configurations on an idle aarch64 Linux box and drove the *ordinary* case directly (`--backup --backup-dir=bak --ignore-times`, real `bak` directory, `use chroot = false`, tempdir module) — the same shape as `an_ordinary_backup_dir_gets_the_pre_transfer_bytes`.

Crucially I read the **daemon's own log**, which reports its sandbox state. No test does this today, and it is what made the result interpretable.

| cell | landlock | seccomp | client exit | backup |
|---|---|---|---|---|
| `--workspace` (default features) | `fully enforced over 1 root(s)` | `unavailable in this build` | **0** | correct |
| `-p bin --all-features` | `fully enforced over 1 root(s)` | `unavailable in this build` | 0 | correct |
| `--workspace --all-features` | `fully enforced over 1 root(s)` | **`BPF filter engaged (EPERM on unlisted syscalls)`** | **0** | correct |

Three findings.

**1. ⚠ My middle cell was INERT, and the log caught it.** `cargo build -p bin --all-features` enables all features of the **`bin` package only** — not every member's. `cargo nextest run --workspace --all-features` (what CI runs) enables `daemon/daemon-seccomp`; `-p bin --all-features` does not. Both binaries logged `seccomp unavailable`, which is how I noticed. Worth stating plainly because the two spellings look interchangeable and are not.

**2. Seccomp being engaged is NOT sufficient.** The third cell has the exact CI feature configuration, seccomp genuinely installed, and it still exits 0 with a correct backup. So `daemon-seccomp` alone does not explain the failure.

**3. `--backup-dir` does NOT widen the Landlock ruleset.** All three cells log **`over 1 root(s)`**. My earlier "these PRs are the first to reach a multi-root ruleset" claim is wrong — `extra_allowed_paths` stays empty here, and `bak` is covered because it already sits inside the module root.

### The lead: this is arch-asymmetric, and the seccomp allowlist is arch-asymmetric in exactly the same way

`crates/daemon/src/daemon/sections/seccomp.rs` allowlists **only the `*at` syscall forms** — `openat`, `openat2`, `newfstatat`, `statx`, `renameat`, `renameat2`, `unlinkat`, `mkdirat`, `linkat`, `symlinkat`, `readlinkat`, `fchmodat`, `fchownat`. The legacy `open` / `stat` / `rename` / `unlink` / `mkdir` / `link` are absent, and the only `#[cfg(target_arch = "x86_64")]` additions are `arch_prctl` plus portability shims for `rseq` / `sendfile` / `fadvise64` — **no legacy syscalls are added back on x86_64**.

On **aarch64 those legacy syscalls do not exist**, so a `*at`-only allowlist is complete by construction. On **x86_64 they do exist**, and anything issuing one gets `EPERM`. That predicts exactly what is observed: green on aarch64 with seccomp engaged, red on x86_64 CI.

**This is a hypothesis, not a conclusion.** The unresolved half is the errno: seccomp's default action is `EPERM` (1) and the failure reports `EACCES` (13), so if this is the cause there is an intermediate step — a fallback path, or a later Landlock-checked operation reached only after an earlier legacy syscall was refused. I am not asserting the mechanism.

**What settles it:** the same manual repro on an **x86_64** host with a `--workspace --all-features` binary, reading the daemon log and `strace`. Our x86_64 box is at 98 % disk with another testsuite run in flight, so I have not run it — I filled a different host's disk earlier today with a `--workspace` nextest build and will not repeat that on a box with 13 G free.

### Two things worth fixing regardless of the root cause

- **The fixture discards the daemon's own diagnosis.** `rsyncd.log` carries the landlock root count, the enforcement tier and the seccomp state; the tempdir is dropped and no test reads it. Surfacing it on failure would have made this a one-run diagnosis instead of a day's work.
- **Two of the five tests still pass vacuously** (the escape assertions — an EACCES abort also fails to escape). That must be re-confirmed once the transfer succeeds.

oferchen:
## Failure set measured on the post-rebase commit (eefcd7242), x86_64 CI

`nextest (stable)` — `Summary [470.5s] 32472/33289 run: 32469 passed, 3 failed, 145 skipped`.
All three failures are in this PR's own new test file, and **all three are positive
controls**, not the escape assertion:

| test | asserts |
|---|---|
| `an_ordinary_backup_dir_gets_the_pre_transfer_bytes` | a legitimate in-module `--backup-dir` works |
| `an_ordinary_backup_dir_works_under_delay_updates` | same, under `--delay-updates` |
| `a_backup_dir_symlink_staying_inside_the_module_is_still_followed` | an in-module symlink is still followed |

Identical error in all three:

```
```

That is the **receiver's destination temp file** in the module root — not the backup
path — failing with EACCES before any backup logic runs. So the cells are not
disagreeing about confinement policy; the transfer never gets far enough to have an
opinion.

⚠ The list is a lower bound: 817 of 33289 tests did not run (fail-fast).

### Two hypotheses examined and refuted

- **seccomp EPERM.** The default action is `EPERM(1)`; the observed errno is
  `EACCES(13)`. Different value, and no intermediate step has been demonstrated.
- **The `umask(2)` route** (blocked syscall → `cached_umask()` = `u32::MAX` →
  `dflt_perms` = 0 → mode-000 destination → later EACCES). Closed at the source:
  `cached_umask()` falls back to a lazy capture, and `sanitize_umask()` maps any
  value with bits outside `0o777` to `FALLBACK_UMASK = 0o022`. It cannot yield 0.

### The next discriminator

The fixture shape here — daemon push, `--backup-dir` present — is the same one
already recorded as failing *before* this branch existed. The question this PR
cannot answer from its own red is whether the fixture trips a pre-existing defect
or one introduced here.

The test that settles it: apply **only this PR's test file** on top of pristine
master and run it. If the three positive controls fail there too, the defect is
inherited and belongs in its own change; if they pass, it is this branch's.

Not proposing a fix until that runs — the same failure appears in a separately
written test file on the sibling PR, so a per-PR patch would mask a shared cause.

oferchen:
Root-caused, and it is **not this PR**. The shared failure both PRs hit is
fixed on **#7545**.

`crates/fast_io/src/owner_walk.rs` opened the walk's anchor (`/`) via
`rustix::fs::open`, which on x86_64 lowers to the raw legacy `SYS_open` — a
syscall the daemon's seccomp worker allowlist deliberately omits in favour of
the `*at` variants. strace, in the daemon process:

```
```

so the anchor open is refused and every confined resolution beneath it fails —
surfacing as `mkstemp "payload" (in data) failed: Permission denied (13)`.

Proven by A/B on one binary: `--all-features` fails, `+OC_RSYNC_NO_SECCOMP=1`
passes; and with #7545 applied the same cell passes with seccomp still on.

`git log -S` puts the raw `open` in exactly one commit — #7541, merged earlier
today — so this is a same-day regression on master, not a defect in either PR.
Rebase once #7545 lands.

oferchen:
Diagnosis of the `nextest (stable)` red, measured — not yet root-caused, but two candidate causes are now **eliminated**.

**The failure** is one cell, `an_ordinary_backup_dir_works_under_delay_updates`
(the non-vacuity companion, not an escape pin):

```
```

with the daemon log showing `landlock fully enforced` + `seccomp BPF filter engaged`.
⚠ nextest fail-fast cut the run at 32544/33364, so that is a **lower bound** on the
failure set, not the failure set.

**Eliminated 1 — stale base.** The PR builds `refs/pull/7538/merge` (16b7544a0), and
both recent sandbox fixes are ancestors of it: `243836639` (#7541, O_PATH for
intermediate dirfds) and `c68548259` (#7545, walk anchor opens with `openat`). So this
is not the "rebase and it goes away" shape.

**Eliminated 2 — seccomp.** I rebuilt the exact merge ref and reproduced the cell's
fixture in shell (real daemon, `use chroot = false`, `--backup --backup-dir=bak
--ignore-times [--delay-updates]` over `rsync://`). Four cells, one run each:

| binary | `--delay-updates` | seccomp | result |
|---|---|---|---|
| master `6c9d002b1` | no | off | exit 0 |
| master `6c9d002b1` | yes | off | exit 0 |
| 7538 merge `16b7544a0` | no | off | exit 0 |
| 7538 merge `16b7544a0` | yes | off | exit 0 |
| 7538 merge `16b7544a0` | no | **engaged** | exit 0 |
| 7538 merge `16b7544a0` | yes | **engaged** | exit 0 |

All six pass. `daemon-seccomp` is an optional feature declared only in
`crates/daemon/Cargo.toml` and not re-exported upward, so a plain
`cargo build --bin oc-rsync` gets no seccomp at all — the last two rows needed
`--features daemon/daemon-seccomp` to match what CI's `--all-features` run installs.
With the filter genuinely engaged the cell still passes.

**What is left.** The only remaining difference between my measurement and CI is the
architecture: I measured on **aarch64**, CI's `nextest (stable)` is **x86_64**. A
seccomp allowlist is per-architecture and the syscall sets differ (x86_64 has bare
`open`; aarch64 only `openat`), so `mkstemp`'s underlying call can be admitted on one
and EPERM'd on the other. That is the same family as #7541/#7545 and the sweep that
followed them. I could not test it: the x86_64 host (172.16.1.74) is unreachable from
here right now.

So the next step is to run this one cell — or just the shell fixture above — on x86_64
with `--features daemon/daemon-seccomp`, and if it reproduces, `strace -f` the daemon
worker for the EPERM'd syscall.

oferchen:
Correction to my comment above, before the arch hypothesis gets acted on.

I reasoned from the CI text as if it named the failing syscall:

```
```

It may not. A **backup-mechanism failure can render under that same `mkstemp … failed`
message** via `attempted_temp_path(...).unwrap_or(dest)` — so the string names a temp
create while the refused operation is something else. If that holds here, the operation
actually being refused is the confined backup link/rename, which is what this PR changes,
and the EPERM has nothing to do with seccomp or with the architecture.

That is a better-supported root cause than "x86_64 seccomp allowlist gap", it explains why
the failure is `--delay-updates`-shaped, and — unlike the arch hypothesis — it is settled
by reading the render site rather than by finding an x86_64 runner.

So treat the table in my previous comment as sound (those six cells were really measured)
but its **conclusion as unsafe**: the two eliminations stand, the residual "therefore it is
the architecture" does not, because it was derived from an error string that may be naming
the wrong operation.

Credit to cf-harness, whose task-1049 finding is what caught it.

oferchen:
Precision correction to my own correction — I over-claimed in the same shape I had just flagged.

I wrote that "the operation actually being refused is the confined backup link/rename,
which is what this PR changes." **That is not established**, and I reached it by the same
move that produced the seccomp and arch hypotheses: picking the candidate nearest the
change under suspicion.

What the message text actually settles, verified against master's source:

- `mkstemp_failure` (`crates/transfer/src/pipeline/receiver.rs:304-310`) names
  `attempted_temp_path(error).unwrap_or(dest)`.
- `attempted_temp_path` (`temp_guard.rs:54`) returns `Some` only for a `TempOpenError`
  from `open_tmpfile`; its own doc says *"`None` for every other error, so callers fall
  back to whatever name they already have."*
- Both call sites (`:421`, `:484`) sit in the `Ok(Err(e))` arm of the **whole commit**,
  which spans temp-create → write → backup → rename → metadata.
- CI printed `mkstemp "payload"` — the bare destination, not a `.payload.XXXXXX` temp
  name. The function's own doc gives the intended form as
  `mkstemp ".new.txt.a2dBeL" (in romod) failed`.

So: **the failing operation is not the temp create.** It is one of backup, rename, or
metadata — and the message cannot distinguish them. Naming which one requires evidence
this log line structurally cannot carry.

Two further defects in the same three lines, both on master and neither caused by this PR:

1. **The errno is hardcoded in the format string** — `failed: Permission denied (13)`.
   `error` is consumed only by `attempted_temp_path` and never supplies the errno text, so
   an EPERM(1) would still print `(13)`.
2. **The upstream citation is doing two jobs and only one is right.** `receiver.c:297` is
   cited to justify `FERROR_XFER` (so the peer sets `got_xfer_error` and the run exits 23)
   — that half is correct and load-bearing. But upstream emits that *text* only at the
   real `mkstemp` site; oc hoisted it to cover five operations.

⚠ And the information is already in hand: both call sites do
`meta_errors.push((path, e.to_string()))`, so the true error is captured at the site and
simply not rendered into the user-visible message.

Credit to cf-core, who confirmed the chain and — unlike me — declined to name the backup
on the strength of proximity.

oferchen:
Narrowing measurement: the red is **Linux-only**, and that rules out the backup logic.

Measured on macOS 26.5.2 (aarch64), at this PR's exact head `0329ff026`, with the
binary freshly built first:

```
cargo nextest run -p transfer --all-features -E 'test(an_ordinary_backup_dir)'
  Summary [1.459s] 2 tests run: 2 passed, 2767 skipped
```

Both cells pass here — including `an_ordinary_backup_dir_works_under_delay_updates`,
the one CI fails on all three toolchains. So the assertion at
`daemon_backup_dir_symlink_confinement.rs:467` is not a platform-independent
consequence of this branch's logic.

**Population, stated explicitly:** one host, one arch, one OS. This establishes
*Linux-only*, not *environmental* — a genuine Linux defect and a macOS-passing
build are entirely consistent with each other.

What that buys us: the discriminator is no longer "what does `--delay-updates` do
to the backup path", because that code is identical on both platforms and passes
here. The remaining Linux-specific surface under `Sandbox::Enforced` is the
confined-walk family — the same one as #7541 (O_PATH traversal) and #7545 (openat
anchor), both now merged and both ancestors of this head, so neither is the cause.

⚠ Two things I am NOT claiming, because I have not measured them:
- that the cause is Landlock. It is the largest Linux-only surface in scope, which
  makes it the first thing to A/B (`OC_RSYNC_NO_LANDLOCK`), not the answer.
- that the sibling passing (`..._gets_the_pre_transfer_bytes`, no `--delay-updates`)
  localises the fault to the delay-updates code. It equally fits a staging path
  that only `--delay-updates` reaches.

Next step is one Linux run with the Landlock A/B; the host I use for these was
unreachable at the time of writing, so this is a checkpoint rather than a conclusion.

oferchen:
The run completed, so the log is retrievable. Two measured facts, and the second is
the one that matters.

```
an_ordinary_backup_dir_works_under_delay_updates
  panicked at daemon_backup_dir_symlink_confinement.rs:467:5
  assertion `left == right` failed: the --delay-updates push must succeed
      stderr:
        left: Some(23)
       right: Some(0)
```

**1. Exit 23, not a crash or hang.** The transfer ran and set `got_xfer_error` — oc's
signature for a commit-path failure surfaced as `ErrorXfer`.

**2. ⚠ `stderr:` rendered with nothing after it.** Commit `9089e0189` on this branch
exists specifically to surface the daemon's rsyncd.log on this assertion, and it
surfaced empty. The failing operation produced **no diagnostic text at all**.

So the reason nobody can name what broke is not that the log was unavailable — it is
that **oc does not say**. Before #7552, every commit-path failure is one
undifferentiated `ErrorXfer` carrying neither an operation name nor an errno. That
makes #7552 the *instrument* for this red rather than adjacent work.

**Platform scope, measured on three hosts — stated with its population:**

| host | arch | result |
|---|---|---|
| macOS, this head | aarch64 | both `an_ordinary_backup_dir` cells **pass** (2/2) |
| Linux (192.168.21.226) | **aarch64** | passes, with and without `OC_RSYNC_NO_LANDLOCK=1` |
| CI `nextest (stable)` | **x86_64** (`ci.yml:252`) | **fails** |

⚠ Both greens are aarch64, so neither exonerates CI's Linux — they isolate OS while
holding arch constant, which is a cleaner pair than it looks, but it is not a
clearance. Landlock is A/B'd and not implicated on aarch64.

⚠ Method note worth recording: the first Linux run used `-p transfer --all-features`,
where CI uses `--workspace --all-features` (`ci.yml:299`). Those produce different
feature unification for `transfer` — the exact discriminator measured on this codebase
in #7545 (`--workspace` passes / `--all-features` fails). Caught before reporting; a
`--workspace` re-run is in flight.

**Open lead, not settled.** oc's delay-updates staging branch is gated
`config.delay_updates` and is mutually exclusive with the backup branch's
`!config.delay_updates` — so under `--delay-updates` this test does not take the
backup path at all. Which makes it genuinely odd that a backup-dir confinement change
reddens it. Two shapes to separate before anyone edits `transfer.rs`:

- (a) this branch's union resolution widened a refusal the *staging* rename now hits;
- (b) the staging path already returned `ErrorXfer` on this fixture and this branch
  merely added the first test that looks — which would make the red **pre-existing and
  mis-attributed**.

Not editing the branch until that is settled.

PR #7545, opened 2026-08-27
fix(fast_io): open the walk anchor with openat so the worker filter admits it

oferchen:
## What breaks

A plain in-module daemon push with `--backup-dir` fails on x86_64 Linux under
`--all-features`:

```
```

The transfer dies when the receiver creates its destination temp file, before
any backup or symlink logic runs.

## Root cause

`open_start_dir` opened the walk's starting directory — `/` for an absolute
path — through `rustix::fs::open`. On x86_64 rustix lowers the no-dirfd form to
the **raw legacy `SYS_open`**, and the daemon's seccomp worker allowlist admits
only the `*at` variants. strace, in the daemon process:

```
```

The anchor open is refused, so every confined resolution beneath it fails.

Measured A/B — same binary, same flags, one environment variable:

| run | config | result |
|---|---|---|
| A | `--all-features` | FAIL, `mkstemp … Permission denied (13)` |
| B | `--all-features` + `OC_RSYNC_NO_SECCOMP=1` | **1 passed** |

With the fix applied and seccomp left **on**, the same cell passes.

## Why the fix is at the call site

The allowlist is right and the callers were wrong. `openat(AT_FDCWD, p, ..)` is
the identical operation and is already permitted, so this uses the `at` form
rather than widening the filter.

**Two sites, not one.** `open_start_dir` is the one that fires today; the
opt-out arm in `operator_open_with_owner_walk` carried the same raw `open` and
is latent only because the confinement session is not yet installed. Both are
fixed here — leaving the second would resurface this the moment that gets
wired.

`aarch64` has no `SYS_open` at all, which is why only x86_64 could observe it,
and why the three `nextest` cells were the only ones affected.

## Provenance

`git log -S 'rustix::fs::open(if absolute'` returns exactly one commit:
`243836639` (#7541, merged earlier today), which rewrote this call into the raw
form. Same-day regression.

## Test

The pin installs the **production** worker filter through
`apply_worker_seccomp_filter` — not a hand-copied rule set — and then runs the
real ownership walk, so it tracks the live allowlist as that changes. No
existing test called `apply_worker_seccomp_filter` at all, which is the
coverage gap that let this land.

## Unblocks

#7535 and #7538 both fail on this shared cause, from two independently written
fixture files with a byte-identical error. Neither should be patched in
isolation; both should go green once this lands.

### DIFF
--- a/crates/daemon/tests/seccomp_worker_filter.rs
+++ b/crates/daemon/tests/seccomp_worker_filter.rs
@@ -20,6 +20,8 @@ use std::collections::BTreeMap;
 use std::io;
 use std::os::unix::process::ExitStatusExt;
 
+use daemon::seccomp_test_support::{SeccompOutcome, apply_worker_seccomp_filter};
+
 /// Architecture detected at build time.
 fn target_arch() -> Option<TargetArch> {
     if cfg!(target_arch = "x86_64") {
@@ -169,5 +171,57 @@ fn blocked_syscall_traps_with_sigsys() {
     assert_ne!(
         code, 99,
         "child reached past blocked syscall - filter not enforcing",
+    );
+}
+
+/// The ownership walk must keep resolving once the worker filter engages.
+///
+/// The walk opens its starting directory - `/` for an absolute path - and
+/// that open has to be an `openat`. On x86_64 rustix lowers the no-dirfd
+/// `rustix::fs::open` to the legacy `SYS_open`, which this allowlist
+/// deliberately omits in favour of the `*at` variants, so a plain `open`
+/// returns `EPERM` inside a worker and every confined resolution beneath
+/// it fails. Regression pin: a daemon receiver could not create its
+/// destination temp file, surfacing as `mkstemp ... Permission denied`.
+///
+/// Installs the production filter through `apply_worker_seccomp_filter`
+/// rather than a hand-copied rule set, so the pin tracks the real
+/// allowlist as it changes.
+#[test]
+fn the_ownership_walk_resolves_under_the_worker_filter() {
+    let probe = std::env::temp_dir().join(format!("oc-owner-walk-probe-{}", std::process::id()));
+    std::fs::write(&probe, b"probe").expect("probe file must be creatable");
+    assert!(
+        probe.is_absolute(),
+        "the probe must be absolute so the walk starts at `/`",
+    );
+
+    let child_path = probe.clone();
+    let raw = fork_run(move || {
+        match apply_worker_seccomp_filter() {
+            SeccompOutcome::Installed => {}
+            // Kernel refused the filter, or this build cannot install one:
+            // treat as a skip, matching the sibling tests above.
+            _ => return 77,
+        }
+        match fast_io::owner_walk::operator_open_read(&child_path) {
+            Ok(_) => 0,
+            Err(_) => 99,
+        }
+    });
+
+    let _ = std::fs::remove_file(&probe);
+    let status = std::process::ExitStatus::from_raw(raw);
+    if let Some(sig) = status.signal() {
+        panic!("child killed by signal {sig} while walking under the worker filter");
+    }
+    let code = status.code().expect("child must exit");
+    if code == 77 {
+        eprintln!("seccomp filter install rejected by kernel; skipping");
+        return;
+    }
+    assert_eq!(
+        code, 0,
+        "the ownership walk was refused under the worker filter - a syscall the walk issues is missing from the allowlist",
     );
 }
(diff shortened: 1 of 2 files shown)


oferchen:
## Upstream fidelity of this change (rsync 3.5.0 `syscall.c`)

Checked before merging, because the change edits code that mirrors upstream's
ownership walk. Three findings, two of them divergences worth stating plainly.

### 1. The change is a spelling, not a behaviour change - and upstream already prefers the `at` form

Upstream's per-component descend is the `at` form:

```c
int next = openat(dfd, comp, O_RDONLY | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);   /* syscall.c:493 */
```

and its relative anchor is not a syscall at all - `int dfd = AT_FDCWD;` (`syscall.c:305`),
used directly as the dirfd for the first `fstatat`/`openat`.

`openat(AT_FDCWD, p, flags, mode)` is defined to be the identical kernel operation to
`open(p, flags, mode)`. So this commit changes which syscall number is issued and nothing
else - no flag, no resolution semantics, no ordering.

### 2. DIVERGENCE (deliberate, oc-only): upstream opens its ABSOLUTE anchor with a bare `open`

```c
/* Absolute path: pin "/" as the starting dfd. */
if (remaining[0] == '/') {
        dfd = open("/", O_RDONLY | O_DIRECTORY | O_CLOEXEC);   /* syscall.c:349-351 */
```

That is exactly what this branch replaces. Mirroring upstream literally is not available
here, and the reason is that **upstream ships no seccomp filter at all** - there is no
`seccomp` anywhere in the 3.5.0 tree. The daemon worker allowlist is an oc-only hardening
extension, so this interaction has no upstream counterpart to mirror.

The allowlist admits the `*at` family (`openat`, `openat2`, `mkdirat`, `linkat`,
`renameat`, `statx`) and deliberately omits the legacy non-`at` spellings. On x86_64 rustix
lowers the no-dirfd form to the raw legacy `SYS_open`, so upstream's spelling is `EPERM`'d
inside a worker. Correcting the call site rather than widening the filter keeps the
allowlist as the tighter of the two.

### 3. Pre-existing divergence, unchanged here, recorded so it is not mistaken for this commit

- `traversal_dir_flags()` uses `O_PATH` on Linux where upstream uses `O_RDONLY`
  (`syscall.c:493`). Deliberate: a Landlock-restricted process needs `FS_READ_DIR` for an
  `O_RDONLY` dirfd but no right for `O_PATH`. Already documented at the definition.
- oc opens `"."` to obtain a real fd for the relative anchor where upstream uses the
  `AT_FDCWD` sentinel and issues no syscall. Semantically equivalent; costs one syscall and
  one descriptor per walk. Not touched by this change - filed rather than folded in.

oferchen:
## Correction to my earlier note on this PR

I wrote above that the seccomp interaction "has no upstream counterpart to mirror."
That is **too strong, and the distinction matters.**

Upstream installs no seccomp filter - that part holds, and I re-verified it (zero
`seccomp_load`, `SCMP_`, or `prctl(...SECCOMP...)` sites anywhere in the 3.5.0
tree). But upstream is **not silent about externally-imposed filters**, and it has
a written policy for exactly the situation this PR is about.

`generator.c:1444-1487`:

```c
/* True when this build compiled no fd-relative primitive able to create this
 * kind of node.  That is a property of the build, not of the call, so it is
 * decided here rather than inferred from an errno. */
static int no_atfd_mknod_primitive(mode_t mode)
```

and at its call site:

```c
/* Fall through to the unconfined path-based create only where this build
 * compiled no fd-relative primitive for this kind of node -- SECURITY.md's
 * rule for a platform that cannot be secure at all.  Testing errno == ENOSYS
 * is not that test: a live mknodat() or mkfifoat() can return ENOSYS too
 * (an unimplemented FUSE mknod, or seccomp), which would drop confinement on
 * a platform that does have the secure primitive. */
```

`:1455` names the shape directly - "a seccomp policy permitting `open()` but not
`openat()`".

**This strengthens rather than undermines the change merged here.** Upstream's
rule is: prefer the `*at` primitive, and never degrade to the unconfined form
because of a runtime denial. Moving the walk anchor onto `openat` is the same
direction. The one claim to strike is "no counterpart" - the filter has none, the
policy does.

**Residual, filed separately rather than folded in:** oc has two helpers that key
a drop-to-unconfined fallback on a runtime `Err` of any errno -
`open_delete_sandbox` (`crates/engine/src/delete/context/core.rs:684`) and
`open_sandbox_for_dest` (`crates/transfer/src/receiver/transfer/setup/sandbox.rs:47`) -
both logging at debug level 2 only. Whether a live sandbox denial can actually
reach either arm is unmeasured, so that is a measure-first item, not a fix to
write on the strength of a source reading.

PR #7588, opened 2026-09-01
test(daemon): run the module filter-directive tests un-detached

oferchen:
## The defect

All three tests in `tests/integration_daemon_filter_directives.rs` started the daemon without `--no-detach`. `RuntimeOptions::detach` defaults to `cfg!(unix)`, so `run_daemon` reached `become_daemon()` in the accept loop and the parent of that fork - the test binary itself - exited 0 before any assertion ran.

Proven by mutation on the base commit. With an unconditional `panic!` inserted immediately after the daemon spawn:

```
Starting 3 tests across 1 binary (88 binaries skipped)
     Summary [   0.025s] 3 tests run: 3 passed, 0 skipped
```

The tell was the runtime: 0.029s for three full TCP transfers through an in-process daemon.

## What making them execute surfaced

The tests checked only that `run_client` returned `Ok`. A daemon-refused push does not return `Err`:

- `generator.c:1669-1671` reports the refusal as `FERROR_XFER`
- `log.c:337-338` sets `got_xfer_error` on receipt
- `cleanup.c:217-218` lifts the exit status to `RERR_PARTIAL` (23, `errcode.h:43`)

That status arrives on `ClientSummary::io_error_exit_code`, not as an error, so the push's actual outcome was never asserted. It is 23 in all three scenarios, and nothing noticed.

This matters because the assertions that did exist are one-sided. `!files.contains("drop1.log")` is satisfied just as well by a transfer that achieved nothing at all. Each test now asserts exit 23 as the positive signal that the module directive actually fired.

Two smaller corrections in the same pass:

- `collect_files` carries file contents alongside the names, so a kept file has to arrive intact rather than merely exist.
- The module header's upstream citations were on 3.4.4 line numbers. Moved onto the 3.5.0 pin: `clientserver.c:934-951`, `receiver.c:889`, `generator.c:1663`.

## Mutation proofs

Each mutation reddens a distinct assertion, so neither is carrying the other.

**Silencing the `FERROR_XFER` frame** in `crates/transfer/src/receiver/transfer/candidates.rs` - files are still excluded, so the on-disk assertions stay green and only the new assertion fires:

```
thread 'daemon_filter_directive_excludes_match_pattern' panicked at tests/integration_daemon_filter_directives.rs:120:5:
assertion `left == right` failed: a push whose files the module refused must exit 23 (cleanup.c:217-218), not report success
  left: None
 right: Some(23)
 Summary [   0.107s] 3 tests run: 0 passed, 3 failed, 0 skipped
```

Without the added exit-code assertion this regression would have been invisible.

**Disabling the daemon filter check** - the excluded files land and the file's own claim fires:

```
thread 'daemon_filter_directive_excludes_match_pattern' panicked at tests/integration_daemon_filter_directives.rs:251:5:
drop1.log must be excluded by `filter = - *.log`, got ["drop1.log", "drop2.log", "keep1.txt", "keep2.txt"]
 Summary [   0.122s] 3 tests run: 0 passed, 3 failed, 0 skipped
```

Unmutated: `3 tests run: 3 passed, 0 skipped`.

## Guard

`tools/ci/check_daemon_no_detach.sh` covered only `crates/daemon/src/tests`. Root-level integration tests live outside the daemon crate and cannot reach its test-support helpers, so checks 1-3 were blind to them while the fork killed the test binary there just the same. A fourth check counts root tests calling `run_daemon(` without `--no-detach`, with a shrink-only ceiling at the 2 files still to be migrated.

Verified the guard fires rather than passing vacuously - temporarily lowering the ceiling to 1:

```
Root tests running a detaching daemon: 2 (ceiling 1)
FAIL: 2 root integration test(s) call run_daemon() without --no-detach, ceiling is 1:
  tests/integration_daemon_max_connections_cap.rs
  tests/integration_daemon_server.rs
```

Those two are being migrated separately. This check overlaps the one in #7578, which fixes `integration_daemon_max_connections_cap.rs`; whichever lands second should lower the ceiling again.

## Gates

- `cargo clippy --workspace --all-targets --all-features --no-deps -- -D warnings` clean
- `bash tools/ci/check_daemon_no_detach.sh` - no violations
- `cargo xtask citations` - exit 0, all citations resolve against the rsync 3.5.0 pin

### DIFF
--- a/tests/integration_daemon_filter_directives.rs
+++ b/tests/integration_daemon_filter_directives.rs
@@ -8,11 +8,19 @@
 //! directive never land on disk - regardless of whether the client supplied
 //! any filter on its own.
 //!
-//! upstream: clientserver.c:874-893 - `rsync_module()` builds
+//! upstream: clientserver.c:934-951 - `rsync_module()` builds
 //! `daemon_filter_list` from `filter` / `include_from` / `include` /
 //! `exclude_from` / `exclude` in that order, then `check_filter()` at
-//! `receiver.c:711-714` and `generator.c:1273-1275` consults it before any
+//! `receiver.c:889` and `generator.c:1663` consults it before any
 //! per-file action.
+//!
+//! A refused file is never dropped in silence: `generator.c:1669-1671` reports
+//! it as `FERROR_XFER`, `log.c:337-338` sets `got_xfer_error` on receipt, and
+//! `cleanup.c:217-218` lifts the push's exit status to `RERR_PARTIAL` (23,
+//! `errcode.h:43`). Every test below asserts that status as well as the on-disk
+//! end state, because the status is the signal that the module directive
+//! actually fired: an absent `.log` file is equally consistent with a transfer
+//! that achieved nothing at all.
 //!
 //! Gated `#[cfg(unix)]` because the in-process client + daemon split uses
 //! POSIX TCP and the helper walks the destination with `read_dir`; the test
@@ -22,7 +30,7 @@
 
 #![cfg(unix)]
 
-use std::collections::HashSet;
+use std::collections::BTreeMap;
 use std::ffi::OsString;
 use std::fs;
 use std::net::{Ipv4Addr, TcpListener};
@@ -40,6 +48,11 @@ use tempfile::tempdir;
 /// concurrently on a constrained CI runner.
 static TEST_LOCK: Mutex<()> = Mutex::new(());
 
+/// Upstream's exit status for a transfer that lost files to an error.
+///
+/// upstream: errcode.h:43 - `#define RERR_PARTIAL    23`.
+const RERR_PARTIAL: i32 = 23;
+
 /// Allocates a free TCP port for the test daemon. Returns both the port and
 /// the bound `TcpListener` so the listener can be handed to the daemon via
 /// `pre_bound_listener` - this closes the TOCTOU window between port
@@ -50,10 +63,13 @@ fn allocate_test_port() -> Option<(u16, TcpListener)> {
     Some((port, listener))
 }
 
-/// Walks `root` recursively and returns the set of relative file paths. Used
-/// to assert the destination contains exactly the files we expect.
-fn collect_files(root: &Path) -> HashSet<String> {
-    let mut out = HashSet::new();
+/// Walks `root` recursively and maps each relative file path to its contents.
+///
+/// The contents travel with the names so that a kept file has to arrive intact
+/// to satisfy the positive assertions. Name-only presence would also be
+/// satisfied by an empty placeholder left behind by a half-finished transfer.
+fn collect_files(root: &Path) -> BTreeMap<String, Vec<u8>> {
+    let mut out = BTreeMap::new();
     let mut stack = vec![root.to_path_buf()];
     while let Some(dir) = stack.pop() {
         let Ok(entries) = fs::read_dir(&dir) else {
@@ -65,8 +81,8 @@ fn collect_files(root: &Path) -> HashSet<String> {
             if meta.is_dir() {
                 stack.push(path);
             } else if meta.is_file() {
-                if let Ok(rel) = path.strip_prefix(root) {
-                    out.insert(rel.to_string_lossy().into_owned());
+                if let (Ok(rel), Ok(contents)) = (path.strip_prefix(root), fs::read(&path)) {
+                    out.insert(rel.to_string_lossy().into_owned(), contents);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


oferchen:
**CI red triage — evidence, not yet a root cause. The failing test is in a file this PR does not touch.**

Two Linux cells failed identically on `f0a4d1e`-era head:

- `Feature flag combinations / default-features (linux)`
- `Feature flag combinations / no-incremental-flist (linux)`

Both on the same single test, with the same message:

```
test session_error_does_not_stop_the_accept_loop ... FAILED
thread '...' panicked at tests/daemon_session_error_keeps_listener.rs:169:5:
```

**Three things worth separating before anyone reaches for a fix.**

**1. The assertion text names a handshake desync, not a plain mismatch.** The module name the server reports is `@RSYNCD: 32.0 md5 md4` — the daemon greeting line itself, consumed as if it were the module request. So the server was one line out of step in the `@RSYNCD` exchange when that client arrived. It did not merely answer where the test expected a drop; it answered from the wrong handshake state.

**2. It is NOT deterministic on this commit, and that is the load-bearing observation.** Five sibling Linux cells ran the same test on the same commit and passed: `copy_file_range`, `landlock`, `openssl-vendored`, `zlib-ng`, `zlib-rs`. Two failed. A feature-flag-dependent logic error would not split that way — the feature sets that failed (default, and default-minus-incremental-flist) have no obvious relationship to daemon handshake code. Same commit + same test + both outcomes = a race, so any fix that assumes a deterministic logic defect is likely to be aimed at the wrong thing.

**3. `daemon_session_error_keeps_listener.rs` is not in this PR's diff.** The change set is `tests/integration_daemon_filter_directives.rs`, `tests/integration_daemon_server.rs`, and `tools/ci/check_daemon_no_detach.sh`. So if this PR is responsible, it is by *interference*, not by editing the failing test — and the plausible mechanism is exactly what the PR does: running daemons **un-detached** changes daemon process/port lifetime, and these are all root-level integration binaries running concurrently under nextest. A lingering or differently-scheduled daemon answering another test's client would produce precisely a wrong-handshake-state reply.

That is a hypothesis, not a conclusion. **The measurement that would settle it** is whether `session_error_does_not_stop_the_accept_loop` fails on master at the same commit base — I could not run that here because master's own runs were still in flight at the time of writing. If it reproduces on master, this PR is exonerated and the test is flaky independently (which would make it a pre-existing defect worth its own fix, not a blocker for this PR).

⚠ Two caveats on the numbers above:

- `1316/33035 tests run` — nextest fail-fast truncated the run, so **one failing test is a lower bound, not the failure set**. There may be more behind it.
- I have not reproduced this locally. Everything above is read from the two job logs.

oferchen:
**This PR is EXONERATED. The red is a master defect, and my earlier triage on this PR was wrong on both counts — retracting it.**

I previously wrote that the failure looked like a handshake desync caused by cross-test interference from un-detached daemons, and that the 5-pass/2-fail split across sibling cells made it a race. Both claims are refuted below. The mechanism is deterministic and has nothing to do with this PR.

## What actually happens

`run_provoker` writes two banner lines in a single `write_all`:

```
@RSYNCD: 32.0 md5 md4\n@RSYNCD: 32.0 md5 md4\n
```

Line 1 is the version echo. Line 2 is *intended* to re-drive `Greeting -> ModuleSelect` from `ModuleSelect`, which the FSM rejects (`valid_transitions`: `ModuleSelect => [Authenticating, Transferring]`), producing the session error the test needs.

Master no longer does that. `session_runtime.rs` deliberately treats the second banner as a module name:

```rust
// upstream: clientserver.c:1534-1538 start_daemon() - the version line is read
// exactly once, by exchange_protocols(); the very next read_line_old() is the
// request line whatever it contains. A second `@RSYNCD:` banner is therefore a
// module name, not a re-greeting, and falls through to the unknown-module
// refusal. Taking the Greeting -> ModuleSelect edge again instead would fail the
// FSM's ordering check, and a session error is fatal to the whole listener
// (workers.rs join_worker), so a peer could end the daemon with two greeting lines.
Ok(LegacyDaemonMessage::Version(_)) => {}
```

So the daemon answers `@ERROR: Unknown module '@RSYNCD: 32.0 md5 md4'` — **exactly the string the failing assertion reports**. No desync, no interference, no race.

## Why it landed green and only surfaces now

| commit | PR | merged |
|---|---|---|
| `34b49e9cd` the fallthrough arm | #7579 | 2026-09-01 **14:09** |
| `7e2d9391a` the test | #7590 | 2026-09-01 **16:39** |

The arm merged **2h30m before** the test. But #7590's tested head (`53d155c07`) does **not** contain `34b49e9cd` — verified with `git merge-base --is-ancestor`. Up-to-date-branch enforcement is off on this repo, so #7590 was validated against a master that still rejected the repeated banner, merged green, and landed onto a master that had already changed that behaviour. Two individually-correct commits, merged out of order, never tested together.

## It is not this PR, and not only this PR

- The failing test file is in none of this PR's three changed files.
- Every open PR whose head carries **both** commits is affected. Measured across the open set: #7614, #7613, #7612, #7611, #7607, #7598 and this one all carry both. This PR is simply the first whose feature-flag cells have finished — #7611 and #7598 have those two cells still in flight.
- No PR anywhere shows SUCCESS on those cells while carrying both commits, which is the check that would have refuted this.

The "5 siblings passed" observation I leaned on earlier does not establish a race: those cells are a different feature set and the split is not evidence of nondeterminism once the mechanism is deterministic and read from source.

## The test's own guard worked

This is not a silent breakage. The assertion that fires is the deliberate non-vacuity check, whose comment predicted precisely this:

> *"If the daemon ever starts answering a repeated banner cleanly this fixture stops provoking anything, and the property above would pass while proving nothing - so fail loudly here instead."*

Without it the transfer assertions would still pass and the test would have gone quietly inert.

## What the fix is not

Restoring the FSM rejection would revert #7579 deliberately and re-open what its comment names: a peer ending the whole daemon with two greeting lines. The arm is upstream-faithful and security-motivated; it should stay.

The fix belongs on the test — it needs a provoker that still produces a genuine session error under the current daemon, or an explicit decision that peer-triggerable session errors no longer exist (which is arguably what #7579 achieved, and would make this fixture unsatisfiable by construction rather than merely stale). That is a call for the owners of #7590 / #7579, and it is a master fix, not a change to this PR.

⚠ Caveat kept from before: nextest fail-fast truncated the run, so one failing test is a lower bound, not the failure set.

oferchen:
**Correction to my own blast-radius claim: it is every open PR, not the seven I listed.**

I earlier classified affected PRs by whether their **head** carries both `34b49e9cd` (the arm) and `7e2d9391a` (the test), and named seven. That predicate is wrong, and #7594 refutes it directly:

```
7594 head  d597376646e0b62ec41df13ae3e4e75e0607afa7  -> does NOT carry the test
7594 CI    Linux musl (stable) FAILS session_error_does_not_stop_the_accept_loop:169
```

A head that does not contain the test cannot fail it — unless the tested tree is not the head. It is not:

```
git fetch origin refs/pull/7594/merge
git merge-base --is-ancestor 7e2d9391a FETCH_HEAD   ->  MERGE ref CARRIES the test
```

`pull_request` CI builds **merge(head, base)**, so every open PR compiles master's current content, arm and test included, whatever its own branch holds. The correct statement is therefore: **every open PR hits this once its CI runs**, and rebasing is not what exposes it.

**The cell set is also wider than the two I named.** `Linux musl (stable)` fails the same test on both #7594 and #7588 — same test, same line, same assertion. Any cell that runs the root integration tests reaches it, not just the two `Feature flag combinations` cells I happened to see first.

Both corrections make the case for the fix stronger, not weaker: the mechanism, the root cause and the fix in #7615 are unchanged — only my estimate of who is affected was too narrow.

PR #7597, opened 2026-09-01
fix(receiver): rmdir a directory obstacle instead of backing it up

oferchen:
## What

A directory standing where the receiver has to create a symlink, FIFO, socket, or device node was neither removed nor reported. The five `backup_existing_before_replace` call sites in `links.rs` / `special.rs` each did:

```rust
Ok(false) => { let _ = fast_io::unlink_via_sandbox_or_fallback(.., UnlinkFlags::File); }
Err(e)    => { debug_log!(Recv, 1, "failed to back up existing obstacle …"); continue; }
```

`File`-only, so it structurally cannot `rmdir` a directory by any route; the result discarded; the failure on a verbosity gate with no `MessageCode` and no effect on the exit code.

## Why the removal and the reporting are ONE change

Upstream decides this in one place, and the decision is ordered:

```c
/* generator.c:2469 */
int skip_atomic, dir_in_the_way = del_for_flag && S_ISDIR(sxp->st.st_mode);
if (!del_for_flag || dir_in_the_way || tmpdir || !get_tmpname(tmpname, fname, True))
        skip_atomic = 1;
...
if (make_backups > 0 && !dir_in_the_way) {          /* :2477 */
        if (!make_backup(fname, skip_atomic)) return 0;
} else if (skip_atomic) {                           /* :2480 */
        int del_opts = delete_mode || force_delete ? DEL_RECURSE : 0;
        if (delete_item(fname, sxp->st.st_mode, del_opts | del_for_flag) != 0) return 0;
}
```

`dir_in_the_way` forces `skip_atomic`, so **a directory obstacle reaches `delete_item` whether or not `--backup` is set** — the `make_backup` arm is unreachable for it in either mode. `delete_item` then splits on the same test one level down: the `rmdir` arm at `delete.c:222-226`, the `make_backup`/`unlink` arm in the `else` at `delete.c:227-238`.

Two consequences, and they are what make this a single commit:

1. **A guard on the backup path cannot work.** It could only cover the `--backup`-on half; the four `--backup`-off shapes in the table below would still be decided in `links.rs`/`special.rs`. That is a divergence from upstream's structure by construction, not by taste.
2. **The reporting alone is a regression.** Measured by mutation, with the removal disabled and the diagnostics in place, a symlink over an **empty** directory gives:

   ```
   rsync: [receiver] delete_file: unlink(obstacle) failed: Operation not permitted (1)
   could not make way for new symlink: obstacle
   left: Some(23)   right: Some(0)
   ```

   — exit 23 on a shape 3.5.0 completes at exit 0.

So the five call sites now share one `make_way_for_replacement`, which owns both arms and emits upstream's diagnostics itself. `backup_existing_before_replace` is now called from exactly one place.

## Measured

rsync 3.5.0 (protocol 32) vs oc, over a real `--server` child on both ends (`-e` wrapper, `--rsync-path` pinning both sides). Not the local-copy executor: that path refuses a directory obstacle in its own error type (`local_copy/error.rs:427`) and exercises none of this.

`--backup` off, `-rlDI`, source `obstacle` replacing `dst/obstacle/`:

| src kind | obstacle | rsync 3.5.0 | oc before | oc after |
|---|---|---|---|---|
| symlink | empty dir | **0**, symlink placed | **12** `server error: File exists (17)`, dir kept | **0**, symlink placed |
| symlink | non-empty dir | **23** + two diagnostics | **12**, dir kept | **23** + same two diagnostics |
| fifo | empty dir | **0**, fifo placed | **0 SILENT**, dir kept | **0**, fifo placed |
| fifo | non-empty dir | **23** + two diagnostics | **0 SILENT**, dir kept | **23** + same two diagnostics |

Three things in that table are worse than the exit codes suggest:

- **Both FIFO rows reported success while losing the node.** Silent data loss, not a loud failure.
- **The failing rows abandoned the rest of the batch.** The sibling file upstream transfers at exit 23 was never written.
- **Under `--backup`, oc RENAMED the directory into the backup area** (`--backup-dir=bak` → `bak/obstacle`), which upstream never does. A wrong-data outcome at exit 0, which no exit-code assertion could have caught. Now removed.

The two `regular file` rows of the same matrix (12 vs 0/23) are the `DEL_FOR_FILE` site at `generator.c:2148` and are deliberately untouched here — see "Relationship to the wider obstacle-removal work" below.

Also verified by strace of the server child, on the exact shape the failing testsuite cell hits (`--backup-dir` equal to the obstacle):

```
linkat(AT_FDCWD, ".../dst/a", AT_FDCWD, ".../dst/a/a", 0) = -1 EPERM
renameat(AT_FDCWD, ".../dst/a", AT_FDCWD, ".../dst/a/a")  = -1 EINVAL
```

…and then nothing. No `rmdir`, no `symlinkat`, no output at `-vvv --debug=RECV5,BACKUP5`.

## Diagnostics

Upstream's text and routing, per arm:

- populated directory → `cannot delete non-empty directory: %s` at `FINFO` (`delete.c:178-181`), then `could not make way for new %s: %s` at `FERROR_XFER` (`delete.c:283-285`). The `FERROR_XFER` sets the peer's `got_xfer_error` (`log.c:310-311`), which lifts a zero exit to `RERR_PARTIAL` (`cleanup.c:217-218`).
- other `rmdir`/`unlink` errno → `delete_file: rmdir(...) failed` / `unlink(...) failed` at `FERROR_XFER` (`delete.c:264-266`), then the same refusal line.
- `ENOENT` → not a failure; upstream maps it to `DR_SUCCESS` (`delete.c:268-269`) and creates the replacement.
- backup-arm failure → `keep_backup failed: <path>` at **`FERROR`**, not `FERROR_XFER`. Upstream leaves `got_xfer_error` clear on that arm (`backup.c:400-401`), so the exit code is unchanged there.

The noun comes from the **new** entry's type, not the obstacle's (`generator.c:2041-2047` picks `DEL_FOR_DEVICE`/`DEL_FOR_SPECIAL`; `delete.c:275-282` turns it into the word), so a FIFO says `special file` where a symlink says `symlink`.

The obstacle type is read from the same sandbox-anchored `fstatat` the removal is anchored on (new `LstatOutcome::is_dir`), so a TOCTOU swap between classifying and removing cannot make the two disagree.

## Test

`tests/directory_obstacle_removal.rs`, six cells, all through a real `--server` child.

**Mutation-proven, each half reverted in turn:**

| mutation | red | green |
|---|---|---|
| reporting restored to the swallow (removal intact) | the 3 refusal cells (23 → 0, no diagnostics) | the 3 removal cells |
| removal arm deleted (reporting intact) | 5 of 6, incl. `an_empty_directory_obstacle_is_removed_for_a_symlink` at **23 where upstream is 0** | 1 |

The FIFO-over-empty-directory cell is the one that separates the halves: it is **green** under the first mutation and **red** under the second, so it distinguishes "the removal works" from "the removal now fails loudly instead of silently". Nothing else in the set does.

The non-vacuity companion is `a_directory_obstacle_is_removed_not_backed_up_under_backup`: a shape upstream completes at exit 0 must still exit 0 after the change, and must not leave the directory in the backup area.

**Regression surface.** One existing unit test had to move: `receiver::tests::munge_symlinks::create_symlinks_surfaces_non_eacces_error` planted an **empty directory** at the leaf to make `symlinkat` return `EEXIST`, and closed with `assert!(dest.join("blocked").is_dir(), "the planted directory must persist")` — an assertion on the defect. Its invariant (non-EACCES → `Err`) is real and kept; the fixture now plants a regular file where a path **component** belongs, giving `ENOTDIR`. `delete_item` only ever clears the leaf, so no obstacle removal can make way for that one.

## Verification

- Full 3.5.0 python suite, 345 cells, nonroot, aarch64 Linux: 253 pass / 6 fail / 86 skip. The six failures are **exactly** the six the expect manifest already names (`basis-xname-traversal`, `batch-only-remove-source-regression`, `filter-merge-content-echo`, `partial-protected-regular-retry-linux`, `rrsync-backup-dir-inband-pivot`, `rrsync-merge-file-confine`). No row moves in either direction, so `tools/ci/upstream-3.5.0-expect.{nonroot,root}.txt` are unchanged.
- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --all-features --no-deps -- -D warnings` — no finding in any changed file.
- `cargo nextest run --workspace --all-features` — 32881 tests, 32880 passed. The one failure is `xtask … backdated_tree_populates_and_backdates_the_root`, a pre-existing macOS-host failure (`touch -d @1614830767`: BSD `touch` rejects the `@epoch` form), unrelated to this change.
- `tools/ci/citation_drift_audit.py` — exit 0.

## `rrsync-backup-dir-inband-pivot`: 2 of 3 assertions flipped, and the third is named

This is the cell that motivated the change. It has three assertions past the escape check. Before: all three failed at the first one. After:

| assertion | before | after |
|---|---|---|
| `got.returncode != 0` | rc=0 ❌ | **rc=12 ✅** |
| `pivot == '../outside'` | `None` ❌ | **`'../outside'` ✅** |
| `'/proc/self/fd/' in got.stderr` | ❌ | ❌ still |

The mechanism now works end to end: the `rmdir` frees the name (so the transfer's own symlink is planted, `pivot='../outside'`) **and** destroys the inode `rrsync` holds on fd 5, so the later `authorized_keys` backup opens a deleted directory and fails. 3.5.0 does exactly this and exits 11 at `rsync.c:894-896`.

The remaining assertion is **not** obstacle-removal territory, and there are two independent reasons for it:

1. The failure is `ENOENT`, and `pipeline/receiver.rs:445/508` reaches `commit_failure` only on the `is_permission_error` branch; `ENOENT` takes the fatal `return Err(e)` instead, so the upstream-shaped line is never rendered.
2. Even on that branch the operand would be wrong: `disk_commit/process/commit.rs:110-116` passes `attach_commit_op(CommitOp::Backup, &begin.file_path, e)` — the **pre-image**, not the backup destination. The comment at `pipeline/receiver.rs:316-320` already documents this ("oc names the pre-image only; the backup destination is chosen inside make_backup and is not carried back on the error"), so `/proc/self/fd/5/authorized_keys` would still not appear.

Both belong to #7582 ("abort with 11 when a commit-path backup fails"), which routes the fatal backup failure through `commit_failure` and flushes the queued line. The manifest row is left at `fail` rather than re-baselined.

## Relationship to the wider obstacle-removal work

The general missing removal is upstream's `delete_item(… DEL_FOR_FILE)` at `generator.c:2148`, which covers a **regular file** landing on an obstacle. That is a different call site and is not touched here — the two `regular file` rows of the matrix above are unchanged (12 where upstream is 0 or 23). This change builds the shared decision function with both of `delete_item`'s arms, so routing that site onto it later is a call-site change rather than a reconciliation of two overlapping mechanisms. #7583's `write_devices && stype == FT_DEVICE` carve-out is the other half of that line and is orthogonal to both.

## Known residual, measured and deliberately out of scope

`generator.c:2481` computes `int del_opts = delete_mode || force_delete ? DEL_RECURSE : 0`, so under `--delete` or `--force` upstream removes a **populated** directory obstacle recursively and completes at exit 0. Measured against 3.5.0: `--force`, `--delete`, and both together each give `rc=0` with the symlink in place, where this arm refuses at 23.

Neither flag is reachable from the receiver. `--force` is parsed and **discarded** by the server arg parser (`cli/src/frontend/server/flags.rs:615`, `"--force" => {}`) and has no field in `ParsedServerFlags`. Wiring it is a server-arg and config change, not an obstacle-removal one. Before this change the same shape aborted the whole transfer at 12 with the sibling files unsent, so the refusal is strictly closer to upstream, not a new divergence. Recorded in the module doc so it is not rediscovered as a surprise.

### DIFF
--- a/crates/fast_io/src/dir_sandbox/at_syscalls/lstat.rs
+++ b/crates/fast_io/src/dir_sandbox/at_syscalls/lstat.rs
@@ -43,6 +43,20 @@ impl LstatOutcome {
         }
     }
 
+    /// Reports whether the entry is a directory.
+    ///
+    /// The receiver's obstacle decision splits on exactly this test, so it has
+    /// to read the type from the same sandbox-anchored `fstatat` that observed
+    /// the entry rather than re-resolving the path: a swap in between would let
+    /// the removal act on a different inode from the one it classified.
+    #[must_use]
+    pub fn is_dir(&self) -> bool {
+        match self {
+            Self::At(meta) => meta.is_dir(),
+            Self::Std(meta) => meta.is_dir(),
+        }
+    }
+
     /// Inode number of the entry.
     #[must_use]
     pub fn ino(&self) -> u64 {
--- a/crates/transfer/src/pipeline/receiver.rs
+++ b/crates/transfer/src/pipeline/receiver.rs
@@ -726,7 +726,7 @@ fn is_permission_error(err: &io::Error) -> bool {
 /// verbatim, with no parenthesised number to invent.
 ///
 /// upstream: `rsync-3.5.0/log.c` `rsyserr()` - `"%s: %s (%d)"`.
-fn upstream_errno_text(error: &io::Error) -> String {
+pub(crate) fn upstream_errno_text(error: &io::Error) -> String {
     let display = error.to_string();
     // A tagged failure is an `io::Error::new(kind, payload)`, i.e. the `Custom`
     // variant, whose `raw_os_error()` is `None` even though the payload wraps a
--- a/crates/transfer/src/receiver/directory/backup.rs
+++ b/crates/transfer/src/receiver/directory/backup.rs
@@ -493,9 +493,16 @@ impl ReceiverContext {
     /// backup mechanism failed; the caller skips creating the replacement,
     /// mirroring upstream `atomic_create` returning 0 when `make_backup` fails.
     ///
+    /// This is the backup ARM of the obstacle decision, not the decision
+    /// itself: it is reached only from
+    /// [`ReceiverContext::make_way_for_replacement`], which has already ruled
+    /// out a directory. Upstream's `dir_in_the_way` (`generator.c:2469`) makes
+    /// `make_backup` unreachable for a directory in both modes, so calling this
+    /// with one would back up something upstream removes.
+    ///
     /// # Upstream Reference
     ///
-    /// - `generator.c:2018-2020` - `if (make_backups > 0 ...) if (!make_backup(fname, ...)) return 0;`
+    /// - `generator.c:2476-2478` - `if (make_backups > 0 && !dir_in_the_way) { if (!make_backup(fname, skip_atomic)) return 0; }`
     #[cfg(unix)]
     pub(in crate::receiver) fn backup_existing_before_replace(
         &self,
--- a/crates/transfer/src/receiver/directory/links.rs
+++ b/crates/transfer/src/receiver/directory/links.rs
@@ -15,6 +15,8 @@ use protocol::flist::{trace_leader_is, trace_looking_for_leader, trace_virtual_f
 
 use crate::generator::ItemFlags;
 use crate::receiver::ReceiverContext;
+#[cfg(any(unix, windows))]
+use crate::receiver::directory::obstacle::MakeWayFor;
 
 impl ReceiverContext {
     /// Reports whether a received symlink's mtime should be preserved.
@@ -190,41 +192,21 @@ impl ReceiverContext {
                     continue;
                 }
                 pre_replace_symlink_meta = fs::symlink_metadata(&link_path).ok();
-                // upstream: generator.c:2018-2020 atomic_create - back the old
-                // symlink up before it is removed when --backup is set; on
-                // backup-mechanism failure upstream skips the entry.
-                match self.backup_existing_before_replace(
-                    &link_path,
-                    relative_path,
-                    dest_dir,
-                    sandbox,
-                ) {
-                    Ok(true) => {}
-                    Ok(false) => {
-                        // SEC-1.g: route the obstacle unlink through the sandbox
-                        // dirfd when the destination parent is the sandbox root,
-                        // so a TOCTOU swap on `link_path` between the readlink
-                        // above and this unlink cannot redirect the syscall to
-                        // an attacker-chosen parent. Falls back to path-based
-                        // `remove_file` otherwise.
-                        let _ = fast_io::unlink_via_sandbox_or_fallback(
-                            sandbox,
-                            dest_dir,
-                            relative_path,
-                            &link_path,
-                            fast_io::UnlinkFlags::File,
-                        );
-                    }
-                    Err(error) => {
-                        debug_log!(
-                            Recv,
-                            1,
-                            "failed to back up existing symlink {}: {}",
-                            link_path.display(),
-                            error
-                        );
-                        continue;
-                    }
+                // upstream: generator.c:2002 atomic_create(..., DEL_FOR_SYMLINK)
+                // - one decision for the obstacle: rmdir a directory, back up
+                // or unlink anything else. The refusal is reported inside.
+                if self
+                    .make_way_for_replacement(
+                        writer,
+                        &link_path,
+                        relative_path,
+                        dest_dir,
+                        sandbox,
+                        MakeWayFor::Symlink,
+                    )
+                    .is_err()
+                {
+                    continue;
                 }
             } else if fast_io::lstat_via_sandbox_or_fallback(
                 sandbox,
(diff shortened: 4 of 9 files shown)


oferchen:
**Measured: this branch appears SUPERSEDED by master. Recommend closing rather than rebasing.**

I picked this up as a conflicting PR and attempted the rebase in a scratch branch (no worktree touched, rebase aborted, nothing pushed). Reporting what the conflict actually is, because it is not an ordinary two-sided merge.

**The conflict shape is the tell.** Rebasing `1faf3c46f` onto master (`92e50abcc`) produces four conflicts, and two of them are **add/add**:

```
CONFLICT (add/add): crates/transfer/src/receiver/directory/obstacle.rs
CONFLICT (add/add): tests/directory_obstacle_removal.rs
```

Both files already exist on master. An add/add on the two files a PR exists to introduce means master gained them independently.

**Direction of the diff settles it.** `git diff origin/master 1faf3c46f` over those two files:

```
 crates/transfer/src/receiver/directory/obstacle.rs |  29 +---
 tests/directory_obstacle_removal.rs                | 162 +--------------------
 2 files changed, 5 insertions(+), 186 deletions(-)
```

Master → PR head is **+5 / -186**. The branch holds strictly *less* than master. Merging it would delete 186 lines of currently-shipping code and tests.

**The 5 lines the branch has and master does not are all doc prose** — no code:

```
//! stands where a symlink, FIFO, socket, or device node has to be created.
    /// Clears whatever stands at `existing` so a fresh non-regular entry can be
    /// created there, reporting any refusal the way upstream reports it.
//! A directory standing where a symlink or special file has to be created is
//! removed, never backed up - and a populated one is refused out loud.
```

So there is no functional residue to salvage: the substance landed, and master's wording has since moved on.

**Recommendation:** close this PR as superseded. If any of that doc wording is preferred over master's, it is a separate one-file docs change, not a rebase of this branch.

⚠ Caveat on scope: I compared only the two add/add files, since those are the ones the PR exists to add. If you believe the branch carries substance in the other two conflicted files (`receiver/directory/mod.rs`, `receiver/tests/munge_symlinks.rs`), that is worth checking before closing — I did not diff those, and a `--stat` on the whole branch would not distinguish supersession from divergence.

oferchen:
**This PR is SUPERSEDED, not merely conflicting — rebasing it would regress master.**

The conflict shape is the tell: two of the three files conflict **add/add**, meaning master
already has a file at each path.

```
CONFLICT (content):  crates/transfer/src/receiver/directory/mod.rs
CONFLICT (add/add):  crates/transfer/src/receiver/directory/obstacle.rs
CONFLICT (add/add):  tests/directory_obstacle_removal.rs
```

Both files landed on master via a different PR — **#7602**, commit `81c158629`
(`fix(receiver): remove a non-regular obstacle before a file transfer`).

Diffing master's copies against this branch's (direction: master → this PR):

| file | delta |
|---|---|
| `crates/transfer/src/receiver/directory/obstacle.rs` | 3 insertions, **16 deletions** |
| `tests/directory_obstacle_removal.rs` | 2 insertions, **160 deletions** |

This branch is a strict **subset**. Resolving the conflict in its favour would remove 176
lines of production code and tests that master already carries, so the 70 green checks here
are green against a tree that no longer reflects the intended end state.

Recommend closing as superseded by #7602 rather than rebasing. If some hunk here is genuinely
absent from master, it should be extracted as its own small PR on top of current master — the
`--stat` deltas above say the residual is subtractive, not additive, but a hunk-level read
would settle it definitively before anything is discarded.

oferchen:
Closing as superseded: master already carries this work, merged via #7602.

The conflict shape is the evidence. Both sides ADD the same new files - `obstacle.rs` and `tests/directory_obstacle_removal.rs` - which is the signature of supersession, not of divergence: two branches that independently introduced the same construct, one of which landed first. Divergent work conflicts on modified lines, not on whole new files existing twice for the same reason.

Rebasing would mean hand-resolving add/add conflicts on files present on both sides for identical reasons, and arriving at an empty residual. Closing records that outcome directly.

The work itself is not lost - it is on master. If a residual behaviour from this branch turns out to be missing from #7602's version, that is a fresh, separately-scoped change rather than a rebase of this one.

PR #7600, opened 2026-09-01
perf(fast_io): stop waiting for the SEND_ZC notification CQE

oferchen:
## The regression

`ZeroCopyPolicy` reaching `IORING_OP_SEND_ZC` on the daemon-sender's socket
made a 200 MB loopback daemon pull **5.8x slower**, not faster.

| cell | median (11 alternating reps) | throughput |
|---|---|---|
| plain socket writer (no cargo feature) | 0.0866 s | 2422 MB/s |
| SEND_ZC (`iouring-send-zc`) | 0.5046 s | 416 MB/s |

## Diagnosis, measured before the fix

`try_send_zc` submits one SQE and blocks until it has seen **both**
completions: the transfer CQE ("the bytes are queued for transmit", carrying
the byte count and `IORING_CQE_F_MORE`) and the notification CQE
(`IORING_CQE_F_NOTIF`, "the kernel released your pages"). Waiting for the
second is the entire cost.

Per-call instrumentation of the daemon-sender for one 200 MB pull, recorded
in memory and dumped once at the end so the trace does not perturb the run
(traced 0.447 s vs untraced ~0.50 s):

```
send count = 6400
xfer_us:  sum=0.084 s  p50=12  p90=19  p99=36
notif_us: sum=0.249 s  p50=28  p90=51  p99=91
total_us: sum=0.333 s
waits: [(1, 10), (2, 6390)]
payload histogram: 32776 -> 6398, 32795 -> 1, 32796 -> 1
```

- The notification wait is **75%** of the time spent inside `try_send_zc`,
  and 0.249 s of the 0.41 s end-to-end gap.
- **Syscall count ruled out**: exactly two `submit_and_wait(1)` entries per
  send in 6390/6400 cases - one for each CQE, never a storm.
- **`SEND_ZC_MIN_BYTES` ruled out**: every send is 32776 bytes (a 32 KiB
  multiplex frame plus its 8-byte header). There are no sub-threshold sends
  at all, so the 16 KiB / 4 KiB threshold is not selecting anything.
- **Per-call ring construction ruled out**: a counter in
  `PerThreadRing::new` reported `per_thread_ring_builds=1` for the whole
  transfer.
- **The transfer CQE is not the problem**: 0.084 s of 0.333 s, p50 12 us,
  which is roughly what the send itself costs.

## The fix

Split the two waits.

`SendZcPipeline` still waits for the **transfer** CQE. That is what carries
the byte count, surfaces `-errno`, handles short sends by resubmitting the
remainder, and - once observed - guarantees the bytes are in the socket's
transmit queue, which is what keeps wire ordering against `submit_send_batch`
and the synchronous path on the same fd. It does **not** wait for the
notification.

### Why that is sound

Only the buffered flush path moves. The two call sites differ in who owns
the memory:

- `flush_buffer` sends from the writer's own buffer. It hands that buffer to
  the pipeline by **swap**, getting back a buffer the kernel has already
  released. No caller ever names the memory the kernel still holds, so there
  is nothing to observe, refill, or free early. A four-buffer pool keeps the
  pipe full and bounds what an un-reaped notification can hold at
  `4 * buffer_size` per connection.
- `write()` with a caller slice at or above `buffer_size` sends memory the
  caller may free the moment `write` returns. It stays on the synchronous
  `try_send_zc`. The SAFETY comments in `flush_buffer` and `write` that
  justified themselves with "try_send_zc waits for the notification CQE
  before returning" are rewritten to say which path they now describe.

Accounting: `pending` is incremented when an SQE is pushed and decremented
exactly once per submission - by the notification CQE, or by a transfer CQE
with `IORING_CQE_F_MORE` clear, which is the kernel saying no notification
will follow. Counting up at submit time keeps it an upper bound at every
instant, so a buffer is never considered free while the kernel might still
be reading it, and the drain always terminates.

### The Drop drain

`SendZcPipeline::drop` drains every outstanding notification before any
pooled buffer is deallocated. It is unconditional: the field is dropped as
part of dropping the writer on the normal path, the error path and during
unwinding. If the drain itself fails, the still-pinned buffers are leaked
with `mem::forget` - leaking memory is sound, freeing memory the kernel
holds a reference to is not. An `EINTR` from `io_uring_enter(2)` resumes the
wait (a syscall restart, not a retry policy), because abandoning it would
force that leak.

Closing the ring fd is not a substitute: io_uring teardown is deferred to
`io_ring_exit_work`, so the kernel's page references can outlive `close(2)`.

### Why a private ring

Deferred notifications stay queued across calls. Every other io_uring
consumer on the calling thread - `submit_send_batch` in particular - drains
the per-thread ring's completion queue unconditionally and demultiplexes on
`user_data` alone, so a notification left there would be consumed by, and
misread as, somebody else's completion. The pipeline owns a ring nothing
else submits to or reaps from. It is one ring per socket writer, i.e. per
connection, so it does not reintroduce the shared-ring serialisation IUR-3
removed; `tools/ci/check_shared_ring_removal.sh` still passes.

`RegisteredBufferGroup` was evaluated and rejected as the vehicle: it is
registered against one specific ring, the SEND_ZC path never sets
`buf_index` so the registration would be inert, and using it forces a memcpy
into the slot - a copy that zero-copy exists to remove.

## Results

Same host, same 200 MB source, 11 alternating reps per cell, `/dev/shm`
destination, page cache warmed once, medians reported.

| A/B | A | B | A median | B median |
|---|---|---|---|---|
| feature-only, before | plain | SEND_ZC | 0.0866 s | 0.5046 s |
| feature-only, after | plain | SEND_ZC pipelined | 0.0735 s | 0.1067 s |
| fix isolated (both feature-on) | before | after | 0.5125 s | 0.1068 s |

**The fix is a 4.8x speedup and removes the regression. It does not make
SEND_ZC beat a plain socket write on loopback** - 0.107 s vs 0.074 s, still
about 1.45x behind.

Post-fix instrumentation says the remaining gap is not pool starvation:

```
send count = 6400
send_slot (transfer CQE waits):        sum=0.046 s  p50=5 us
acquire_free_slot (pool exhaustion):   sum=0.018 s  p50=0  p90=0
pending after each send: 0->1365 1->2470 2->1352 3->724 4->488 5->1
```

Time inside the send primitive drops from 0.333 s to 0.064 s. A free buffer
was available immediately in over 90% of sends, so a deeper pool would
recover at most 0.018 s.

Loopback simply has nothing for zero-copy to amortise: the writer copies the
frame into its own buffer either way, SEND_ZC replaces the kernel's copy
with page pinning, and the receiver copies out regardless. A test that could
show SEND_ZC winning needs a real NIC with scatter-gather and checksum
offload (so the kernel copy it removes is the actual bottleneck), payloads
large enough to amortise `get_user_pages_fast`, and a link fast enough that
the copy, not the wire, is the limit.

## Verification

On the Linux host (kernel 7.0, aarch64), against this branch alone:

- `cargo fmt --all -- --check`
- `cargo clippy -p fast_io --all-targets -- -D warnings`, and again with
  `-F iouring-send-zc`
- `cargo nextest run -p fast_io` (1132 passed) and with `-F iouring-send-zc`
  (1135 passed)
- `tools/ci/check_shared_ring_removal.sh`
- A 200 MB daemon pull through the pipelined writer is byte-identical
  (`cmp` plus matching sha256).

New tests in `send_zc_pipeline`: the swap hands the kernel an allocation the
caller no longer holds; more sends than pool slots round-trip byte-exactly
and drain to zero pending (and drain is idempotent); dropping with a send
outstanding still delivers every byte.

They deliberately do not assert "the notification is still outstanding when
`send_staged` returns". On loopback it usually is not - the packet reaches
the receiving socket immediately, so the kernel posts the notification right
behind the transfer CQE and one quiet send reaps both in a single drain.
What the pipeline changes is that it never *waits*, which is visible only
under load; the A/B above is what pins it.

## Scope

oc-only: upstream rsync has no io_uring path, so there is no upstream
behaviour to mirror. The default build does not enable `iouring-send-zc`, so
the stock socket-send path is untouched.

### DIFF
--- a/crates/fast_io/src/io_uring/mod.rs
+++ b/crates/fast_io/src/io_uring/mod.rs
@@ -116,6 +116,9 @@ pub mod registered_buffers;
 pub mod renameat2;
 /// `IORING_OP_SEND_ZC` zero-copy socket-send primitive (Linux 6.0+).
 pub mod send_zc;
+/// Pipelined `IORING_OP_SEND_ZC` submission that defers the page-release
+/// notification instead of blocking on it.
+mod send_zc_pipeline;
 /// Pool of long-lived io_uring instances shared across consumers in a session.
 pub mod session_pool;
 /// Single io_uring ring shared by a reader fd and a writer fd in one session.
--- a/crates/fast_io/src/io_uring/send_zc.rs
+++ b/crates/fast_io/src/io_uring/send_zc.rs
@@ -12,10 +12,18 @@
 //!    `result()` is unused.
 //!
 //! The buffer passed to a SEND_ZC submission must remain valid and unmodified
-//! until the notification CQE arrives. This module enforces that contract by
-//! blocking on both CQEs before returning. Callers therefore see SEND_ZC as
-//! a synchronous primitive even though the kernel performs the page release
-//! asynchronously.
+//! until the notification CQE arrives. [`try_send_zc`] discharges that
+//! contract the blunt way: it blocks on both CQEs, so callers see SEND_ZC as
+//! a synchronous primitive that may be handed a borrowed slice.
+//!
+//! That is correct but slow. The kernel releases the pages only once the peer
+//! has consumed the data, so waiting for the notification couples the
+//! sender's rate to the receiver's drain rate and keeps exactly one send in
+//! flight. [`super::send_zc_pipeline::SendZcPipeline`] is the throughput
+//! path: it waits only for the transfer CQE and keeps a pool of owned buffers
+//! alive until their notifications land. Use [`try_send_zc`] when the payload
+//! is a caller-owned slice that must be released on return; use the pipeline
+//! when the payload can be moved into a buffer the sender owns.
 //!
 //! See `docs/design/iouring-send-zc.md` for the full design.
 //!
@@ -47,10 +55,10 @@ use super::config::IoUringConfig;
 use super::registered_buffers::RegisteredBufferGroup;
 
 /// CQE flag set on the transfer CQE; signals a notification CQE will follow.
-const IORING_CQE_F_MORE: u32 = 1 << 1;
+pub(super) const IORING_CQE_F_MORE: u32 = 1 << 1;
 
 /// CQE flag set on the notification CQE.
-const IORING_CQE_F_NOTIF: u32 = 1 << 3;
+pub(super) const IORING_CQE_F_NOTIF: u32 = 1 << 3;
 
 /// Sentinel `user_data` mask used to keep the SEND_ZC submission from
 /// colliding with other CQEs draining on the same ring.
--- a/crates/fast_io/src/io_uring/send_zc_pipeline.rs
+++ b/crates/fast_io/src/io_uring/send_zc_pipeline.rs
@@ -1,0 +1,562 @@
+//! Pipelined `IORING_OP_SEND_ZC` submission for the socket write path.
+//!
+//! `IORING_OP_SEND_ZC` posts two completions per submission: a **transfer**
+//! CQE ("the bytes are queued for transmit", carrying the byte count and
+//! `IORING_CQE_F_MORE` when a notification will follow) and a
+//! **notification** CQE (`IORING_CQE_F_NOTIF`, "the kernel has released its
+//! reference to your pages").
+//!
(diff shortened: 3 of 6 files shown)


oferchen:
FYI yes, these are real numbers from my benchmarks.

PR #7617, opened 2026-09-01
test(daemon): pin the listener against a refusal, not a silent drop

oferchen:
test(daemon): pin the listener against a refusal, not a silent drop

`session_error_does_not_stop_the_accept_loop` fails on pristine master. Its
property - one bad client must not take the accept loop down - is correct and
still holds. Both of its supporting assertions describe a daemon neither oc
nor upstream has.

The fixture connects, completes the version exchange, then sends the
`@RSYNCD:` banner a second time. The test called that "an invalid state
transition the connection FSM rejects", and asserted the client is dropped
mid-session with an empty reply plus a `failed to serve legacy handshake`
log line naming an `invalid daemon connection transition`.

Upstream reads exactly ONE line after the version exchange and takes it as
the module name (clientserver.c:1538). A line that names no module is
refused, not dropped - clientserver.c:1568-1570 logs

    unknown module '%s' tried from %s (%s)

and answers

    @ERROR: Unknown module '%s'

before closing. Measured against the 3.5.0 oracle and against oc: both
produce that pair, byte-shape identical, for this exact input. So the daemon
is right and the test is wrong on both counts - the reply is not empty, and
no handshake-failure line is logged because an orderly refusal is not a
session failure. A probe confirmed the second assertion would also have
failed had the first not fired first.

Both assertions now pin the refusal itself:

- the provoker must receive exactly `@ERROR: Unknown module '<banner>'`
- the log must carry exactly one `unknown module '<banner>' tried from` line

That is strictly stronger than the old pair, because it is what the daemon
actually does. The banner is named once as `REPEATED_BANNER` so the fixture
and both assertions cannot drift apart. The module doc comment loses the
false FSM claim and cites clientserver.c:1538-1570.

NON-VACUITY: mutating the fixture so the second line names a real module -
the daemon then SERVES the first client - reddens assertion one
(`left: ["@RSYNCD: OK"]`). The guard discriminates on daemon behaviour, not
on fixture shape.

Verified: 3/3 green at 0.64s each on the corrected file, fmt clean.

### DIFF
--- a/tests/daemon_session_error_keeps_listener.rs
+++ b/tests/daemon_session_error_keeps_listener.rs
@@ -9,9 +9,9 @@
 //! failure and `fork` failure each `continue`. Only listener setup can end the
 //! daemon (`exit_cleanup(RERR_SOCKETIO)` at socket.c:699 and socket.c:715).
 //!
-//! This test pins that contract for the threaded model: client one drives an
-//! invalid `@RSYNCD:` state transition, and client two must still complete a
-//! real transfer against the same listener afterwards.
+//! This test pins that contract for the threaded model: client one is refused
+//! mid-handshake, and client two must still complete a real transfer against
+//! the same listener afterwards.
 
 #![cfg(unix)]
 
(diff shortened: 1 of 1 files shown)


oferchen:
Closing in favour of #7615, which fixes the same master red better.

Both PRs reached the same root cause: the daemon is upstream-faithful, and the test's two supporting assertions describe behaviour neither oc nor upstream has. We diverged on the remedy.

This PR re-pointed the ASSERTIONS at the orderly `@ERROR: Unknown module` refusal the daemon actually produces. That is green and true, but a refusal is not a session error - `report_session_failure` never fires on it - so the test would stop covering the error path its own name claims.

#7615 re-points the PROVOCATION instead, at a non-UTF-8 request line. That still raises a genuine session error, so the `JoinHandle`-carried failure and the accept loop's survival of it stay exercised. The property survives intact rather than being weakened to fit the daemon.

#7615 also carries the bisection this PR lacks: #7579's fallthrough arm merged at 14:09 and #7590's test at 16:39, and #7590's tested head does not contain the arm. Two individually-correct commits, never run together until both were on master.

PR #7620, opened 2026-09-02
test(daemon): report the whole exchange when the accept-loop pin fails

oferchen:
test(daemon): report the whole exchange when the accept-loop pin fails

`session_error_does_not_stop_the_accept_loop` fails on every Linux cell of
every open PR, and its assertion discards the evidence needed to diagnose it:

That names the symptom - the daemon answered the client's VERSION line as if
it were the module request - and then throws away everything that identifies
which line the daemon read first: the greeting the provoker actually received,
the daemon's own log, and both peers' results. The assertion two blocks above
it already carries `log_contents`; this one, the one that actually fires, does
not.

It is not deterministic. #7615's own `nextest (stable)` was SUCCESS on this
exact merge commit, and it fails on other PRs' merge commits with the same
base - so it is a flake that reproduces only in CI. 65 local runs across five
regimes on an aarch64 Linux host all pass: 3x alone (default features), 6x
alone (--all-features), the parallel package run, 24-way concurrent, and
32-way pinned to a single CPU. No x86_64 Linux host is reachable to run the
real cell (`--workspace --all-features`, 33482 tests), so the next CI failure
is the only instrument available - and it has to carry more than it does now.

Two changes, both test-only:

1. `run_provoker` returns a named `ProvokerOutcome { greeting, trailing }`
   instead of a bare `Vec<String>`. The greeting was read and dropped; it is
   the control for "did the daemon speak first at all", which is precisely the
   question the failure raises.

2. The non-vacuity assertion reports the greeting, the trailing lines, both
   results and the daemon log.

Nothing about the property under test moves: the same condition is asserted on
the same value.

Verified: the diagnostic is non-vacuous - mutating the predicate to `false`
makes it fire and print the greeting (`@RSYNCD: 32.0 sha512 sha256 sha1 md5
md4`), both results, and the full daemon log including the
`failed to serve legacy handshake ... stream did not contain valid UTF-8`
line. Restored, re-run, green. `rustfmt --edition 2024 --check` clean.

### DIFF
--- a/tests/daemon_session_error_keeps_listener.rs
+++ b/tests/daemon_session_error_keeps_listener.rs
@@ -50,7 +50,20 @@ fn allocate_test_port() -> Option<(u16, TcpListener)> {
 /// `BufRead::read_line` into a `String`, which fails with `InvalidData` -
 /// a session error that is neither a broken pipe nor a reset, so
 /// `report_session_failure` logs it instead of treating it as a normal close.
-fn run_provoker(port: u16, deadline: Instant) -> Result<Vec<String>, String> {
+/// What the provoking client observed on its own socket.
+///
+/// The greeting is kept alongside the trailing lines because the assertion that
+/// fails is about which line the daemon treated as the module request - and that
+/// is unanswerable from the trailing lines alone.
+struct ProvokerOutcome {
+    /// The daemon's greeting line, as received.
+    greeting: String,
+    /// Everything the daemon sent after the request line. Must be empty: the
+    /// session error drops the connection rather than answering it.
+    trailing: Vec<String>,
+}
+
+fn run_provoker(port: u16, deadline: Instant) -> Result<ProvokerOutcome, String> {
     let target = SocketAddr::from((Ipv4Addr::LOCALHOST, port));
     let mut stream = loop {
         match TcpStream::connect_timeout(&target, Duration::from_millis(500)) {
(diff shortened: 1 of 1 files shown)


oferchen:
Narrowing from a fresh CI capture (run 33582539617, job 100105256144, #7619 `nextest (stable)`), which reproduces the same failure verbatim:

```
```

`@RSYNCD: 32.0 md5 md4` is the provoker's own version line, not the daemon's greeting - the greeting carries the full digest list (`@RSYNCD: 32.0 sha512 sha256 sha1 md5 md4`). So the daemon answered the VERSION line as the MODULE REQUEST.

That pins a necessary condition. At `session_runtime.rs:315-360` the version arm is guarded:

```rust
Ok(LegacyDaemonMessage::Version(version)) if negotiated_protocol.is_none() => { ... }
```

and `negotiated_protocol` has exactly one assignment - that arm. The unknown-module refusal is therefore only reachable for a well-formed banner when the guard has already failed, i.e. **some earlier line on that connection was already consumed as a Version**. The provoker sends exactly two lines.

Excluded by reading, so nobody re-checks them:

- `parse_legacy_daemon_message` is a pure function of the line; it cannot classify one string two ways.
- `reject_malformed_client_greeting` cannot emit "Unknown module" - the sole producer is `UNKNOWN_MODULE_PAYLOAD` via `daemon/sections/module_access/request.rs`.
- No stale buffer: `session_runtime.rs:289` builds a fresh `BufReader::new(stream)` per session, with no pooling.
- `read_line` validates only the bytes up to and including the newline, so line 1 always parses and line 2 (`\xff\xfe`) always errors - which is the passing path.

What supplies the earlier Version line is not answerable from source; the widened assert here is what will say. One caveat worth stating: the diagnostic only helps once it is on master, since it fires nowhere else. Worth merging as soon as this is green rather than holding it.

PR #7638, opened 2026-09-02
test(ci): run the 3.5.0 upstream testsuite on macOS

oferchen:
Adds a fifth upstream-testsuite leg. The four existing ones are all Linux
(`{nonroot,root} x {pipe,tcp}`), so a platform-conditional divergence is
invisible to every required context.

## What the leg buys, measured

Full 345-cell nonroot/pipe run on macOS: **230 passed / 10 failed / 105
skipped**, against 257 / 3 / 85 for the Linux nonroot/pipe leg.

Only **two** of the ten failures are shared with Linux. The other **eight** are
divergences no existing context can observe:

| cell | Linux | |
|---|---|---|
| `basis-xname-traversal` | fail | shared |
| `filter-merge-content-echo` | fail | shared |
| `chmod-setid` | pass | **macOS-only** |
| `daemon-scan-dir-escape` | pass | **macOS-only** |
| `operator-path-partial-dir-daemon` | pass | **macOS-only** |
| `sender-flist-symlink-leak` | pass | **macOS-only** |
| `symlink-race-source` | pass | **macOS-only** |
| `crtimes` | skip | **macOS-only** |
| `macos-setgid-ordinary-mode-regression` | skip | **macOS-only** |
| `partial-protected-regular-retry-policy` | skip | **macOS-only** |

The last three **skip** on Linux, so they had never executed in this repo's CI
at all - including `macos-setgid-ordinary-mode-regression`, a cell named for the
platform that no leg was running.

None of the ten appears in upstream's own `testsuite/skiplist/macos.txt`, so
none is a platform artefact upstream itself waives.

⚠ Reading trap worth naming: `macos.txt` *does* list
`partial-protected-regular-retry-linux`, which is a **different cell** from the
failing `...-retry-policy`. They are not the same row.

## Coverage this leg does not buy

23 cells skip here that run on Linux. Twenty are exactly upstream's own
`macos.txt` entries (`acls-*`, `sparse`, `preallocate`, `simd-checksum`,
`rrsync-*`, ...) and three (`chown`, `devices`,
`daemon-namecvt-empty-response`) are privilege-keyed. The leg produces ~240 real
outcomes, not 345 - the workflow header says so rather than letting the row
count imply otherwise.

## Why a separate workflow

`tools/ci/run_upstream_testsuite.sh` **is** reused verbatim - it carries only
three Linux-isms, each already guarded to degrade safely (a `mountpoint` /
`/proc/mounts` probe behind `command -v`, and a tmpfs concern scoped to
root-on-Linux).

What is not portable is the *workflow*. `_upstream-testsuite.yml`'s job body is
Linux-shaped throughout: an `APT_PACKAGES` env, two `cache-apt-pkgs-action`
steps, two `ensure_apt_packages.sh` verifications, and a `sudo env` re-exec.
Threading a `runs-on` input through it means guarding every one of those - and
it publishes **ten required contexts**, so a mistake there breaks gates with
nothing to do with macOS. Duplicating ~60 lines of YAML is the cheaper and safer
half of that trade.

## Manifest provenance

Generated by the runner's own `EMIT_EXPECT_RESULT`, never typed. It records the
ten failures as the **expected** outcome, so the gate fires on *drift* rather
than on the known baseline - exactly how the Linux legs work today with their
own `fail` rows.

⚠ The header states plainly that a locally-generated baseline is a **prediction**
about the GitHub runner until the first CI run confirms it. The two hosts can
differ in ways this suite is sensitive to - OS version, which optional libraries
upstream's `./configure` finds (and therefore what the oracle rsync can do), and
sysctl-style knobs. If this PR's own run disagrees, the fix is to replace the
file wholesale from that run's `expect-result.macos.nonroot.txt` artifact, not
to edit rows until it goes green.

## Non-root only, by design

`sudo` works on a GitHub macOS runner, but the root-only cells a second leg
would unlock are already in upstream's `macos.txt` skip set or privilege-keyed
away - it would run the same outcomes at twice the cost. Worth adding only if a
measurement shows it moves cells.

## Not yet required

Registering a context in branch protection is a repo-admin action. The manifest
makes the leg non-vacuous from the first run regardless.


oferchen:
## The macOS leg never ran a test — `./configure` aborted

The failing run's artifact is 1275 bytes and holds **no** `expect-result.macos.nonroot.txt`, only a configure log. Upstream's own build died before the suite started:

```
- Failed to find openssl/md4.h and openssl/md5.h for openssl crypto lib support.
- Failed to find xxhash.h for xxhash checksum support.
configure.sh: error: Aborting configure run
```

That is upstream's deliberate design — `configure.ac` aborts whenever `no_lib` is non-empty, so a missing library is never silently compiled out — and it is exactly what the "Record the upstream build toolchain" step asks for. It leaves two ways forward, and they **produce different oracles**:

| | oracle |
|---|---|
| install the libraries | matches the Linux legs |
| `--disable-openssl --disable-xxhash` | diverges from the Linux legs |

Installing is what this takes. `_upstream-testsuite.yml` installs `libxxhash-dev` via `APT_PACKAGES` and the ubuntu image supplies openssl, so **both Linux legs already build an oracle with openssl and xxhash ENABLED**. Disabling them here would mean the two platforms negotiate different checksum lists, and a macOS-only divergence could then be the oracle rather than oc — defeating the comparison this leg exists to make.

## Why the flags, not just the install

Measured on macOS 26.5 / arm64:

| probe | result |
|---|---|
| bare `cc` over both headers | **fails** on `xxhash.h` |
| `+ -I$(brew --prefix)/include` | compiles |
| `+ -I/-L $(brew --prefix openssl@3)` | compiles **and links** |

macOS `cc` does not search the Homebrew prefix, and `openssl@3` is keg-only so its headers are not linked into that prefix at all. **A `brew install` on its own would leave `./configure` aborting exactly as before.**

End-to-end A/B on the real 3.5.0 tree, with the exact flag list `run_upstream_testsuite.sh` passes:

| arm | `configure` rc | `USE_OPENSSL` | `SUPPORT_XXHASH` |
|---|---|---|---|
| without brew flags | **1** — the CI message verbatim | absent | absent |
| with brew flags | **0** | **ENABLED** | **ENABLED** |

## Three details that are load-bearing

- The flags are published as `UPSTREAM_CPPFLAGS`/`UPSTREAM_LDFLAGS` and mapped onto `CPPFLAGS`/`LDFLAGS` in the testsuite step's own `env:`. `$GITHUB_ENV` is process-wide, and bare `CPPFLAGS`/`LDFLAGS` naming the Homebrew prefix would also reach `Build oc-rsync (release)`, where cc-crate build scripts honour them — it could change which zlib/openssl the binary under test links against.
- A compile **and link** probe runs before the flags are published, so a missing library still surfaces as a failure naming the header rather than as a silently feature-poor oracle. Mutation-proven: blanking `CPPFLAGS` reddens the step.
- The source cache key now includes this workflow file. `configure` is skipped when `shconfig` is present, so a tree cached under a different flag set would otherwise be reused with its **original** feature set and no step would notice.

## The manifest is deliberately untouched

`tools/ci/upstream-3.5.0-expect.macos.nonroot.txt` is still the hand-written prediction its own header declares it to be, and **no run has ever validated it**. Editing it now would be guessing twice. The fixed job emits `expect-result.macos.nonroot.txt` as an artifact; the baseline gets taken from that real run and committed as a reviewable diff of outcomes.

So this PR is expected to stay red on `upstream testsuite (macOS)` until that regeneration lands — but red for the right reason (real outcomes disagreeing with a prediction) rather than because the oracle could not be built.

## Verification

- The new `run:` block was extracted back out of the committed YAML with `yaml.safe_load` and executed, rather than testing a hand copy — a hand copy cannot catch a YAML-level quoting or indentation change.
- `actionlint` finding count unchanged at 4, all pre-existing SC2129 in the job-summary step.

PR #7676, opened 2026-09-04
fix(filters): report an operator-named merge file by path and line

oferchen:
A bad rule inside a `.rsync-filter` named by `-F` told the operator nothing
about which file to fix:

    upstream: Unknown filter rule: <rule from /abs/src/.rsync-filter line 1>
    oc:       Unknown filter rule: <rule from a file read earlier>

Both redact the rule TEXT, which is the security property and is unchanged.
oc additionally withheld the LOCATION, and it did so unconditionally.

Upstream makes that a TWO-ARM decision, at exclude.c:1753:

    rule_src_file = named_by_file ? NULL : src_name;

`named_by_file` is `TEXT_FROM_FILE(template)` (exclude.c:60) - whether the
merge file's OWN NAME came from a file. With `-F` the name came from the
command line, so upstream sets `rule_src_file` to the file's path and
`rule_src_line` to the physical line number. Only a merge file NAMED BY
another file's contents collapses to `a file read earlier` (exclude.c:78),
because there the name is itself peer-controlled text.

oc hardcoded the second arm at both parse call sites, so the first was
unreachable. `filters::RuleSource` already had every upstream arm, including
`File { name, line }` and `FileWordSplit { name }` - this is a call-site
defect, not a missing mechanism.

`MergeFileOrigin` is upstream's `template` argument reduced to the one bit
`rule_src_where` reads, and it owns the choice in one place. The loader
supplies it because the loader is the only thing that knows how the file was
named: `Argument` for a `-F` / `--filter=': NAME'` per-dir merge, `NamedByFile`
for a merge whose name came out of another merge file's contents - the site
whose own comment already stated that provenance correctly.

Two details that are load-bearing rather than cosmetic:

  * the line counter increments for EVERY physical line, including blanks and
    comments, matching exclude.c:1759-1760 - so the number names the line the
    operator will see in their editor;
  * the whitespace-split parser passes no line at all. Upstream sets
    `rule_src_line = -1` there (exclude.c:1754) because a word has no line,
    and `FileWordSplit` renders the name alone.

⚠ MY FIRST MUTATION RUN KILLED NOTHING, and that is what found the real gap:
both mutations passed 3/3 because the new test called the parser with a
`RuleSource` literal, so it exercised `rule_text` and never `MergeFileOrigin`
- the function that owns the decision had no test at all. A truth table over
all four (arm x line-present) rows now covers it, and it is a TABLE because
each arm is separately load-bearing: a test on one arm alone is satisfied by
collapsing both, in either direction.

Mutation-proven in ISOLATION, 4 run / 3 passed / 1 failed each (passed+failed
== run, so no fail-fast truncation):

  * Argument arm collapsed to FileReadEarlier (the pre-fix defect) -> lethal
  * NamedByFile arm leaking the path (the security regression) -> lethal

Both kill exactly the truth table; the redaction pin and the argument-echo
control stay green under each.

Measured against the real rsync 3.5.0 binary, same host and fixture
(`.rsync-filter` holding `Zbad-perdir`, `-r -F src/ dest/`): oc now emits the
same redaction AND the same `<path> line 1` location, rc=1 on both.

engine 4990/4990; cli + core 6627/6627. `cargo fmt --all -- --check`, the
whole-tree tools/ci/check_rustfmt_all.py, and
`cargo clippy --workspace --all-targets --all-features -D warnings` clean on
the PINNED 1.88.0 toolchain.

Upstream 3.5.0's `filter-merge-content-echo` goes FAIL -> PASS, closing the
last section of that cell. It was the only remaining failure on the macOS
nonroot stdio-pipe leg, which measured 238 passed / 1 failed / 106 skipped
immediately before this change.


oferchen:
Post-merge measurement, run at this PR head on macOS with the release binary rebuilt from it:

```
----- overall results:
      239 passed
      106 skipped
overall result is 0
```

The full macOS nonroot stdio-pipe leg is green — 238/1/106 before this change, 239/0/106 after, and `filter-merge-content-echo` was the single failure.

PR #7679, opened 2026-09-04
fix(daemon): deliver a refusal to the log, in upstream's words

oferchen:
Two daemon refusals never reached the log correctly. Both are faithful ports of named upstream sites, and they are the log half of the `daemon-refuse-compress-threads-alias` and `daemon-argv-limit` testsuite cells, whose assertions are log greps.

## 1. The refused-option text had two owners and they drifted

Upstream builds it once and delivers it to both destinations:

- `options.c:1409-1423` — `create_refuse_error()` fills `err_buf` with `"The server is configured to refuse --<longName>"`.
- `options.c:907-918` — `option_error()` emits that same buffer via `rprintf(FERROR, RSYNC_NAME ": %s", err_buf)`. A daemon's `FERROR` lands in the log file.

So upstream has ONE string serving as both the peer's `@ERROR` payload and the logged line. oc formatted them independently: the peer saw upstream's wording while the log carried an oc-invented

```
refusing option '--compress-threads' for module 'mod' from localhost (127.0.0.1)
```

that no upstream site emits. `refused_option_message()` is now the single owner, consumed by both `@ERROR` sites in `request.rs` and by the log, which writes

```
rsync: The server is configured to refuse --compress-threads
```

and nothing more. The dropped `module` / `host` / `peer_ip` operands are deliberate: upstream names none of them here, because the connection is already identified by the line's pid stamp and by the preceding `rsync allowed access on module %s from %s (%s)` (`clientserver.c:787`), which oc already emits faithfully.

The log *code* is left as it was. Upstream routes this at `FERROR` while oc records it at info level; both reach the daemon log file, which is the observable this mirrors, and re-coding it would change oc's message routing for a reason this change has not measured.

## 2. The daemon-argument ceiling fired invisibly

`read_args()` reports its own refusal where it happens — `io.c:1477-1478`, `rprintf(FERROR, "too many daemon arguments\n")` — and a daemon's `FERROR` reaches the log. oc sent that refusal to the peer and logged nothing at all, so an operator watching the log saw a connection cut with no recorded reason.

The refusal now reaches the log too, emitted bare: unlike `option_error()` this site adds no `rsync: ` prefix.

## Verification

Both are pinned end to end against a real daemon and its real log file, and mutation-proven in isolation:

| mutation | result |
|---|---|
| restore the oc-invented `refusing option '...'` wording | only `run_daemon_logs_a_refused_option_in_upstream_words` reddens |
| drop the `log_client_args_failure` call | only `run_daemon_logs_an_oversized_client_argv` reddens |

`run_daemon_records_log_file_entries` stays green under both as the negative control. The refusal test also asserts the log does **not** contain `refusing option`, so the two owners cannot silently drift apart again.

The argv test sends exactly `MAX_DAEMON_ARGS - 1` arguments — the count that trips upstream's before-append check — so the daemon consumes every byte written and neither side is left blocking on the other.

`cargo fmt` + whole-tree rustfmt gate clean, workspace `clippy --all-targets --all-features` clean, `-p daemon` 1821/1821, all at the pinned 1.88.0 toolchain.

### DIFF
--- a/crates/daemon/src/daemon/sections/auth_helpers.rs
+++ b/crates/daemon/src/daemon/sections/auth_helpers.rs
@@ -60,15 +60,56 @@ fn log_module_lock_error(
     log_message(log, &message);
 }
 
-fn log_module_refused_option(
-    log: &SharedLogSink,
-    host: &str,
-    peer_ip: IpAddr,
-    module: &str,
-    refused: &str,
-) {
-    let text = format!("refusing option '{refused}' for module '{module}' from {host} ({peer_ip})");
+/// The refusal text upstream builds once and delivers to BOTH the peer and the
+/// daemon log.
+///
+/// upstream: `options.c:1409-1423` `create_refuse_error()` fills `err_buf` with
+/// `"The server is configured to refuse --<longName>\n"`, and the post-`@RSYNCD:
+/// OK` failure branch at `clientserver.c:1254` reaches `option_error()`
+/// (`options.c:907-918`), which emits that same buffer via
+/// `rprintf(FERROR, RSYNC_NAME ": %s", err_buf)`. A daemon's `FERROR` lands in
+/// the log file, so upstream has ONE string serving as both the peer's `@ERROR`
+/// payload and the logged line.
+///
+/// This is a single owner for exactly that reason: oc previously formatted the
+/// peer's text and the log's text independently, and they drifted - the peer
+/// saw upstream's wording while the log carried an oc-invented
+/// `refusing option '...' for module '...'` line that no upstream site emits.
+pub(crate) fn refused_option_message(refused: &str) -> String {
+    format!("The server is configured to refuse {refused}")
+}
+
+/// Logs a refused option in upstream's words.
+///
+/// upstream: `options.c:915` - `rprintf(FERROR, RSYNC_NAME ": %s", err_buf)`.
+/// The logged line therefore carries the `rsync: ` prefix `option_error()` adds
+/// and nothing more: upstream names no module, host or peer here, because the
+/// connection is already identified by the line's pid stamp and by the
+/// preceding `rsync allowed access on module %s from %s (%s)`
+/// (`clientserver.c:787`), which oc emits faithfully.
+///
+/// The log CODE is deliberately left as it was. Upstream routes this at
+/// `FERROR` while oc records it at info level; both reach the daemon log file,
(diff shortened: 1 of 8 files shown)


oferchen:
Measured correction to this PR's framing, from a real TCP-leg run on Linux.

I claimed these two log fixes were what closes `daemon-refuse-compress-threads-alias` and `daemon-argv-limit`. **They are not.** Both cells fail before any logging happens — the daemon never starts:

```
rsyncd exited before listening on port 12934 (status=13)
```

Running the cell's own config directly gives the reason:

```
failed to open log file
'/…/testtmp/daemon-argv-limit/testtmp/daemon-argv-limit/argv-limit.log': No such file or directory (code 13)
```

The relative global `log file` is resolved against the config file's directory, so the path is doubled. Controls:

| binary | result |
|---|---|
| real rsync 3.5.0 | `daemon-argv-limit` **PASS** |
| oc master `cdb2d2416` (this PR's base) | same doubled path, exit 13 |
| oc this branch `bd8fdd54a` | same doubled path, exit 13 |

So the path-doubling is **master's, not this PR's** — this branch is not a regression — and it is tracked separately.

That defect is two divergences from `log.c`'s `logfile_open()`: upstream opens `logfile_name` verbatim (CWD-relative, no config-dir rebase), and on failure it falls back to syslog with `Ignoring "log file" setting.` and **keeps serving** rather than exiting `RERR_MESSAGEIO`.

This PR stands on its own: it is a faithfulness fix for the refusal wording and the missing argv-limit log line, both mutation-proven in isolation. It just does not, by itself, flip those two manifest rows — no manifest row is touched here.

oferchen:
Correcting my earlier comment on this PR, which said this change closes neither `daemon-argv-limit` nor `daemon-refuse-compress-threads-alias`.

That was written from a stale A/B. Measured on Linux at the time, master alone failed both cells at daemon **startup** (`rsyncd exited before listening on port (status=13)`), so the run never reached the assertion this branch's refusal wording satisfies — which is why the change looked inert. That startup failure was fixed on master by #7677.

The current CI run on this branch (33929533383) reports both cells as XPASS on **both** TCP legs, and they were the run's only mismatches on either leg:

    daemon-argv-limit: expected fail, got pass (xpass)
    daemon-refuse-compress-threads-alias: expected fail, got pass (xpass)

So this branch closes both on its own. Manifest rows flipped `fail` -> `pass` in both TCP files in 33f241403.

PR #7680, opened 2026-09-04
fix(daemon): open operator path parameters verbatim and log the banner before readiness

oferchen:
Three commits, all daemon startup, all converging on `daemon-standalone-detach`
- which is `pass` in both TCP manifests and is what makes master red.

**⚠ Correction to my earlier message on this PR.** I claimed the `pid file`
fix closes the detach cell in CI. That is refuted by the CI log: the failure
there reports an existing-but-truncated log with the pid file *present*, so
the paths were already correct on that leg. Both defects are real; only the
third commit moves the CI row. The measured table is below.

### 1-2. `log file` and `pid file` are opened verbatim

`loadparm.c` stores every daemon parameter value as written and does no path
rebasing, so a relative value resolves against the daemon's **working
directory**, never the config file's directory:

- **`log file`** - `log.c:216-218` `log_init()` takes `logfile_name` from
  `lp_log_file(module_id)`; `log.c:169` `logfile_open()` hands it straight to
  `open_no_attacker_symlinks()`. Three sites in oc (module value, global
  module-default, global startup value) read one config line through one
  `lp_log_file()` accessor upstream, so rebasing one and not the others would
  open two different files for a single directive.
- **`pid file`** - `clientserver.c:1584` `create_pid_file()` opens the value
  `lp_pid_file()` returns, directly.

oc prefixed the config file's directory, so a relative value with a directory
component landed in a doubled path that `create_dir_all` obligingly created -
silently, because the write then succeeded. upstream's own `runtests.py`
defaults to `scratchbase=./testtmp`, so on that default the detach cell's pid
file lands somewhere the harness never looks.

The shared helper is `daemon_parameter_path`, named for loadparm's storage
rule rather than as a `log file` special case.
`resolve_config_relative_path` stays for `include` / `motd file` and the other
directives upstream really does resolve differently.

### 3. The banner is logged before the pid file appears

Upstream logs `rsyncd version %s starting, listening on port %d`
(`clientserver.c:1770`) and only then calls `start_accept_loop()` (`:1776`),
which is where `socket.c` binds *and* listens - so reaching the port implies
the banner is already in the log.

oc binds earlier on purpose (a bind failure must reach the launching
terminal's stderr, not a detached child's log), so the socket is already
listening by the time the daemon detaches and that guarantee cannot be
restated as "log before listen". oc emitted the banner after the pid file,
after the chroot and after the privilege drop - behind every readiness signal
an observer acts on. It now precedes the pid file, which is the signal
operators, init systems and the cell all wait on. That inverts upstream's line
order and is the deliberate cost of the earlier bind; the property upstream
actually provides is preserved.

### Measured

Linux, upstream 3.5.0 suite, `daemon-standalone-detach` only:

| tree | scratchbase | result |
|---|---|---|
| master `cdb2d2416` | absolute (CI's shape) | 3 passed / 7 failed of 10 |
| + `pid file` verbatim | absolute | 8 passed / 17 failed of 25 |
| + banner ordering | absolute | **25 passed / 0 failed of 25** |
| + banner ordering | relative (`runtests.py` default) | **10 passed / 0 failed of 10** |

The recurrence guard is that cell: a required context on both TCP legs, which
reproduced the ordering defect at 17/25.

Each parser arm is mutation-proven in isolation, restoring
`resolve_config_relative_path` at one site at a time:

| mutation | result |
|---|---|
| global `log file` rebased | 27 passed / 1 failed - only `parse_global_log_file_relative_is_kept_verbatim` |
| module `log file` rebased | 27 passed / 1 failed - only `parse_module_log_file_relative_is_kept_verbatim` |
| `pid file` rebased | 41 passed / 2 failed - exactly the two pid-file tests, all four log-file tests green |

`runtime_options_loads_pid_file_from_config` asserted the config file's
directory as the prefix, so it encoded the behaviour being removed. It is
re-pinned against the upstream value with the citation, not preserved.

`cargo fmt --all -- --check` clean; `tools/ci/check_rustfmt_all.py` clean over
3401 files; `-p daemon` clippy `--all-targets --all-features -D warnings`
clean.

### Not fixed here

`logfile_open()`'s failure path is non-fatal upstream (`log.c:175-182`): it
falls back to syslog, prints `Ignoring "log file" setting.` and keeps serving,
where oc exits `RERR_MESSAGEIO` from five startup call sites. Separate
divergence, deliberately out of scope.

### DIFF
--- a/crates/daemon/src/daemon/sections/config_parsing/global_directives/dispatch.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/global_directives/dispatch.rs
@@ -142,7 +142,7 @@ fn apply_global_directive(
                 ));
             }
 
-            let resolved = resolve_config_relative_path(path, trimmed);
+            let resolved = daemon_parameter_path(trimmed);
             store_global_directive(&mut state.pid_file, resolved, canonical, line_number);
         }
         "reverselookup" => {
@@ -594,7 +594,7 @@ fn apply_global_directive(
         // lost - measured against rsync 3.5.0, which writes three of them here.
         "logfile" => {
             if !value.is_empty() {
-                let resolved = log_file_path(value);
+                let resolved = daemon_parameter_path(value);
                 state.module_defaults.log_file = Some(resolved.clone());
                 store_global_directive(&mut state.log_file, resolved, canonical, line_number);
             }
--- a/crates/daemon/src/daemon/sections/config_parsing/global_directives/dispatch.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/global_directives/dispatch.rs
@@ -594,7 +594,7 @@ fn apply_global_directive(
         // lost - measured against rsync 3.5.0, which writes three of them here.
         "logfile" => {
             if !value.is_empty() {
-                let resolved = resolve_config_relative_path(canonical, value);
+                let resolved = log_file_path(value);
                 state.module_defaults.log_file = Some(resolved.clone());
                 store_global_directive(&mut state.log_file, resolved, canonical, line_number);
             }
--- a/crates/daemon/src/daemon/sections/config_parsing/module_directives.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/module_directives.rs
@@ -259,7 +259,7 @@ fn apply_module_directive(
                     "'log file' directive must not be empty",
                 ));
             }
-            builder.set_log_file(log_file_path(value));
+            builder.set_log_file(daemon_parameter_path(value));
         }
         "dontcompress" => {
             let patterns = if value.is_empty() {
--- a/crates/daemon/src/daemon/sections/config_parsing/module_directives.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/module_directives.rs
@@ -259,8 +259,7 @@ fn apply_module_directive(
                     "'log file' directive must not be empty",
                 ));
             }
-            let resolved = resolve_config_relative_path(canonical, value);
-            builder.set_log_file(resolved);
+            builder.set_log_file(log_file_path(value));
         }
         "dontcompress" => {
             let patterns = if value.is_empty() {
--- a/crates/daemon/src/daemon/sections/config_parsing/parser.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/parser.rs
@@ -402,20 +402,24 @@ fn continuation_offset(line: &str) -> Option<usize> {
     trimmed.ends_with('\\').then(|| trimmed.len() - 1)
 }
 
-/// Resolves an operator-supplied `log file` value.
+/// Resolves an operator-supplied daemon path parameter.
 ///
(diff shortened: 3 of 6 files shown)


oferchen:
Master's TCP nonroot leg confirms the scope: run 33927388946, job 101200383155
reports exactly one mismatch and nothing else.

```
----- expected-result mismatches:
      daemon-standalone-detach: expected pass, got fail
overall result is 1
```

That is the cell this PR closes, and it is master's only red on that leg - so
the banner-ordering commit is the whole of what stands between master and a
green required context here.

oferchen:
### Full four-leg Linux measurement

Branch tree `283183ecb` (tree-identical to the pushed head), upstream 3.5.0
suite, this repo's own runner and the committed manifests, all four legs run
sequentially on one aarch64 host:

| leg | passed | failed | skipped | expected-result mismatches | overall |
|---|---|---|---|---|---|
| nonroot pipe | 259 | 0 | 86 | `simd-checksum` | 1 |
| root pipe | 287 | 1 | 57 | `simd-checksum`, `batch-file-symlink` | 2 |
| nonroot tcp | 110 | 12 | 33 | *none* | **0** |
| root tcp | 123 | 17 | 15 | *none* | **0** |

Both TCP legs are clean against the committed manifests, and the nonroot TCP
leg reports:

```
PASS    daemon-standalone-detach
```

That is the cell, and the leg, that master fails - run 33927388946 job
101200383155, where it is the sole mismatch. So this branch closes master's
only red.

The `FAIL` rows on the TCP legs are all `fail` in the manifest already. Two of
them, `daemon-argv-limit` and `daemon-refuse-compress-threads-alias`, are the
cells #7679 fixes; they failing here is the expected result for a tree without
that branch, and it corroborates #7679's manifest flip from the other side.

### The two pipe-leg mismatches are NOT from this branch

Both are architecture artefacts of the measuring host, and both are present on
master. Neither manifest row was re-baselined.

**`simd-checksum`** - `Makefile.in:543` builds `simdtest` from
`simd-checksum-x86_64.cpp` alone, so on aarch64 the target produces a zero-byte
non-executable and the cell's own guard
(`simdtest.is_file() and os.access(simdtest, os.X_OK)`) skips it. The manifests
were generated on x86_64, where `pass` is correct.

**`batch-file-symlink`** - A/B measured, same runner, same manifest, same host:

| root pipe leg | passed | failed | skipped | mismatches |
|---|---|---|---|---|
| this branch `283183ecb` | 287 | 1 | 57 | `simd-checksum`, `batch-file-symlink` |
| master `cdb2d2416` | 287 | 1 | 57 | `simd-checksum`, `batch-file-symlink` |

Identical. And `upstream testsuite (root)` is SUCCESS on master's x86_64 CI, so
the cell is host-conditional rather than a live regression on either tree. It
is structurally unreachable from this diff in any case - daemon config parsing
and accept-loop ordering versus `--read-batch` operand validation - but the A/B
is what establishes that, not the reasoning.

PR #7682, opened 2026-09-05
fix(daemon): report a refused name-converter token instead of swallowing it

oferchen:
Closes the `daemon-namecvt-newline-token` cell of the upstream 3.5.0 testsuite, and the deeper defect found while closing it.

## The cell

`daemon-namecvt-newline-token_test.py` asserts two things:

1. a peer-supplied name carrying a control byte never reaches the name converter, and
2. the daemon **reports** `invalid name-converter token` — in the client bytes or in the daemon log.

oc already satisfies (1). `is_safe_token` is byte-for-byte upstream's `namecvt_safe_token` (`clientserver.c:1362`, rejecting `ch < ' ' || ch == 0x7f`), and it runs before anything is written to the converter pipe, so the stream desync CVE-2026-53788 describes is structurally impossible here. oc failed (2): it returned `ConverterOutcome::Refused` silently.

## Why the obvious one-line fix would have shipped inert

Emitting the diagnostic at the guard is correct and insufficient. The post-transfer drain in `execute_transfer` read `drain_events_coded(LogCode::Log)`, so an `FERROR` raised mid-session was produced, was correct, and was **discarded before any sink saw it**. With only part one applied the cell still fails.

That narrow filter is itself the divergence. Upstream does not pick codes for the daemon log:

```c
} else if (am_daemon || logfile_name) {          /* log.c:310 */
        ...
        logit(priority, buf);                    /* every code that reaches here */
        ...
}
if (code == FLOG)                                /* log.c:329-331 */
        return;
```

`FLOG` is not special-cased *into* the log — it is special-cased *out of* the stream switch that follows.

## Shape

`drain_events_for_daemon_log()` states that rule once, in `crates/logging` beside the existing `drain_events_for_peer()`, so the routing policy lives with its siblings instead of as a predicate in the daemon. It takes every code except `FNONE`, which `rwrite()` is never handed at all (`rsync.h:276` "never sent").

## Verification

**Mutation-proven in isolation.** Dropping the emission fails only `a_refused_token_is_reported_not_merely_unresolved` (19 passed / 1 failed). Narrowing the drain back to `FLOG` fails *both* that test and `the_daemon_log_takes_every_code_it_is_handed_not_only_flog` (`left: [Log]`, `right: [Log, Error, Warning, ErrorXfer, Info]`) — the coupling made visible, and the reason part one could not ship alone.

Each test carries its own non-vacuity control: an accepted name reaches the converter and reports nothing, and the excluded `FNONE` event is shown still sitting in the buffer for a later drain rather than dropped.

**Blast radius measured, not assumed.** The widened drain runs after `run_daemon_transfer` returns, so anything the receiver pipeline already framed to the peer through `drain_events_for_peer` is gone from the buffer and cannot be logged twice. That ordering also states the residual honestly — upstream's `rwrite()` reaches `logit()` and *then* the `am_server` frame, feeding both sinks, whereas a drain removes — so on the receiver role an `FERROR` reaches the peer instead of the log. Widening here can only add deliveries, never move one. It is written into the comment rather than left for the next reader.

`cargo fmt --all -- --check`, `clippy --workspace --all-targets --all-features -D warnings`, and `nextest -p logging -p daemon` (2171/2171) all clean on the pinned 1.88.0 toolchain.

### DIFF
--- a/crates/daemon/src/daemon/sections/module_access/transfer/streams.rs
+++ b/crates/daemon/src/daemon/sections/module_access/transfer/streams.rs
@@ -347,11 +347,25 @@ fn execute_transfer(
     // matching upstream rsync's socket I/O model.
     let result = run_daemon_transfer(config, handshake, read_stream, write_stream);
 
-    // upstream: log.c:290-305 - FLOG-classified diagnostics emitted while the
-    // transfer ran belong in the daemon log only; they never travel to the
-    // client. Consume them from the thread-local queue by log code.
+    // Diagnostics emitted while the transfer ran go to the daemon log.
+    //
+    // upstream: log.c:310-327 `else if (am_daemon || logfile_name)` calls
+    // `logit()` for EVERY code that reaches it, not for `FLOG` alone - `FLOG`
+    // is merely the code that `return`s afterwards instead of also reaching
+    // the stream switch. So a daemon that raises `FERROR` mid-session logs it
+    // too, and a drain that took only `FLOG` would discard it: the message
+    // would exist, be correct, and never be delivered.
+    //
+    // This runs after `run_daemon_transfer` returns, so anything the receiver
+    // pipeline already framed to the peer through `drain_events_for_peer` is
+    // gone from the buffer by now and cannot be logged twice. That ordering
+    // also states the residual honestly: upstream's `rwrite()` reaches
+    // `logit()` and then the `am_server` frame, feeding BOTH sinks, whereas a
+    // drain removes - so on the receiver role an `FERROR` reaches the peer
+    // instead of the log. Widening here can only add deliveries, never move
+    // one.
     if let Some(log) = ctx.log_sink {
-        for event in logging::drain_events_coded(logging::LogCode::Log) {
+        for event in logging::drain_events_for_daemon_log() {
             let (logging::DiagnosticEvent::Info { message, .. }
             | logging::DiagnosticEvent::Debug { message, .. }) = event;
             log_message(log, &rsync_info!(message).with_role(Role::Daemon));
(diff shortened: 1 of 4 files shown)


oferchen:
## Measured on the Linux TCP leg: this does **not** close `daemon-namecvt-newline-token`

I applied this PR's diff on top of master and ran the cell it names, on the transport that actually runs it (`USE_TCP=yes DAEMON_TESTS_ONLY=yes`, aarch64 Linux, release binary). The leg is **112 passed / 10 failed** both before and after — byte-identical failure set, `daemon-namecvt-newline-token` included.

That is worth stating plainly because the cell had never been executed against this change before; the emission and drain were verified by unit test only.

### Why it does not fire

The cell's own log:

```
converter requests: []
... [pid] receiving file list
... [pid] oc-rsync error: transfer failed to localhost: module=recv error=multiplexed …
```

The converter is **never consulted**. This PR's guard lives at `query()`'s `is_safe_token` check — correct placement, and correct once a request is made — but on this path no request is ever made, so the refusal has nothing to refuse.

### The cell's oracle is two-part, and oc passes the wrong half

```python
if any(line == 'name' for line in cvt_lines):   # (a) no split-token leak
    test_fail(...)
if not rejected:                                 # (b) the refusal must be REPORTED
    test_fail(...)
```

oc satisfies (a) **vacuously** — `cvt_lines` is empty because nothing was ever written to the pipe — and fails (b). That is immunity by *absence*, not immunity by *defence*, and a log-grep assertion cannot be satisfied by absence. Upstream reaches the converter (`recv_user_name` → `recv_add_id` → `user_to_uid` → `namecvt_call`, during file-list receipt, per the cell's own header) and refuses there with a message.

### What the real fix is

Not this diff. The daemon receiver has to actually route incoming file-list user/group names through the configured `name converter` under `numeric ids = no`, and refuse control chars at that call. Until that path exists, any guard on `query()` is unreachable from this cell.

### What this PR still is

The emission is upstream-faithful (`clientserver.c:1317-1319`), the widened daemon-log drain is a real and separately-measured fix, and both are mutation-proven. I am not proposing to close it — only correcting the claim that it closes this testsuite cell. It does not, and the manifest row must not be flipped on the strength of it.

PR #7686, opened 2026-09-05
fix(daemon): the PROXY header parse must not consume the greeting behind it

oferchen:
A proxy sends the PROXY header and the client's first rsync bytes in the
same segment. `parse_proxy_header` wrapped the connection in a `BufReader`,
which pulls up to its capacity on the first `read_exact` and discards the
surplus when it drops - so `@RSYNCD: 30.0\nmod\n` was swallowed with the
header. The daemon then read an empty module name and answered with a
module LIST where the client expected `@RSYNCD: OK`.

The parser itself was never wrong. `parse_v1_header` already reads the v1
line one byte at a time, a faithful mirror of upstream's

    while (1) {
            if (read_buf(fd, p, 1) <= 0)
                    return 0;

at clientname.c:245-247, whose entire purpose is to stop exactly at the
newline because rsync protocol data follows on the same fd. The v2 path is
equally exact-sized. Only the caller over-read, and the doc comment now
states why the stream is passed through unbuffered.

The new test goes through the wrapper over a real socket on purpose. The
generic `parse_proxy_header_from` is exact-read and passes over a `Cursor`
whether or not the wrapper buffers, so a test at that level cannot see this
defect - it lived in the caller, not the parser. Restoring the `BufReader`
kills the new test and leaves all three `Cursor`-based siblings green,
which is exactly the discrimination it claims:

    4 tests run: 3 passed, 1 failed

daemon suite green; fmt and clippy clean on the pinned 1.88.0 toolchain.

## Testsuite context

Found while working the Linux nonroot TCP daemon leg's `daemon-proxy-protocol`
cell. It is the *second* defect in that cell: #7685 gets the daemon to start
(it was refusing `hosts allow =` at parse time), and this one gets the
handshake past the header. The two are independent in code and this PR is
based directly on master, not stacked.

### DIFF
--- a/crates/daemon/src/daemon/sections/proxy_protocol.rs
+++ b/crates/daemon/src/daemon/sections/proxy_protocol.rs
@@ -211,14 +211,76 @@ fn parse_v1_header<R: Read>(prefix: &[u8; 12], reader: &mut R) -> io::Result<Opt
 ///
 /// upstream: clientserver.c:1298 - `read_proxy_protocol_header()` is called
 /// before any rsync protocol data when `proxy protocol = true` in the config.
+///
+/// ⚠ The stream is passed through **unbuffered, deliberately**. The PROXY
+/// header is immediately followed on the same connection by the client's
+/// `@RSYNCD:` greeting, and a proxy normally forwards both in one segment, so
+/// any read that consumes past the header destroys the greeting. That is why
+/// [`parse_v1_header`] reads a byte at a time - it mirrors upstream's
+/// `read_buf(fd, p, 1)` loop (clientname.c:245-247), whose entire purpose is
+/// to stop exactly at the newline. Wrapping this call in a `BufReader` defeats
+/// that: the reader pulls up to its capacity on the first `read_exact`, and
+/// the surplus is discarded when it drops. Measured symptom: the daemon read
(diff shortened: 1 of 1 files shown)


oferchen:
## MEASURED: the cell closes

Linux nonroot TCP daemon leg (`USE_TCP=yes DAEMON_TESTS_ONLY=yes`), run on a
tree that is `origin/master` (9ee2115e6) plus exactly two cherry-picks: #7685's
`hosts allow =` fix and this one. Release binary rebuilt from that tree.

```
before (master only):   112 passed / 10 failed / 33 skipped
after  (both fixes):    113 passed /  9 failed / 33 skipped

daemon-proxy-protocol:  FAIL -> PASS
```

The remaining 9 are unrelated and unchanged: `daemon-handshake-timeout`,
`daemon-namecvt-newline-token`, `daemon-zstd-thread-exhaustion`,
`malicious-server-partial-basis-symlink-overwrite`, and the five `proto-*`
cells (one shared root cause, receiver INC_RECURSE, tracked separately).

Two fixes were needed because the cell fails in two stages. #7685 gets the
daemon to start at all — it was refusing the config's bare `hosts allow =` and
exiting 1 before it ever listened. Only with the daemon listening does the
PROXY handshake run, and that is where this defect lives. Neither fix alone
moves the row; the counts above are for both applied.

PR #7719, opened 2026-09-06
fix(daemon): fork per connection so chroot cannot leak between sessions

oferchen:
The daemon served every accepted connection on a **thread**. `chroot(2)` and the
working directory are process-wide, so a session's chroot leaked into every later
connection served by the same process. That is the mechanism behind the root-leg
chroot cluster in the upstream-3.5.0 testsuite.

Upstream forks instead — `socket.c:753-772` `start_accept_loop()`. This routes
`SessionWorker` onto that contract on Unix.

## Measured

Root TCP leg, upstream 3.5.0 testsuite, against the committed expect-manifest:

| | before | after |
|---|---|---|
| pass | 131 | **135** |
| fail | 9 | **5** |
| skip | 15 | 15 |

Exactly four rows moved, every one of them `fail` -> `pass`. Nothing regressed:

```
daemon-chroot
daemon-chroot-munge-default
daemon-module-private-parent
nonroot-restrictive-perms
```

The manifest in this PR is **regenerated** from that run (`EMIT_EXPECT_RESULT`)
and verified byte-identical to the committed copy by sha256 — never hand-edited.

Also: `cargo nextest run -p daemon` 1875/1875, including the live TCP session
tests; `tools/ci/check_rustfmt_all.py` over 3422 files clean; full-workspace
`clippy --all-targets --all-features -D warnings` clean. Toolchain 1.88.0, the
pin — not stable.

## Design notes worth a reviewer's attention

**`SessionBacking` is cfg-split by PLATFORM, not by a runtime enum.** Windows has
no `fork`; making the choice dynamic would put a permanently unreachable arm in
every Unix build. The split is a property of the target, and the type says so.

**The child closes every listener it inherited** before serving
(`AcceptEngine::listener_fds` -> `close_inherited_listeners`, upstream
`socket.c:753-760`). Without this a live child holds the port bound, and a daemon
restart races its own outgoing sessions for the address.

**Reaping stays per-pid.** A bulk `waitpid(-1, ..)` is the obvious shape and is
wrong here: it would consume the statuses of the daemon's *other* children — the
name converter, the `exec` hooks, the auth helper — which `std::process::Child`
reaps through its own `wait`.

**Upstream's "child closes the pid-file fd" step is N/A, by measurement, not by
omission.** `PidFileGuard` holds a `PathBuf` and no descriptor. The property that
actually matters — the child must not unlink the parent's pid file — is delivered
structurally by `exit_child`'s `_exit`, which runs no `Drop`.

## Deliberate residual

The QUIC acceptors are **not** closed in the child: `QuicAcceptor` exposes no
descriptor, and those listeners are inert until the QUIC session handoff exists.
Named here rather than left implicit.

## On the tests

The reap assertions need a real oracle for "the forked child has ended", and
`try_reap` consumes the worker by design, so there is no borrowing readiness
predicate to poll. The fixture creates a pipe before the fork; the child inherits
the write end and never touches it, and the parent reads EOF — which the kernel
delivers at `_exit`. A byte written by the child before exiting would prove only
that it got *close* to exiting, and would leave the race in.


oferchen:
### Blocking prerequisite: the per-module connection cap is process-local

`daemon_caps_concurrent_module_connections_at_max_connections` is the only
remaining failure (full-workspace nextest on Linux: 33417 run, 33416 passed).
It is a real regression from fork-per-connection, not a fixture artefact.

`ModuleRuntime::acquire` enforces `max connections` through **two** mechanisms:

- `connection_limiter.acquire(..)` — a lock file, cross-process
- `acquire_local_slot(..)` — a process-local `AtomicU32`

Under thread-per-connection they agree. Under fork-per-connection each admitted
session runs in a child, so its increment lands in a copy-on-write address space
the parent and siblings never observe. The refusing connection reports
`current=` from a stale snapshot, which is the assertion that fails. Enforcement
itself survived only because the lock file is what actually refused.

Upstream has no in-process counter. `clientserver.c:791` is

    if (!claim_connection(lp_lock_file(i), lp_max_connections(i))) {

and `lock_file` carries `DEFAULT_LOCK_FILE` (`rsync.h:33`), so the lock file is
always present and is the sole owner of both the decision and the count.

oc's `lock_file()` is an `Option` with no default. Where a module has no lock
file configured, `acquire_local_slot` is the only enforcement — and under fork
that is per-process, i.e. the cap stops being enforced across sessions. Patching
only the `current=` field would leave that hole, so this PR should not merge
until the cap has a cross-process owner.

oferchen:
### Resolved: the cap count now comes from the mechanism that refused

Follow-up to the note above, which said this should not merge until the cap has
a cross-process owner. Measuring the failing test settled both halves.

**The reported count.** `handle_max_connections_exceeded` read `current` back
from the module's process-local `AtomicU32` *after* the refusal had already been
decided. Under fork-per-connection that counter is not merely stale, it is
structurally blind: `try_acquire_connection`'s only production call site is in
the per-connection path, so a child starts at zero and `acquire_local_slot` can
never refuse. The lock file is what refuses; the atomic supplied the number.

Fixed by carrying the observed count on the refusal itself, so each of the three
refusal sites reports what it saw — the unix lock file reports
`limit.slot_count()` (every slot held), the counting lock file reports the value
it read out of the file, and the process-local slot reports its CAS observation.
A disabling negative `max connections` reports zero: refused by configuration,
not by capacity. `handle_max_connections_exceeded` no longer takes the module,
because it no longer reads state the refusal did not produce.

Verified on Linux at the pinned toolchain (1.88.0): whole-tree rustfmt clean,
`clippy -p daemon --all-targets --all-features -D warnings` clean, the daemon
crate's 40 limiter/cap tests pass, and
`daemon_caps_concurrent_module_connections_at_max_connections` — the sole
remaining failure — now passes. The RED baseline is the parent commit, where
that test failed on `current={CONNECTION_CAP}`.

**The enforcement gap is separable and pre-existing.** A module that sets
`max connections` with no `lock file` has no cross-process owner, because oc's
`lock_file()` is an `Option` with no default while upstream's carries
`DEFAULT_LOCK_FILE` (`rsync.h:33`) and always claims through it
(`clientserver.c:791`). That hole is not introduced here: the `--server` stdio
and inetd paths have each been their own process all along, so the atomic was
already blind on them — which is what #7440 addressed by wiring the lock file
through. Fork widens the affected path set; it does not create the divergence.

Adopting upstream's default is the faithful fix and is confined by upstream's
own ordering — `claim_connection` returns success for a zero limit *before*
opening the lock file (`connection.c:26-46`), which oc mirrors, so only modules
that explicitly set `max connections` would ever touch the default path. That is
a config-default change with its own operator-visible failure mode, so it goes in
its own PR rather than riding along here.

PR #7727, opened 2026-09-07
fix(fast_io): anchor a nested receiver op with the portable confined walk off Linux

oferchen:
## The defect

`anchor_parent()` (`crates/fast_io/src/dir_sandbox/at_syscalls/nested.rs`) pins the parent of a multi-component receiver path on a dirfd before the terminal `*at` op. On Linux it does that with `openat2(RESOLVE_BENEATH)`. Off Linux it returned `ParentAnchor::Fallback` unconditionally, and the caller then took `ConfinedFallback::confined()` — whose ownership walk **trusts a euid-owned symlink** (`confined_fallback.rs` -> `owner_trusted_parent_kind`) and whose escape check `outside_session_root()` returns `false` when `SESSION_ROOT` is `None`.

A non-daemon client pull installs no confinement root, so off Linux there was nothing left to refuse a peer-planted `--partial-dir` symlink. `remove_partial_basis_confined()` (`crates/transfer/src/disk_commit/process/commit.rs:256-271`) followed it and unlinked a file outside the destination tree.

That is upstream testsuite cell `malicious-server-partial-basis-symlink-overwrite`: `fail` on both macOS root legs, `pass` on both Linux legs. **The escaping component is the PREFIX `.rsync-partial`, not the leaf** — `unlink(2)` never follows a terminal symlink, but it resolves every earlier component.

## Why this is upstream's own answer, not a policy invention

Upstream does not derive a confinement root from the destination argument, and it is not unconfined. It confines this **structurally**, by anchoring the transfer-path resolver at the process CWD — which for a receiver IS the destination after `change_dir()` — and it selects that resolver on relative-vs-absolute, not on a policy flag:

- `receiver.c:1065-1071` — a relative `FNAMECMP_PARTIAL_DIR` basis goes to `secure_relative_open(NULL, fnamecmp, O_RDONLY, 0)`. With `basedir == NULL` the anchor is `AT_FDCWD`.
- `syscall.c:2891-2965` `ds_descend()` — a held-dirfd stack, one `O_RDONLY|O_DIRECTORY|O_NOFOLLOW` open per component. Refuses an **absolute** symlink target outright (`:2953-2956`, `errno = ELOOP`), refuses `..` above the anchor (`:2896-2899`), and **follows** a relative in-tree directory symlink (`:2961`).
- `syscall.c:142-143` — `confinement_root()` is NULL for a non-daemon caller. The anchor descriptor is the confinement; no root is involved.

oc already carries a faithful port of `ds_descend` as `ConfinedWalk` (`crates/fast_io/src/dir_sandbox/mod.rs:836-896`), used today only by `open_dest_anchor_confined`. This PR exposes it as `DirSandbox::open_subdir_confined()` and routes the non-Linux `anchor_parent` arm through it. The two arms now agree on policy and differ only in errno: `EXDEV` from the kernel, `ELOOP` from the walk.

## Deliberately not done

- **No client-side confinement root derived from the destination operand.** `SESSION_ROOT` is process-global; arming it would arm every `PathKind::Confined` site at once — the `--delete` path, `--backup-dir`, `--temp-dir`, `--link-dest`/`--copy-dest`/`--compare-dest`, merge files — and a legitimate absolute `--backup-dir` outside the destination tree would become `ELOOP`.
- **The cleanup unlink is not the defence.** `receiver.c:1295-1297` wraps `do_unlink_at(partialptr)` in `operator_path_resolve = 1`, whose ownership walk follows a euid-owned symlink and whose escape check is inert without a root. If control reached that line, upstream would escape too. Upstream's **ordering** is the defence: `receiver.c:1093-1107` resets `fnamecmp = fname` when the secure open refuses, disarming the cleanup unlink before it can run.

## The old arm was the wrong resolver in BOTH directions

oc has two resolvers because upstream has two, and they answer different questions:

- `ds_descend` (`syscall.c:2891-2965`) is for **peer-derived transfer paths**. It confines by anchor descriptor and does **not** look at ownership.
- the ownership walk (`syscall.c:304`, `:406`) is for **operator-supplied paths**. It confines by owner uid — `st_uid != 0 && st_uid != trusted_uid` is `ELOOP` — and has no anchor.

Off Linux, `anchor_parent`'s `Fallback` sent a *peer-derived* receiver path to the *operator* resolver. That is wrong on both axes at once:

- **too lax on escape** — it trusts a euid-owned symlink, which is exactly what the malicious server plants, and its `outside_session_root()` check is inert with no root. This is the bug being fixed.
- **too strict on ownership** — it refused a prefix symlink owned by some third uid (`owner_walk.rs:486-487`), which `ds_descend` and `RESOLVE_BENEATH` both follow. So this PR also *widens* that one axis off Linux, to Linux's behaviour and to upstream's.

Note the widening does not weaken confinement: a third-uid symlink inside the destination tree still resolves inside it, and the anchor still refuses anything that leaves.

## Blast radius

`anchor_parent` serves six call sites, so off Linux this changes all of them, not only the partial-basis unlink: `lstat_via_sandbox_or_fallback`, `unlink_via_sandbox_or_fallback`, `recursive_unlinkat_via_sandbox_or_fallback`, `mkdirat_via_sandbox_or_fallback`, `symlinkat_via_sandbox_or_fallback`, `renameat_via_sandbox_or_fallback`. The probe run below is what measures that surface rather than only the cell.

The `insecure links` opt-out is unaffected: under the opt-out no sandbox is opened at all (`setup/sandbox.rs:74-76`, `:136-138`), so `anchor_parent` returns `Fallback` at its `sandbox is None` gate and the opt-out arm is reached exactly as before.

## Scope

Only the cleanup unlink is live for oc. `resolve_use_inplace()` returns `inplace && !one_inplace_partial_dir`, and oc does not implement `one_inplace` on the wire receiver (documented gap at `transfer_ops/mod.rs:260-267`), so oc never writes through `partialptr` and the cell's overwrite assertion is structurally unreachable.

## Evidence

Everything below was measured on Linux at the pinned toolchain 1.88.0.

**The changed arm is `#[cfg(not(target_os = "linux"))]`, so a Linux CI leg cannot execute it.** To measure it anyway, a throwaway probe deleted the Linux `openat2` block so `anchor_parent` compiled and ran through the portable walk on Linux:

| run | result |
|---|---|
| baseline (`openat2` arm) | 4015 tests run, **4015 passed**, 4 skipped |
| probe (portable-walk arm) | 4015 tests run, **4015 passed**, 4 skipped |
| diff of failure sets | empty |

Poison control — with the probe applied, making `open_subdir_confined` return `ENOSYS` unconditionally broke **19** tests, so the probed path is genuinely live, not silently bypassed. Those 19 include live daemon transfers and `daemon_relative_implied_parent_confinement::an_in_module_symlinked_component_still_receives_the_transfer`, the availability case.

Mutation kill list (`--no-fail-fast`; `passed + failed == run` on every row):

| mutation | kills |
|---|---|
| drop the absolute-target refusal (`syscall.c:2953-2956`) | `open_subdir_confined::an_absolute_symlink_prefix_is_refused` |
| let `..` rise above the anchor (`syscall.c:2896-2899`) | `open_subdir_confined::a_relative_symlink_prefix_climbing_out_is_refused` |
| refuse **every** symlink | `open_subdir_confined::a_relative_in_tree_directory_symlink_is_still_followed` |
| never walk, hand back the anchor | all 4 `open_subdir_confined` cells |
| `anchor_parent` declines to anchor (emulates the pre-fix non-Linux arm) | `confined_partial_basis_cleanup::partial_basis_cleanup_refuses_an_absolute_partial_dir_without_a_confine_root` |

The last row is the one that names the defect: with the anchoring arm removed, the rootless escape cell fails on the victim file's contents — which is what the shipped non-Linux arm did.

Non-vacuity: the new cells assert on **what the out-of-tree file contains**, not on an error return (`remove_partial_basis_confined` is best-effort and returns `()`), and each refusal cell is paired with a companion proving a legitimate relative in-tree directory symlink is still followed.

**Availability, all four Linux legs** (upstream testsuite 3.5.0, same host, same harness):

| leg | before | after |
|---|---|---|
| nonroot TCP | 119 pass / 4 fail / 32 skip | 119 / 4 / 32 — emitted manifests **byte-identical** |
| root TCP | 135 pass / 5 fail / 15 skip | 135 / 5 / 15 — emitted manifests **byte-identical** |
| nonroot pipe | — | 261 pass / 0 fail / 84 skip |
| root pipe | — | 289 pass / 0 fail / 56 skip, harness exit 0 |

The failures are the pre-existing `proto-*` cluster plus `daemon-symlink-escape-matrix`, all present before the change. `malicious-server-partial-basis-symlink-overwrite` passes on both Linux legs before and after.

fmt (`tools/ci/check_rustfmt_all.py`), `cargo clippy --locked --workspace --all-targets --all-features --no-deps`, and `cargo nextest run --workspace` are clean at 1.88.0. `citation_drift_audit.py --ratchet` is OK.

## Manifests are NOT re-baselined

The two macOS rows still read `fail`. They can only flip on a real measured macOS root run, which is not available here. Reviewers on macOS: the cell to re-measure is `malicious-server-partial-basis-symlink-overwrite` on the `macos.root.tcp` and `macos.nonroot.tcp` legs.


oferchen:
## The macOS red is this PR's own intent, asserted in reverse

`Feature flag combinations / iconv (macos-latest)` fails one cell:

```
FAIL transfer::fstatat_swap_resistance multi_component_path_anchors_or_falls_back_lstat
panicked at crates/transfer/tests/fstatat_swap_resistance.rs:121:
  multi-component paths degrade to the path-based fallback without openat2
```

That assertion is the **pre-change** behaviour. The test branches on `fast_io::openat2_supported()`:

- supported (Linux) → expect `LstatOutcome::At(_)`
- **not** supported (macOS) → expect `LstatOutcome::Std(_)`

and its own comment says the degrade happens "where it does not (macOS, older kernels)". This PR's entire purpose is to make the nested op anchor through the portable confined walk **off Linux too**, so on macOS the outcome is now `At(_)` where the test still demands `Std(_)`. The diff does not touch this file — the test was simply never updated. Same class as the earlier `-K` symlink red: a test asserting a behaviour the fix deliberately withdrew.

**The predicate is the part that needs attention, not just the expected value.** `openat2_supported()` used to be the deciding capability; after this change it is not. Anchoring off Linux no longer follows from openat2 being present, so flipping the `else` arm's expectation while keeping the probe would leave a test whose gate no longer names what decides the outcome — green, but for a reason that has drifted from the code.

Two things worth settling in the fix rather than assumed:

1. **Does the portable walk now anchor on every Unix, or are there residual conditions that still degrade?** If it always anchors, the branch collapses to a single `At(_)` expectation and the capability probe leaves the test entirely. If something still degrades, the test must gate on *that* condition and say so.
2. The `dev`/`ino` equality assertions below the branch are the part that genuinely holds in both worlds; they should stay and are what make the cell more than a shape check.

⚠ This PR also edits `tools/ci/upstream-3.5.0-expect.macos.{root,nonroot}.tcp.txt`. Those rows must be **regenerated from a real macOS run**, not hand-adjusted to match — the same rule applied to the manifest row in #7734.

oferchen:
## Fixed in `3cea999cc` — the predicate now has one owner

The macOS red was this PR's own change asserted in reverse, and the cause is a **duplicated predicate**, not just a stale expectation.

`anchor_parent`'s in-crate test was corrected here (`nested_anchor_live()` flipped to `!cfg!(linux) || openat2_supported()`), but `crates/transfer/tests/fstatat_swap_resistance.rs` carries its own copy of the same question, spelled `fast_io::openat2_supported()`. The two agreed until the portable arm landed, so the transfer copy kept asserting the pre-change contract while looking correct.

Rather than write a third copy of the corrected formula, the capability now has a single owner:

```rust
// crates/fast_io/src/dir_sandbox/at_syscalls/nested.rs
pub fn nested_parent_anchoring_supported() -> bool {
    !cfg!(target_os = "linux") || crate::linux_capabilities::openat2_supported()
}
```

exported from `fast_io` and consumed by both tests. Off Linux it is unconditionally true (the parent resolves through `open_subdir_confined`, this PR's `ds_descend` port); on Linux it tracks `openat2_supported()`, the one state that still degrades.

### The other three `openat2_supported()` sites are correct, each for its own reason

Checked individually rather than swept:

- **`daemon_pull_symlinked_dest_root.rs:351`** gates on the **dest-root** open (`RESOLVE_NO_SYMLINKS` walking interior components), a different mechanism this PR does not touch. Its docstring says so.
- **`disk_commit/process/tests.rs:964`** is `#[cfg(target_os = "linux")]`, and this PR already updated its doc comment to name the off-Linux extension as a coverage gap rather than a capability gap.
- **`sec_1_m_symlink_swap_attack.rs:81`** still reads `cfg!(linux) && openat2_supported()`. That is now too *narrow* — it skips a refusal assertion macOS could run post-change. A coverage loss, not a failure, and widening it would turn on errno assertions unmeasured off Linux. Left alone deliberately; filed as a residual.

### Verification, at the pinned toolchain 1.88.0 on Linux

- `tools/ci/check_rustfmt_all.py` → 3422 files, exit 0. ⚠ It caught two hand-wrapped `pub use` lists on the first run; rustfmt owns that wrapping now.
- `cargo clippy -p fast_io -p transfer --all-targets --all-features` → clean
- `fast_io` anchor/nested cells → 29/29; `fstatat_swap_resistance` + `sec_1_m_symlink_swap_attack` → 9/9

**⚠ The Linux run cannot discriminate.** On Linux the new predicate reduces to exactly the old expression, so a green Linux suite is consistent with the test never consulting it. Non-vacuity was proven separately by forcing the predicate to `false`; prediction registered beforehand, and it held exactly:

```
41 tests run: 39 passed, 2 failed
  FAIL fast_io  ...::lstat_via_sandbox_multi_component_anchors_or_falls_back
  FAIL transfer::fstatat_swap_resistance multi_component_path_anchors_or_falls_back_lstat
```

Both multi-component cells die, both single-component siblings stay green — so each test does read the predicate, and reads it where the branch matters.

**NOT run here:** macOS (cargo cannot execute in that sandbox). The macOS cell is the instrument for the arm that actually changed, exactly as the manifest rows in this PR are.

PR #7731, opened 2026-09-07
refactor(transfer): one rule for the rendered diagnostic path, not a special case

oferchen:
PR #7728 kept the leading slash for a module rooted at `path = /` by special-casing `module_root == Path::new("/")` in the `full_fname` renderer, alongside the general prefix-stripping rule. This collapses the two into one.

`strip_prefix_root` already maps a `/` module root to `None`, so `p1` becomes `slash_path(curr_dir)` = `"/"`, `p2` is empty, and the render is `"/{tail}"` - byte-identical to what the special case produced, and still correct for a deeper `curr_dir`.

## This fixes no divergence, and that is the claim being defended

The change is a refactor and is labelled as one. The evidence below is what makes deleting #7728's arm safe; it is not offered as a bug fix.

Both arms were built from source and run against real rsync 3.5.0 (`rsync version 3.5.0 protocol version 32`) on Linux, on identical fixtures, comparing the rendered name in the diagnostic. Base is the commit this branch sits on; fix is this branch.

Daemon, the sender flist walk (`rsync: [sender] link_stat/opendir ...`):

| cell | upstream 3.5.0 | base | fix |
|---|---|---|---|
| opendir unreadable subdir | `"noread" (in mod)` | same | same |
| whole-module recurse | `"noread" (in mod)` | same | same |
| `path = /` unreadable | `"/<abs>/noread" (in slashroot)` | same | same |
| `path = /` missing | `"/<abs>/nope" (in slashroot)` | same | same |
| benign pull (control) | rc=0, no diagnostic | same | same |

Server sender over `--rsh`, the non-daemon arm:

| cell | upstream 3.5.0 | base | fix |
|---|---|---|---|
| relative missing | `"/<abs>/src/sub/nope"` | same | same |
| relative unreadable dir | `"/<abs>/src/noread"` | same | same |
| relative with a `./` segment | `"/<abs>/src/sub/nope"` (collapsed) | same | same |
| absolute unreadable dir | `"/<abs>/src/noread"` | same | same |

Nine cells, two transports: base, fix and upstream agree everywhere. The renderer was already upstream-correct, so this adds tests around behaviour that is right and removes a redundant branch.

## Behaviour-neutrality is proven by test, not by reading

#7728's own two assertions now ride inside `module_rooted_at_slash_keeps_the_leading_slash`, and its negative control `module_rooted_at_slash_below_the_root_is_unchanged` is retained unchanged in intent. Both are green.

Merging those assertions rather than deleting the tests that stopped typechecking after the refactor is a visible decision with a visible result: this revision carries all three assertions plus the separate negative control, where an earlier un-rebased revision of the branch had two assertions and no control.

New coverage the renderer previously had none for: a `.` component collapsed out of a rendered name, and a non-daemon name anchored at `curr_dir`.

## Scope

The renderer's only production callers are the transfer crate's sender flist walk (`generator/file_list/walk.rs`, `generator/protocol_io.rs`) and `pipeline/receiver.rs`. A local copy renders its diagnostics through the engine, which never calls this code, so nothing here touches the local path.

That scoping is load-bearing rather than incidental. This task was originally filed as a fix for a local-copy diagnostic divergence; measuring it is what showed the local path does not reach `full_fname` at all. The real divergence is tracked separately with the engine named as its producer.

## Gates

`cargo nextest run -p transfer`: 2478 passed / 0 failed. `tools/ci/check_rustfmt_all.py` exit 0. Clippy exit 0. All re-run on the squashed commit rather than inherited from the pre-squash tip.

Disclosed for completeness: running `cargo test -p transfer --lib` by hand reports 93 failures. They are inherited, not introduced - the base commit reports 91 on the same host, and the two that differ are flaky on the base as well. One failed 1 of 5 runs on base and 4 of 5 on this branch, passed 6 of 6 in isolation, and passed 3 of 3 on both sides at `--test-threads=1`. The cluster is process-global state racing between threads inside one test binary; `cargo nextest`, which CI uses, runs a process per test.

### DIFF
--- a/crates/transfer/src/full_fname.rs
+++ b/crates/transfer/src/full_fname.rs
@@ -15,13 +15,10 @@
 //! `not creating new %s "%s"` at `generator.c:1380`) never gain it, so they must
 //! keep formatting their own quotes.
 //!
-//! # Module-relative rendering
+//! # The two independent axes
 //!
-//! A daemon server `chdir()`s into the module root (`clientserver.c:993`
-//! `change_dir(module_chdir, CD_NORMAL)`), so every path it later handles is
-//! *relative* to that root and the absolute server-side location never reaches
-//! the client. `full_fname()` re-attaches only the part of `curr_dir` that lies
-//! below the module root:
+//! `full_fname()` makes **two** decisions, and they are gated on two different
+//! globals:
 //!
 //! ```c
 //! if (*fn == '/')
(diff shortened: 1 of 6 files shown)


oferchen:
## Windows IOCP red: root-caused, one predicate

`Windows IOCP (--features iocp)` fails 4 cells in `crates/transfer/src/full_fname.rs`, all introduced by this PR. Not flaky, not inherited — master's CI is green on the same job, and the whole diff is inside `crates/transfer`.

**The assertions, from the job log:**

```
appends_module_suffix_after_closing_quote
  left:  "\"/srv/mod/sub/denied.txt\" (in mymod)"
  right: "\"sub/denied.txt\" (in mymod)"

curr_dir_below_module_root_anchors_with_a_leading_slash
  left:  "\"/sub//srv/mod/sub/denied2\" (in mymod)"
  right: "\"/sub/denied2\" (in mymod)"
```

Row 2 is the one that names the mechanism: `/sub` + `/` + `/srv/mod/sub/denied2`. That is `p1 + p2 + tail` with **`tail` never stripped** — the module root got spliced into the middle rather than removed from the front.

**One cause, at `full_fname.rs:168`:**

```rust
let tail = if path.is_absolute() {
    slash_path(path.strip_prefix(self.curr_dir).ok()?)
} else {
    slash_path(path)
};
```

`Path::is_absolute()` is **host-dependent**. On Unix `/srv/mod/...` is absolute; on Windows it has a root but no drive prefix, so `is_absolute()` is **false**, the `else` arm runs, and `tail` keeps the whole server-side path. Both failing shapes fall out of that single inversion, and the other two cells are the same branch.

**The fix is to use upstream's own test.** `util1.c:1445` is `if (*fn == '/')` — a leading-byte check on a POSIX string, not a filesystem-semantics query. These are wire/diagnostic path strings that upstream keeps as `char *`; asking `std::path` whether one is "absolute" imports host rules that upstream never applies. Testing the leading `/` is both the faithful port and host-independent, so it fixes Windows without changing a single Unix result.

Worth stating as a rule for this file, since it already imports `Component`/`Path`/`PathBuf`: **a path that exists to be rendered into a message, or that came off the wire, is a byte string with `/` separators — manipulate it as one.** `Path` semantics are correct for touching the filesystem and wrong for reproducing upstream's string arithmetic. Same class as the `operand_is_remote` and `escape_path` non-Unix divergences already fixed in this tree.

The four cells' `right` values are the upstream-derived expectations and should not move; only the predicate changes.

oferchen:
## The Windows reds on #7731 and #7733 are ONE root cause

Both PRs render an upstream diagnostic path, both are green everywhere except Windows, and both fail for the same reason: **`std::path` treats a `/`-rooted string as not-absolute on Windows.**

`Path::new("/no/such/entry")` has a *root* but no *prefix* (no `D:`), so `Path::is_absolute()` is `false` and `cwd.join(that)` keeps the cwd's **drive** while replacing only the root:

```
#7733  an_absolute_operand_is_named_unchanged
  left:  "D:\\no\\such\\entry"     <- cwd's drive survived the join
  right: "/no/such/entry"
```

#7733's helper doc states the premise that fails: *"An operand that is already absolute is left alone by the join, which is what upstream's `*fn == '/'` branch and its `p1`/`p2` branch both produce for it."* That holds on Unix and is false on Windows — the join is doing the platform's thing, not upstream's.

**Upstream's test is a byte comparison, not a path-type query** — `util1.c:1445`:

```c
if (*fn == '/')
```

It asks "does this string start with a slash", which is host-independent by construction. `Path::is_absolute()` asks "is this absolute *on this platform*", which is a different question that happens to coincide on Unix.

### The fix shape, same for both

Decide the absolute/relative split on the **byte string** — `starts_with('/')` on the rendered path — and keep `std::path` out of that one decision. These are *diagnostic names*: their contract is to reproduce upstream's text, not to name something the local OS can open. A drive-qualified `D:\no\such\entry` is a different name than the operand the user typed, and upstream would print the operand.

The rest of each helper (join-then-lexically-collapse for a genuinely relative operand) is correct and stays — on Windows a relative operand *should* pick up the cwd's drive, and that is what `join` already does.

⚠ Do **not** fix this by gating the tests to Unix. The assertions are upstream-derived; the divergence is in the production predicate, and narrowing the tests would leave a live Windows divergence pinned as intended behaviour.

⚠ The two PRs are independent — same class, different files (`full_fname.rs` vs `operand_diagnostic_name`). Neither fix subsumes the other; each needs its own byte-test change.

PR #7733, opened 2026-09-07
fix(engine): name a failed link_stat operand by its absolute path

oferchen:
## The oracle, measured first

Established by running the real binary, not by reading the C. `rsync version 3.5.0 protocol version 32`, from `/tmp/t1158`, each `-a <operand> dst/`:

```
nope                -> rsync: [sender] link_stat "/tmp/t1158/nope" failed: No such file or directory (2)
sub/nope            -> ... link_stat "/tmp/t1158/sub/nope" failed: ...
./nope              -> ... link_stat "/tmp/t1158/nope" failed: ...
../nope (from sub/) -> ... link_stat "/tmp/t1158/nope" failed: ...
link/nope           -> ... link_stat "/tmp/t1158/link/nope" failed: ...   <- symlink spelling KEPT
/tmp/t1158/nope     -> ... link_stat "/tmp/t1158/nope" failed: ...
nope/               -> a change_dir diagnostic, NOT link_stat
```

The decisive row is `link/nope`. `push_dir()` appends onto the `curr_dir` string and cleans it **lexically** — it never re-reads the path from the kernel — so the fix must join-and-collapse and must **not** `canonicalize()`. A canonicalising fix would have diverged on exactly that row.

## The oc text, measured on the BASE arm

Base arm built from source and run on the same fixture:

```
link_stat "nope" failed: No such file or directory (os error 2)
oc-rsync error: link_stat "nope" failed: ... (code 23) at error.rs(252) [sender=0.6.4]
```

The operand is reported as typed. Exit code 23 (RERR_PARTIAL) already matched.

## Producer confirmed by measurement, not by reading

The brief pointed at `crates/engine/src/.../error.rs:252`. That is **not** where it lives — oc's own `at error.rs(252)` suffix identifies `crates/core/src/client/error.rs:252`, the second *emitter*, not the producer.

The producer is the single construction site of `LocalCopyError::link_stat_failed`, in the `SourceMetadataResult::NotFoundError` arm of `crates/engine/src/local_copy/executor/sources/orchestration.rs`. Confirmed by changing that one site and re-running: **both** emitted lines moved together (the engine's `eprintln!("{error}")` at orchestration.rs:524 and the `ClientError` the same value is mapped to via `first_io_error`). A change anywhere else could not have moved both.

Fix-arm vs oracle, 5/5 quoted names MATCH exactly: `nope`, `sub/nope`, `./nope`, `/abs/nope`, `../nope`.

## The change

- New `operand_diagnostic_name()` in `crates/engine/src/local_copy/mod.rs`: joins the process working directory onto the whole operand and runs the existing `filters::collapse_dot_dot_dirs` port (`lexically_normalize`). Falls back to the operand as given if `current_dir()` fails. Cites `rsync-3.5.0/util1.c:1433-1464` and `rsync-3.5.0/flist.c:2688-2697` — both ranges verified by reading those files at the 3.5.0 pin, not copied from the brief.
- One call site changed in `orchestration.rs`.
- 3 engine unit tests over the helper + a 5-test root integration file that runs the shipped binary end to end, including a cross-implementation leg against the real 3.5.0.

## Mutation results (explicit counts; `run == passed + failed` in every arm)

**M1 — revert the call site to `source_path.to_path_buf()` (the exact pre-fix state):**

- e2e: **5 run / 1 passed / 4 failed.** Kills `a_relative_operand_is_reported_by_its_absolute_path`, `a_relative_operand_with_a_parent_keeps_the_whole_tail`, `dot_and_dot_dot_components_collapse_lexically`, `the_quoted_name_matches_a_real_rsync`. Survivor is `an_absolute_operand_is_reported_unchanged` — the negative control.
- engine lib: **58 run / 58 passed / 0 failed.** Stated plainly: the unit tests **cannot** see a call-site regression, because they exercise the helper directly. That is precisely why the end-to-end file exists and is the load-bearing pin.

**M2 — mutate the helper body to return the operand unchanged:**

- engine lib: **58 run / 56 passed / 2 failed.** Kills `a_relative_operand_is_named_by_its_absolute_path` and `dot_and_dot_dot_collapse_the_way_push_dir_cleans_curr_dir`; `an_absolute_operand_is_named_unchanged` survives.

Fix restored afterwards; both scopes re-run green (58/58 and 5/5), diff footprint byte-identical to pre-mutation.

## A vacuity trap hit and closed

The e2e run first reported `5 tests run: 5 passed` in **0.037 s**. Implausible for five process-spawning tests — and it was: that worktree has no `target/interop`, so `upstream_rsync()` returned `None` and `the_quoted_name_matches_a_real_rsync` **self-skipped silently**. Re-run with `OC_RSYNC_UPSTREAM_RSYNC` pointed at the real 3.5.0 binary: the cell takes 0.23 s against 0.03 s for its siblings, and passes. In CI the interop harness builds the oracle so the cell runs there; but the skip is silent, which is the same shape as the skip-as-pass class already on the ledger. Filed as a follow-up.

## Residual divergences — measured, NOT fixed, filed separately

1. oc line 1 lacks upstream's `rsync: [sender] ` prefix.
2. oc line 2 repeats the link_stat text; upstream prints the generic `rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1394)` summary instead.

Both are pre-existing and independent of this change.

## Verification

Run, at the pinned toolchain 1.88.0 (`rustc 1.88.0 (6b00bc388)`), on Linux:

- `python3 tools/ci/check_rustfmt_all.py` -> 3425 files at edition 2024, exit 0
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` -> exit 0
- `cargo nextest run -p engine --lib --all-features` -> **4872 run / 4872 passed / 23 skipped**
- `cargo nextest run --test link_stat_names_the_absolute_path --all-features` -> **5 run / 5 passed**, with the upstream oracle wired in

NOT run, and why:

- **Full-workspace `--all-features` nextest.** Started, then stopped by PID when the build host fell to 3G free with ~41 of 106 root integration binaries still to link. Named rather than papered over. The host has since been reclaimed to 68G free, so this is re-runnable on request.
- macOS / Windows / musl cells, the 3.5.0 testsuite legs, interop.

### DIFF
--- a/crates/engine/src/local_copy/executor/sources/orchestration.rs
+++ b/crates/engine/src/local_copy/executor/sources/orchestration.rs
@@ -746,8 +746,12 @@ fn process_single_source(
             // RERR_PARTIAL), and the transfer continues with the remaining
             // sources. Distinct from a file that vanishes mid-transfer (exit
             // 24, "file has vanished").
+            //
+            // The name upstream prints is the operand anchored to the working
+            // directory, not the operand as typed: `full_fname()` puts
+            // `curr_dir` in front of every relative `fn` (util1.c:1445-1452).
             return Err(LocalCopyError::link_stat_failed(
-                source_path.to_path_buf(),
+                crate::local_copy::operand_diagnostic_name(source_path),
                 error,
             ));
         }
--- a/crates/engine/src/local_copy/mod.rs
+++ b/crates/engine/src/local_copy/mod.rs
@@ -238,6 +238,51 @@ pub(crate) fn lexically_normalize(path: &std::path::Path) -> std::path::PathBuf 
     filters::collapse_dot_dot_dirs(path)
 }
 
+/// Names a source operand the way upstream's sender diagnostics name it: with
+/// the working directory in front, so a relative operand is reported by its
+/// absolute path.
+///
+/// Upstream arrives at that string in two steps. `send_file_list()` splits each
+/// operand into a `dir`/`fn` pair and `push_dir()`s into `dir` before the
+/// `link_stat()` at `flist.c:2697`, which appends onto the `curr_dir` *string*
+/// and cleans it lexically - it never re-reads the path from the kernel, so a
+/// symlinked component keeps the spelling the operator typed. `full_fname()`
+/// then prefixes the surviving relative `fn` with `curr_dir` before quoting it:
+///
+/// ```text
+/// if (*fn == '/')
+///         p1 = p2 = "";
+/// else {
+///         p1 = curr_dir + module_dirlen;
+///         for (p2 = p1; *p2 == '/'; p2++) {}
+///         if (*p2)
+///                 p2 = "/";
+/// }
+/// ```
+///
+/// `fn` is a single component, so joining the working directory onto the whole
+/// operand and collapsing `.`/`..` lexically reproduces both steps at once. A
+/// local copy is never a daemon server, so `module_dirlen` is 0 and the prefix
+/// is the whole working directory.
+///
+/// An operand that is already absolute is left alone by the join, which is what
+/// upstream's `*fn == '/'` branch and its `p1`/`p2` branch both produce for it.
+///
+/// # Upstream Reference
+///
+/// - `rsync-3.5.0/util1.c:1433-1464` - `full_fname()`.
+/// - `rsync-3.5.0/flist.c:2688-2697` - the `dir`/`fn` split and the `push_dir()`
+///   that makes `curr_dir` the operand's parent before `link_stat()` runs.
+pub(crate) fn operand_diagnostic_name(path: &std::path::Path) -> std::path::PathBuf {
+    let Ok(working_dir) = std::env::current_dir() else {
+        // upstream keeps `curr_dir` in a global it can always read; with no
+        // working directory to anchor against, the operand as given is the only
+        // name left to report.
+        return path.to_path_buf();
+    };
+    lexically_normalize(&working_dir.join(path))
+}
+
 #[cfg(test)]
 pub(crate) fn with_hard_link_override<F, R>(override_fn: F, action: impl FnOnce() -> R) -> R
 where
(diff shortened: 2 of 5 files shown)


oferchen:
## The Windows reds on #7731 and #7733 are ONE root cause

Both PRs render an upstream diagnostic path, both are green everywhere except Windows, and both fail for the same reason: **`std::path` treats a `/`-rooted string as not-absolute on Windows.**

`Path::new("/no/such/entry")` has a *root* but no *prefix* (no `D:`), so `Path::is_absolute()` is `false` and `cwd.join(that)` keeps the cwd's **drive** while replacing only the root:

```
#7733  an_absolute_operand_is_named_unchanged
  left:  "D:\\no\\such\\entry"     <- cwd's drive survived the join
  right: "/no/such/entry"
```

#7733's helper doc states the premise that fails: *"An operand that is already absolute is left alone by the join, which is what upstream's `*fn == '/'` branch and its `p1`/`p2` branch both produce for it."* That holds on Unix and is false on Windows — the join is doing the platform's thing, not upstream's.

**Upstream's test is a byte comparison, not a path-type query** — `util1.c:1445`:

```c
if (*fn == '/')
```

It asks "does this string start with a slash", which is host-independent by construction. `Path::is_absolute()` asks "is this absolute *on this platform*", which is a different question that happens to coincide on Unix.

### The fix shape, same for both

Decide the absolute/relative split on the **byte string** — `starts_with('/')` on the rendered path — and keep `std::path` out of that one decision. These are *diagnostic names*: their contract is to reproduce upstream's text, not to name something the local OS can open. A drive-qualified `D:\no\such\entry` is a different name than the operand the user typed, and upstream would print the operand.

The rest of each helper (join-then-lexically-collapse for a genuinely relative operand) is correct and stays — on Windows a relative operand *should* pick up the cwd's drive, and that is what `join` already does.

⚠ Do **not** fix this by gating the tests to Unix. The assertions are upstream-derived; the divergence is in the production predicate, and narrowing the tests would leave a live Windows divergence pinned as intended behaviour.

⚠ The two PRs are independent — same class, different files (`full_fname.rs` vs `operand_diagnostic_name`). Neither fix subsumes the other; each needs its own byte-test change.

PR #7737, opened 2026-09-07
fix(daemon): a repeated exclude/include/filter directive replaces the previous value

oferchen:
A daemon config that names the same filter directive twice kept BOTH values.
Upstream keeps the last one and discards the first.

    [mod]
    exclude = *.log
    exclude = *.tmp

Upstream serves `a.log` and hides only `*.tmp`. oc hid both. Measured with a
real rsync 3.5.0 binary held constant as the client and only the daemon
varying: all six affected arms diverge on the base and match upstream after
the fix.

The rule is `do_parameter()` reaching `string_set()` (loadparm.c:379-470),
which frees the previous value before storing the new one. There is no
seen-set and no duplicate check anywhere in that path: a second occurrence of
a parameter simply overwrites the first. That holds for the global section
too, where a P_LOCAL parameter writes the default `init_section()` /
`copy_section()` hand to every module.

WHY EXACTLY THESE THREE DIRECTIVES

`GlobalModuleDefaults` has 40 fields. 37 are `Option<T>`; exactly three are
`Vec<String>`, and they are `exclude`, `include` and `filter`. An `Option`
field can only be assigned, so "does a repeat replace?" is true by
construction and no code has to enforce it - the question cannot even be
asked. The three `Vec` fields are the only ones whose type can express
accumulation, and they are precisely the three that got it wrong. That is the
survival mechanism: for 37 of 40 fields there was nothing for a reviewer to
check.

The same shape appears on the module builder, where `last_wins_setters!`
(module_definition/setters.rs) states the convention in its own header -
"Every directive is last-wins ... The single expansion below is the one place
that assignment happens" - and the three filter slots sat outside the macro
because their type is not `Option`. The convention was written down and these
three escaped it.

SCOPE - six arms across two files

Three module arms in `config_parsing/module_directives.rs` and three global
arms in `config_parsing/global_directives/dispatch.rs`. Each now assigns a
one-element `Vec` holding the directive's whole value. The module arms go
through new named setters (`last_wins_rule_setters!`, a sibling of the
existing macro) so the three slots are inside the convention rather than
beside it; one edit to that macro body reaches all three, which the mutation
matrix confirms.

The `Vec` is deliberately NOT collapsed to a `String`. It stays a `Vec` with
one entry because the inheritance path depends on the shape, and that path was
already correct - see the negative control below.

DELETED IN THIS COMMIT: `module_directives.rs` carried the comment
"Repeatable: multiple directives accumulate rules." It is false, it is the
statement this change refutes, and it is why the behaviour read as intended.

WHAT DID NOT CHANGE, AND IS PINNED AS SUCH

A module's own value replacing an inherited global default was ALREADY
correct on both implementations (`exclude = foo` globally plus
`exclude = bar` in the module serves `foo`, hides `bar`). The defect is
repeats within one section only. `parse_module_exclude_replaces_the_inherited
_global_default` pins that boundary: if a future edit made the module arm
extend the inherited default, it fails while every last-wins cell stays green
- a distinction those cells cannot draw alone.

`parse_module_filter_one_directive_keeps_its_whole_value` is the non-vacuity
companion: several rules inside ONE directive value are all kept. Without it,
"last wins" would also be satisfied by a parser that simply dropped rules.

The two other `.push(` sites in the global file are untouched and correct by
intent: `refuse options` clears then pushes and documents "The list is not
additive", and `motd` documents having no upstream counterpart.

TWO EXISTING TESTS ASSERTED THE DEFECT
`parse_module_filter_multiple_rules_accumulate` and its `exclude` sibling
encoded the accumulating behaviour by name. They are converted, not deleted -
same fixture, corrected expectation, renamed to `_last_directive_wins` to
match the ten sibling cells.

MUTATION MATRIX - 8 mutations, all lethal, predictions registered first
Population 15 cells, baseline 15/15 at pinned 1.88.0.

  M1-M3  each module arm back to `push`   -> kills exactly its own cell
  M4-M6  each global arm back to `push`   -> kills exactly its own cell
  M7     setter macro body to `.extend()` -> kills all THREE module cells at
         once and leaves the three global cells green: the single-owner proof
  M8     module filter arm splits the value on whitespace -> kills the
         non-vacuity companion

One registered prediction was WRONG: M8 was predicted to kill 2 cells and
killed 3. It also reddens `parse_module_filter_single_rule`, whose value
`- *.tmp` splits into two tokens. Reported as measured; the mutation is not
re-scoped to fit the prediction.

### DIFF
--- a/crates/daemon/src/daemon/sections/config_parsing/global_directives/dispatch.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/global_directives/dispatch.rs
@@ -538,9 +538,16 @@ fn apply_global_directive(
         // When bInGlobalSection is true, parm_ptr = def_ptr, so the value
         // written becomes the default for all subsequent module sections
         // (via init_section -> copy_section).
+        //
+        // Last-wins here too: `do_parameter()` resolves the same parameter
+        // pointer whether the section is global or a module and assigns it
+        // through `string_set()` (loadparm.c:379-470), so a repeated global
+        // `exclude` REPLACES the default rather than extending it. These three
+        // are the only `Vec`-shaped fields on `GlobalModuleDefaults`; every
+        // other field is an `Option`, which is last-wins by construction.
         "exclude" => {
             if !value.is_empty() {
-                state.module_defaults.exclude.push(value.to_owned());
+                state.module_defaults.exclude = vec![value.to_owned()];
             }
         }
         // upstream: daemon-parm.txt `Locals:` `include` is P_STRING/P_LOCAL
@@ -550,12 +557,12 @@ fn apply_global_directive(
         // default, mirroring `exclude` above.
         "include" => {
             if !value.is_empty() {
-                state.module_defaults.include.push(value.to_owned());
+                state.module_defaults.include = vec![value.to_owned()];
             }
         }
         "filter" => {
             if !value.is_empty() {
-                state.module_defaults.filter.push(value.to_owned());
+                state.module_defaults.filter = vec![value.to_owned()];
             }
         }
         "maxverbosity" => {
--- a/crates/daemon/src/daemon/sections/config_parsing/module_directives.rs
+++ b/crates/daemon/src/daemon/sections/config_parsing/module_directives.rs
@@ -364,23 +364,27 @@ fn apply_module_directive(
             let resolved = resolve_config_relative_path(canonical, value);
             builder.set_include_from(resolved);
         }
-        // upstream: daemon-parm.h - `filter` STRING, P_LOCAL.
-        // Repeatable: multiple directives accumulate rules.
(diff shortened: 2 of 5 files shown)


oferchen:
The interop red was the fixture, not the fix: the daemon-server-side-filter scenario wrote two `exclude` lines in one section and asserted both patterns apply - i.e. it encoded the accumulate semantics this PR removes. Upstream 3.5.0 is last-wins on a repeated directive (loadparm string_set), so with that config only `*.log` is excluded and `build.tmp` transferring is the correct behaviour. Fixed in cc0176ef3 by putting both patterns on one directive line (`exclude = *.tmp *.log`), which both implementations expand identically; the cell still pins server-side exclusion of both suffixes plus the `filter` directive, pull and push. These two configs were the only repeated exclude/include/filter directives in the interop script. The fp=28 cluster in the same job is the known non-blocking RP28 baseline (warned, not gating).

PR #7795, opened 2026-09-09
fix(daemon): honour `merge` and `!` in a module filter rule

oferchen:
A daemon `filter` rule naming a merge file was parsed as an ordinary exclude of the
literal word `merge`, so the file was never opened and none of its rules applied. A
module configured to hide a file served it.

## Measured against real rsync 3.5.0

Rig: oc as daemon over loopback TCP, upstream 3.5.0 as the client (`-r --list-only`),
module holding `bait keep ctl rules sub/f deep/sub/f d/x` plus a file literally named
`!`. Oracle is the same rig with the pinned upstream binary on both ends.

| cell | upstream | oc before | oc after |
|---|---|---|---|
| `merge /abs/rules` | hides `bait` | **serves `bait`** | hides |
| `. /abs/rules` | hides | **serves** | hides |
| `merge ./rules` | hides (module-relative) | **serves** | hides |
| `merge rules` (no slash) | **rc 5** open failure | **rc 0** | rc 5 |
| `merge /nonexistent` | **rc 5** | **rc 0** | rc 5 |
| `merge` (no name) | rc 5 `unexpected end of filter rule` | rc 0 | rc 5 |
| `- ctl merge FILE` | hides both | hides `ctl` only | hides both |
| file `+ bait` / `- *` | serves only `bait` | serves all | matches |
| file `H bait` | hides | serves | hides |
| file `- sub/f` | hides `sub/f` **and** `deep/sub/f` | serves both | matches |
| file `- d/` | hides `d`, `d/x` | serves | matches |
| file `bogusword` | rc 5 `Unknown filter rule` | rc 0 | rc 5 |
| `.!` | rc 5 `invalid modifier '!' at position 1` | rc 0 | rc 5 |
| `- bait ` (trailing space) | **serves** | serves | serves |

25 A/B cells now match upstream on both the served set and the exit code.

## The two clear forms do NOT share a cause

The row was filed as one defect covering `filter = clear` and `filter = !`. Measurement
splits it: **`clear` was already correct** — three cells converged on the unfixed tree.
Only `!` was broken, and it was broken in the direction that loses data:

| cell | upstream | oc before | oc after |
|---|---|---|---|
| `!` | serves the file named `!` | **hides it** | serves |
| `- bait - ctl !` | all served | **hides `bait`** | all served |
| `!,` | rc 0 | **rc 5 refused** | rc 0 |

Two further `!` cells converged on the unfixed tree **by accident** — the glued token's
pattern matched nothing. That is why the committed fixture uses two excludes: with one,
the test passes before and after and proves nothing. The comment in the chunk says so.

## Shape

`merge_rule()` classifies a token; `push_merge_file_rules()` is the one place a merge
file's records are appended, in file order, at the position the merge rule occupied
(`exclude.c:1553-1590`). Records inherit `XFLG_FATAL_ERRORS` only, matching upstream's
`FILTRULES_FROM_CONTAINER` (`exclude.c:1229-1231`), so a record's rules are *not* given
the container's own flags — measured: a module whose `filter = - FILE` serves a file
that `filter = merge FILE` hides. `MAX_MERGE_DEPTH` mirrors `exclude.c:168`.

`.` and `:` are deliberately **not** tokenizer openers. Making them openers would split
`filter = - .git` into a patternless `-` plus a pattern, and refuse the commonest daemon
filter there is. The same reasoning narrows the `!` opener to the whole words `!` and
`!,` rather than any leading `!`.

## Mutations

14 mutations over the 196-test filter family. `M3` (merge dispatch removed) kills 19;
the `!` arms kill 2 and 6; depth-limit removal reports `ABRT` from stack overflow rather
than `FAIL`, verified by hand. `M13` (`e` modifier adds no rule) killed **nothing** on
the first sweep — a real coverage hole, closed by
`the_exclude_self_modifier_hides_the_merge_files_basename`, which it now kills. Restored
after: `-p daemon -p filters --all-features` 3602 passed / 6 skipped.

## Three cells still diverge, all more-restrictive, all named

- `filter = - keep !bait` — upstream rc 5, oc rc 0. Pre-existing token-gluing residual;
  identical on the unfixed tree.
- `filter = .e FILE` — upstream adds the basename exclude then reads **nothing**, via its
  own daemon-filter self-suppression (`exclude.c:1636-1657`); oc adds the exclude and
  still merges.
- `filter = merge,w FILE` — upstream word-splits and refuses; oc reads line-wise and
  builds a harmless compound pattern.

`merge,C` reproduces only the `NO_PREFIXES` half, not the default-cvsignore-list load
(`get_cvs_excludes`); that matched upstream on the measured cell.

## A harness vacuity found on the way, larger than this change

The first version of these end-to-end tests copied `start_daemon_pending_no_detach` from
a sibling chunk. That helper does not force `--no-detach`, so the daemon forks and the
**parent — the test process — calls `process::exit(0)` mid-test**. Under nextest's
process-per-test model that reads as a PASS. All four tests "passed" on the unfixed tree
until they were switched to `start_daemon`.

The repo already knows this shape (`force_no_detach`'s doc, `check_daemon_no_detach.sh`)
and the ceiling is **77 pending call sites** — including
`daemon_exclude_overrides_client_include` and
`daemon_exclude_directory_blocks_subtree`, which are asserting nothing today. Filed
separately; this PR does not touch the ceiling.

## Gates (pinned 1.88.0)

`check_rustfmt_all.py` clean; `clippy -p daemon -p filters --all-targets --no-deps
-D warnings` clean; `nextest -p daemon -p filters --all-features` 3602/3602;
`check_daemon_no_detach.sh` clean at the unchanged 77 ceiling.

Not run: Windows, Linux, musl (macOS only — the new e2e chunk is `#[cfg(unix)]`, the unit
tests are portable), the interop harness, the upstream testsuite, full-workspace nextest.
No expect-manifest is touched, so no derived tally is perturbed.

### DIFF
--- a/crates/daemon/src/daemon/sections/module_access/helpers.rs
+++ b/crates/daemon/src/daemon/sections/module_access/helpers.rs
@@ -223,13 +223,15 @@ fn build_daemon_filter_rules(
     // This function already returns `Result`, and its caller already refuses
     // the connection on the file-reading arms below - the refusal PATH is
     // unchanged here; only which tokens enter it is new.
+    //
+    // A `merge` token expands to the rules of the named FILE, which is why the
+    // loop pushes through [`push_token_rules`] rather than taking one rule per
+    // token. `module.path` is upstream's `dirbuf`: `set_filter_dir(module_dir,
+    // module_dirlen)` runs immediately before this parse (clientserver.c:922-926)
+    // and is what `parse_merge_name()` prepends to a slash-bearing relative name.
     for filter_str in &module.filter {
         for token in split_filter_tokens(filter_str.trim()) {
-            if let Some(rule) =
-                parse_daemon_filter_token(&token).map_err(MalformedRule::into_io_error)?
-            {
-                rules.push(rule);
-            }
+            push_token_rules(&mut rules, &token, &module.path, 0, RuleXflags::Daemon)?;
         }
     }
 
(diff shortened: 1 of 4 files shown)


oferchen:
## Auto-merge disarmed pending one arm — the `!` and `merge` fixes are good

I have taken auto-merge off this PR (nothing was merged). Two of its three parts stand on their
own measurement and I am not disputing them: the bare-`!` fix closes a real data-losing defect
(`!` reaching the bare-word fall-through and becoming a literal exclude of the pattern `!`), and
the `merge` arm is measured. **The blocker is the `dir-merge` / `:` arm only.**

### The claim under dispute

`MergeRule::per_dir`'s doc records, as a measurement:

> Nothing is read here... the daemon list never descends per directory, so it can match nothing.
> MEASURED: `filter = dir-merge rules` serves every file of the module, as `filter = : rules` does.

**That is refuted.** Measured twice, independently — once by an investigating agent, then re-run
by me from its script before accepting the result. Real upstream `rsync 3.5.0` as both daemon and
client over loopback, `-r --list-only`, with a **per-cell `lsof` port-ownership assert** so a
colliding daemon cannot forge the result (that trap has produced a false upstream reading in this
exact area before).

The discriminator is that the merge file exists **only in a subdirectory**, so a one-shot root read
and a per-directory descent give different answers:

```
mod/sub/rules   contains "- f"     <- the ONLY merge file
mod/{bait,keep} mod/sub/{f,other}  mod/deep/sub/f
```

| cell | module filter | served |
|---|---|---|
| A0 | (none) | bait deep/sub/f keep **sub/f** sub/other sub/rules |
| A1 | `filter = dir-merge rules` | bait deep/sub/f keep sub/other sub/rules — **`sub/f` HIDDEN** |
| A2 | `filter = : rules` | identical to A1 |
| B1 | root-only `rules` (`- bait`) | **`bait` HIDDEN**, `sub/f` served |
| C1 | `rules` at both levels | **`bait` AND `sub/f` HIDDEN** |

A1 is decisive and the shape is specifically per-directory: the rule found in `sub/` hides `sub/f`
and does **not** reach `deep/sub/`, which has no rules file. C1 shows inheritance down the tree;
B1 shows it firing at the root. `merge`/`.` giving rc 5 matches this PR's own table, so the two
measurements agree everywhere except the `:` arm.

Upstream's mechanism is consistent with that: `parse_rule_tok` maps both `dir-merge` and `:` to
`ch=':'` with `FILTRULE_PERDIR_MERGE|FILTRULE_FINISH_SETUP` (`exclude.c:1293-1296`, `:1325-1338`,
the `:` arm falling through to `.`'s `FILTRULE_MERGE_FILE`); there is no `FILTRULES_SIDES` bit, so
the wrong-side drop at `:279-285` cannot fire; `add_rule` registers it in `mergelist_parents[]`
(`:388-391`); `push_local_filters` populates per directory (`:886-923`); `check_filter` recurses
into `ent->u.mergelist` (`:1195-1200`).

### Why this blocks the merge, stated precisely

The no-op arm is **not a behavioural regression** — today `:` reaches the bare-word fall-through and
builds a literal exclude that matches nothing, so both before and after, everything is served. What
would land is a **doc comment asserting a measured falsehood**, in the unsafe direction (oc serving
files upstream hides, for a module using `dir-merge` to hide content), and it would quietly close
the open task that implements this properly by telling the next reader there is nothing to do.

I am not merging a PR whose stated measurement is contradicted until the contradiction is resolved,
and it now is: two independent probes, one of them mine, on the real binary.

I have **not** diagnosed why the original cell reported "serves everything" — candidate causes are
fixture layout, an oc rather than upstream daemon in that cell, or a port collision. Both probes
above carry the ownership assert; I could not find one stated for the original.

### What I suggest

Drop the `per_dir` no-op arm and its doc claim from this PR and let the `!` + `merge` fixes merge on
their own measurement. The real `dir-merge` implementation is in flight separately and is
**parser-only** — three sites in `helpers.rs`, everything downstream already exists and is already
driven on both network paths.

⚠ Sequencing, since it is the same three functions: this PR touches `short_rule_prefix`,
`build_prefixed_rule` (**new `xflags: RuleXflags` parameter**), `build_pattern_rule` (**new 3rd
parameter**), `parse_daemon_filter_token`, `split_filter_tokens` and `RULE_KEYWORDS`. So the
`dir-merge` work must rebase **onto** this PR, and that rebase is semantic rather than textual — it
has to thread `RuleXflags` through the new arm. Landing this first is the cheaper order.

Both probe scripts are re-runnable in one command and print their own provenance (`pwd`, host,
`rsync --version`) so the arm under test is never ambiguous.

oferchen:
## Independent verification: the `merge` and `!` arms are both SUBSTANTIVE

Following up my earlier comment. Since one of this PR's three claims turned out to be refuted, I
had the other two measured independently rather than taking them on the PR body — the oc column
specifically, which the body's own table asserts.

**Method.** Both arms built from source at pinned 1.88.0 and proven distinct
(base `e91303b14` md5 `2cb1ffb7…`, fix `46d8f9587` md5 `34a423b5…`). oc as daemon, real
`rsync 3.5.0` as the constant client, `-r --list-only`, per-cell `lsof` port-ownership assert on
distinct ports so a stray daemon cannot cross-contaminate. Fixture carries a file literally named
`!` alongside bait/keep/ctl.

### Arm 1 — `merge`: substantive, and it reads at DAEMON PARSE TIME

The open question was whether this arm is inert for a daemon's own `filter = merge FILE`, since
`generator/filters.rs:533-539` is a bare `continue` premised on "pre-expanded by the client".
Answered **no**, structurally: `push_merge_file_rules` never produces a `RuleType::Merge` wire rule
at all. It opens the file, iterates the records and pushes each one into the *same* rule vec, in
file order, at the position the merge rule occupied. `grep 'RuleType::'` across the branch's
`helpers.rs` returns exactly one line — `RuleType::Clear`. So that `continue` is unreachable from
this path and its premise is never relied on: the daemon pre-expands the file itself.

6/6 cells, base diverges on **all six**:

| cell | upstream | oc base | oc fix |
| --- | --- | --- | --- |
| `merge <abs>/rules` | hides bait | **serves bait** | hides ✓ |
| `. <abs>/rules` | hides bait | **serves bait** | hides ✓ |
| `merge rules` (no slash) | **rc 5** | **rc 0, serves all** | rc 5 ✓ |
| `merge /nonexistent` | **rc 5** | **rc 0, serves all** | rc 5 ✓ |
| `merge` (no name) | rc 5 `unexpected end of filter rule` | rc 0 | rc 5, same text ✓ |
| `- ctl merge <abs>/rules` | hides ctl AND bait | hides ctl only | hides both ✓ |

⚠ **The most serious cell is row 3**: on base, `filter = merge rules` exits **0 and serves every
file in the module** where upstream exits **5** (`failed to open exclude file rules`). An operator
whose merge file fails to load gets a served module instead of a refusal. That is the arm worth
landing promptly.

### Arm 2 — bare `!`: substantive, and both halves are witnessed

4/4 cells, base diverges on all four:

| cell | upstream | oc base | oc fix |
| --- | --- | --- | --- |
| `!` | serves all incl. the file named `!` | **the file named `!` VANISHES** | serves all ✓ |
| `- bait - ctl !` | serves all | **hides bait** (clear never fires) | serves all ✓ |
| `!,` | rc 0, serves all | **rc 5** `'!' rule has trailing characters` | rc 0 ✓ |
| `! - bait` | hides bait, serves `!` | **hides bait AND `!`** | bait hidden, `!` served ✓ |

Row 1 witnesses the literal exclude of pattern `!` eating a real file; row 2 witnesses the clear
failing to fire. Both directly, not inferred. Instrument non-vacuity: `no filter` serves everything
and `- bait` hides bait on both arms, so the rig discriminates in both directions. The committed
e2e chunk uses the two-exclude form, which is the discriminating shape rather than the accidental
pair its own comment warns about.

### Arm 3 — `dir-merge`: unchanged, still the blocker

`dir-merge <abs>/rules` and `: <abs>/rules`: base serves bait, **fix serves bait**, upstream hides
bait. So the no-op arm is measured inert against the exact spelling its doc comment claims to have
measured as serving-everything. Recorded here so all three arms sit in one table.

### Disposition

**10/10 cells across arms 1 and 2 converge from base-diverging to upstream-matching.** Those are
real fixes and I would like them on master — the `merge rules` row especially. The only thing
holding the PR is arm 3's doc comment asserting a measured falsehood; the arm's *behaviour* is
identical to master today (both serve everything), so it is not a regression, and task 1169
implements the real per-directory support on top of this base.

So the smallest correct change is to **correct or drop that one doc comment** — not to rework the
arm. Once that is in, this merges.

Caveats stated: this measured the **oc column only**. The upstream column is my own twice-verified
table from the sibling probes on the same fixture semantics, not re-run here. This PR's own test
suite and gates were not re-run — the mutation record in the body covers the two arms
independently, and nothing was built for macOS/Windows/musl beyond one host's debug build.

